PR #3, opened 2020-10-28
Simplify CI

timburks:
unzip seems to be already installed in the base image.


timburks:
Tests still pass, so let's go with this.

PR #25, opened 2021-01-07
OpenAPI linter example

timburks:
This uses the experimental linter in https://github.com/google/gnostic/tree/master/plugins/gnostic-linter to analyze OpenAPI descriptions against simple design rules, storing the results in the registry.

### DIFF
--- a/cmd/registry/core/lint-openapi.go
+++ b/cmd/registry/core/lint-openapi.go
@@ -15,28 +15,30 @@
 package core
 
 import (
+	"fmt"
 	"io/ioutil"
-	"log"
+	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 
 	"github.com/apigee/registry/rpc"
 	linter "github.com/googleapis/gnostic/metrics/lint"
-	sourceinfo "github.com/googleapis/gnostic/metrics/sourceinfo"
 	"google.golang.org/protobuf/proto"
+	yaml "gopkg.in/yaml.v3"
 )
 
-// NewLintFromOpenAPIv2 runs the API linter and returns the results.
-func NewLintFromOpenAPIv2(name string, b []byte) (*rpc.Lint, error) {
+// NewLintFromOpenAPI runs the API linter and returns the results.
+func NewLintFromOpenAPI(name string, b []byte) (*rpc.Lint, error) {
 	// create a tmp directory
 	root, err := ioutil.TempDir("", "registry-openapi-")
 	if err != nil {
 		return nil, err
 	}
-	log.Printf("running in %s", root)
+	// log.Printf("running in %s", root)
 	name = filepath.Base(name)
 	// whenever we finish, delete the tmp directory
-	//	defer os.RemoveAll(root)
+	defer os.RemoveAll(root)
 	// write the file to the temp directory
 	spec, err := GUnzippedBytes(b)
 	if err != nil {
@@ -50,7 +52,7 @@ func NewLintFromOpenAPIv2(name string, b []byte) (*rpc.Lint, error) {
 	lint := &rpc.Lint{}
 	lint.Name = name
 	// run the linter on the spec
-	lintFile, err := lintFileForOpenAPIv2(name, root)
+	lintFile, err := lintFileForOpenAPI(name, root)
 	if err != nil {
 		return nil, err
 	}
@@ -58,7 +60,7 @@ func NewLintFromOpenAPIv2(name string, b []byte) (*rpc.Lint, error) {
 	return lint, err
 }
 
-func lintFileForOpenAPIv2(path string, root string) (*rpc.LintFile, error) {
+func lintFileForOpenAPI(path string, root string) (*rpc.LintFile, error) {
 	cmd := exec.Command("gnostic", path, "--linter-out=.")
 	cmd.Dir = root
 	_, err := cmd.CombinedOutput()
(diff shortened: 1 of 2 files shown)


cmd/registry/core/lint-openapi.go
```diff
@@ -0,0 +1,157 @@
...
+			c := int32(node.Column)
+			problem.Location = &rpc.LintLocation{}
+			problem.Location.StartPosition = &rpc.LintPosition{
+				LineNumber:   l,
+				ColumnNumber: c,
+			}
+			problem.Location.EndPosition = &rpc.LintPosition{
+				LineNumber:   l,
+				ColumnNumber: c + int32(len(node.Value)-1),
+			}
```
seaneganx, line 98: Is there a reason for constructing `rpc.Lint`, `rpc.LintProblem`, and `rpc.LintFile` one field at a time? This seems like it could be more clearly written as:
```
l, c := int32(node.Line), int32(node.Column)
problem.Location = &rpc.LintLocation{
  StartPosition: &rpc.LintPosition{
    LineNumber: l,
    ColumnNumber: c,
  },
  EndPosition: &rpc.LintPosition{
    LineNumber: l,
    ColumnNumber: c + int32(len(node.Value)-1),
  },
}
```

cmd/registry/core/lint-openapi.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func newNodeFinder(filename string) (*nodeFinder, error) {
+	data, _ := ioutil.ReadFile(filename)
+	var node yaml.Node
+	err := yaml.Unmarshal(data, &node)
+	if err == nil {
+		return &nodeFinder{node: &node}, nil
+	}
+	return nil, err
```
seaneganx, line 118: Nit: When guarding error cases it's more common for the success flow to continue after an error check.
```
if err := yaml.Unmarshal(data, &node); err != nil {
  return nil, err
}
return &nodeFinder{node: &node}, nil
```

cmd/registry/core/lint-openapi.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/apigee/registry/rpc"
+	linter "github.com/googleapis/gnostic/metrics/lint"
+	"google.golang.org/protobuf/proto"
+	yaml "gopkg.in/yaml.v3"
+)
+
+// NewLintFromOpenAPI runs the API linter and returns the results.
+func NewLintFromOpenAPI(name string, b []byte) (*rpc.Lint, error) {
+	// create a tmp directory
+	root, err := ioutil.TempDir("", "registry-openapi-")
```
seaneganx, line 33: Could anything go wrong if this directory isn't cleaned up properly? Using `ioutil.TempDir("", "registry-openapi-*")` would result in different directories between runs.

cmd/registry/core/lint-openapi.go
```diff
@@ -0,0 +1,157 @@
...
+	// run the linter on the spec
+	lint := &rpc.Lint{}
+	lint.Name = name
+	// run the linter on the spec
+	lintFile, err := lintFileForOpenAPI(name, root)
+	if err != nil {
+		return nil, err
+	}
+	lint.Files = append(lint.Files, lintFile)
+	return lint, err
```
seaneganx, line 59: Nit: This is equivalent to `return lint, nil` because of the error check above.

cmd/registry/core/lint-openapi.go
```diff
@@ -0,0 +1,157 @@
...
+	b, err := ioutil.ReadFile(root + "/linter.pb")
+	if err != nil {
+		return nil, err
+	}
+	output := &linter.Linter{}
+	if err := proto.Unmarshal(b, output); err != nil {
+		return nil, err
+	}
+
+	nodeFinder, err := newNodeFinder(root + "/" + path)
```
seaneganx, line 78: Should the loop below still execute even if this returns an error?

seaneganx (changes requested):

cmd/registry/core/lint-openapi.go
seaneganx: I may have misunderstood the docs. My impression was that the directory name would only have randomness if the asterisk is explicitly provided. If a random string is added to the end when no asterisk is provided this should be fine.

cmd/registry/core/lint-openapi.go
```diff
@@ -0,0 +1,154 @@
...
+					LineNumber:   l,
+					ColumnNumber: c + int32(len(node.Value)-1),
+				},
+			}
+		}
+		problems = append(problems, problem)
+	}
+	result := &rpc.LintFile{}
+	result.Problems = problems
+	return result, err
```
seaneganx, line 105: Is `err` ever non-nil now that the error check has been added above? If not, this could be `return result, nil` to remove ambiguity about the error.

seaneganx (approved):

cmd/registry/core/lint-openapi.go
timburks: right, thanks! changed

cmd/registry/core/lint-openapi.go
```diff
@@ -0,0 +1,154 @@
...
+				EndPosition: &rpc.LintPosition{
+					LineNumber:   l,
+					ColumnNumber: c + int32(len(node.Value)-1),
+				},
+			}
+		}
+		problems = append(problems, problem)
+	}
+	result := &rpc.LintFile{}
+	result.Problems = problems
```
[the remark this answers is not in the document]
timburks: yes, changed

PR #31, opened 2021-02-04
Add support to deploy on GKE

wangyh28:
This change contains helper functionalities to deploy the registry server to GKE. Changes include:
- Generalize `make build` so database is configurable while building the container image.
- Add scripts to create and expose services on GKE.
- Add scripts to setup authentication of local clients.
- Add instructions in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,28 @@ message like this:
 `rpc error: code = Unauthenticated desc = Unauthorized: HTTP status code 401`.
 To generate a new token, rerun `source auth/CLOUDRUN.sh`.
 
+## Running the Registry API server on GKE
+
+The [Makefile](Makefile) contains targets that build a Docker image
+(`make build`) and that deploy it to GKE (`make deploy-gke`).
+
+Requirements:
+
+- Ensure you have [gcloud](https://cloud.google.com/sdk/gcloud) and
+  [kubectl](https://cloud.google.com/kubernetes-engine/docs/quickstart)
+  installed.
+
+- If not already done, `gcloud auth login` gets user credentials for subsequent
+  `gcloud` operations and `gcloud config set project PROJECT_ID` can be used to
+  set your project ID to the one where you plan to host your servce.
+
+- The Makefile gets your project ID from the `REGISTRY_PROJECT_IDENTIFIER`
+  environment variable. This can be set automatically by running
+  `source auth/GKE.sh`.
+
+For detailed steps on how to deploy to GKE, please refer to
+[deployments/gke/README.md](deployments/gke/README.md).
+
 ## License
 
 This software is licensed under the Apache License, Version 2.0. See
--- a/deployments/gke/DEPLOY-TO-GKE.sh
+++ b/deployments/gke/DEPLOY-TO-GKE.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+#
+# Copyright 2021 Google LLC. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+if ! [ -x "$(command -v kubectl)" ] || ! [ -x "$(command -v gcloud)" ]; then
+  echo 'ERROR: This script requires `kubectl` and `gcloud`. Please install to continue.' >&2; return
+fi
+
+# Uses deployments/gke/service.yaml to create an external load balancer if no
+# service configuration file is specified.
+SERVICE_CONFIG=$1
+if [ -z "${SERVICE_CONFIG}" ]; then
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -256,6 +256,28 @@ message like this:
...
+  [kubectl](https://cloud.google.com/kubernetes-engine/docs/quickstart)
+  installed.
+
+- If not already done, `gcloud auth login` gets user credentials for subsequent
+  `gcloud` operations and `gcloud config set project PROJECT_ID` can be used to
+  set your project ID to the one where you plan to host your servce.
+
+- The Makefile gets your project ID from the `REGISTRY_PROJECT_IDENTIFIER`
+  environment variable. This can be set automatically by running
+  `source auth/GKE.sh`.
```
Krpate, line 276: Not sure how this can be best addressed but to do `make build` you need `REGISTRY_PROJECT_IDENTIFIER` which can be set by running `source auth/GKE.sh`. But in `GKE.sh`, we are trying to get `ingress _ip` and `service port` which requires the deployment to happen first and to deploy we need to build first.

Kind of a circular dependency. When I tried this in my local, instead of running `source auth/GKE.sh` to set the `REGISTRY_PROJECT_IDENTIFIER`, I just ran `export REGISTRY_PROJECT_IDENTIFIER=$(gcloud config list --format 'value(core.project)')`.

Krpate (changes requested):
This looks good to me. Just a couple of small suggestions.

deployments/gke/README.md
```diff
@@ -0,0 +1,49 @@
...
+   database configuration is
+   [config/registry.yaml](../../config/registry.yaml). You can specify a
+   different config file by using `DB_CONFIG`, e.g.
+   `make build DB_CONFIG=cloudsql-postgres` will use
+   [config/cloudsql-postgres.yaml](../../config/cloudsql-postgres.yaml)
+   instead.
+
+1. Run `make deploy-gke` to create a GKE deployment and expose it through an
+   external load balancer. To expose through an internal load balancer instead,
+   please uncomment the `annotations` section first before running
```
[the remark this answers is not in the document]
wangyh28: Done

deployments/gke/README.md
```diff
@@ -0,0 +1,49 @@
...
+   `make deploy-gke`.
+
+1. Setup the client authentication. This step differs based on the load
```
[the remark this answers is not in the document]
wangyh28: I personally use `1` as you don't have to change all numbers if you modify one of them, e.g. add a new number in the middle. The generated doc will have the right numbering.

deployments/gke/DEPLOY-TO-GKE.sh
```diff
@@ -0,0 +1,41 @@
...
+  gcloud container clusters create registry-backend --zone us-central1-a --scopes=datastore,sql-admin,gke-default
+fi
+
+# Authenticates to the cluster.
+gcloud container clusters get-credentials registry-backend --zone us-central1-a
+
+# Creates a deployment.
+envsubst < deployments/gke/deployment.yaml | kubectl apply -f -
+
+# Creates a service to expose the deployment. By default it creates a
```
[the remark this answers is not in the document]
wangyh28: Done

README.md
wangyh28: Yes I had the same question while setting up for CLOUDRUN and running auth/CLOUDRUN.sh while I didn't have any instance created. I feel this worth a separate CL to put `REGISTRY_PROJECT_IDENTIFIER=$(gcloud config list --format 'value(core.project)')` into the `makefile` itself, so the `makefile` is separated from `auth/.*.sh`. 

To temporarily solve this, I modified `auth/GKE.sh` so it returns while there's no GKE cluster exists or no `ingress_ip` available. But we do need a refactor on this

deployments/gke/README.md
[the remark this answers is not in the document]
Krpate: Ah, I see. Good to know this :)

README.md
Krpate: Thanks for addressing this. We should create a ticket to track the follow up required though.

timburks: Sometimes I've worked around this by running `auth/LOCAL.sh`, but realize now - that was just a hack.

deployments/gke/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: registry-backend
+#  Please uncomment the following lines to create an internal load balancer.
```
timburks, line 5: I'm unfamiliar with the pros and cons of this, but I'm guessing that it's not a problem to leave this commented out.

Could this be reworded to make that more clear? Maybe instead of "Please uncomment..." it might say "If you uncomment the following lines, an internal load balancer will be created." followed by a sentence explaining the consequences of that - I guess something like "the load balancer will distribute requests among nodes that automatically scale up with load" (if that's what will happen) and if it seems appropriate, we might link to https://cloud.google.com/load-balancing/docs/internal

timburks (changes requested):
looks great, I've just added a minor note requesting clarification

README.md
wangyh28: I created a new issue: https://github.com/apigee/registry/issues/35. I'll address separately

deployments/gke/service.yaml
wangyh28: Good catch! To make it less confusing, I created a separate config file for internal LB: deployments/gke/service-internal.yaml and added comments about internal LB on the top. Let me know if this makes thing more clear

timburks (approved):

PR #32, opened 2021-02-05
Refactor + Support revisions for Discovery bulk upload command

seaneganx:
The refactoring and functional changes are included in separate commits and might be easiest to review if viewed separately. Handling of semantic differences in JSON are also separated, and could possibly be removed if the Registry API is updated in the future to handle semantic change detection.


cmd/registry/cmd/upload-bulk-discovery.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/apigee/registry/rpc"
 	rpcpb "github.com/apigee/registry/rpc"
 	discovery "github.com/googleapis/gnostic/discovery"
+	"github.com/nsf/jsondiff"
```
timburks, line 27: Would this also show up in go.mod? Let's include at least that new line and discuss any other differences that seem worth including

timburks (changes requested):
Thanks! Let's add the new dependency to go.mod and go.sum.

cmd/registry/cmd/upload-bulk-discovery.go
seaneganx: Fixed. I just ran `make lite` and let it decide which changes to include.

timburks: Thanks, I see we picked up a few extra things - we can tighten up expectations about that going forward.

timburks (approved):

PR #34, opened 2021-02-08
Linter app

shrutiparabgoogle:
POC implementation for linter application:

Summary of changes:
* cmd/app-server: contains the server code and a Dockerfile to deploy this application.
* cmd/app-server/linterapp: contains the Handler which processes the requests, computes lint and push back into registry.
* the functionality from compute_lint.go is exported and reused in linterapp.

Currently this setup is verified for CloudRun deployments and openapi linters.


cmd/app-server/linterapp/linterapp.go
```diff
@@ -0,0 +1,109 @@
...
+        log.Print("Done")
+
+        lint_task := &cmd.ComputeLintTask{
+        Ctx: ctx,
+        Client: client,
+        SpecName: specName,
+        Linter: "",
+        }
+
+        err = lint_task.Run()
```
timburks, line 100: This task abstraction came from the registry tool's task runner. We might want to also use task queues here to allow this worker to run multiple lint jobs in parallel. I added task queues to the registry tool when I was working with large collections of specs (like the OpenAPI Directory) and found that processing them sequentially was very slow. To get some perspective, it might be useful to run this worker while adding lots of specs to the registry at once - such as by running the "registry bulk upload openapi" command from demos/atlas.sh.

timburks (commented):
Here are some general comments that we could discuss here and possibly move into a design doc:

I think it's great to have this as a first working example. As we go forward from here, I think we'll want to establish patterns that we would recommend users to follow. I can think of three, though there may be more:
1. Each worker is a single binary that does a single task in response to a pubsub message. These workers might use a configuration file, possibly built into the image, that specifies additional options like which linter to use or what linter rules should be used. 
2. We bundle many or all of our workers into a single binary that does multiple things in response to a pubsub message. This would also probably use a configuration file to configure these things.
3. We make the registry tool do this with an option. That would avoid the need to expose internals of the registry tool as an API for these other binaries.

On this spectrum, I think 1 is the easiest for customers to follow as a pattern and 3 is easiest for us to manage and test. Also, 1 makes more sense when we have to bundle other dependencies into the worker (e,g, external linters like spectral and api-linter).

We will also need to think about how workers report and handle errors.

cmd/app-server/linterapp/linterapp.go
```diff
@@ -0,0 +1,109 @@
...
+        return
+    }
+    message := rpc.Notification{}
+    if err := jsonpb.UnmarshalString(data, &message); err != nil {
+        log.Printf("json.Unmarshal: %v", err)
+        fmt.Fprintf(w, "Wrong message format")
+        return
+    }
+
+    if message.Change != rpc.Notification_DELETED {
```
seaneganx, line 71: Should this run for any notification types added in the future? If not, consider checking against specific notifications (e.g. CREATED, UPDATED).

It's also a common pattern ([example](https://golang.org/doc/effective_go.html#if)) to exit early on error/exit conditions rather than wrapping the regular flow in an if-statement. Something like:
```
switch type := message.Change {
    rpc.Notification_CREATED, rpc.Notification_UPDATED:
        log.Printf("running linter for change type %q", type)
    default:
        log.Printf("ignoring change type %q for linting", type)
        return
}
// code for processing CREATED, UPDATED types
```

shrutiparabgoogle:
@googlebot I fixed it.

cmd/app-server/linterapp/linterapp.go
shrutiparabgoogle: Thanks for this suggestion, much better approach. Done :)

cmd/app-server/linterapp/linterapp.go
```diff
@@ -0,0 +1,109 @@
...
+        idToken, err := metadata.Get(tokenURL)
+        if err != nil {
+                log.Printf("metadata.Get: failed to query id_token: %+v", err)
+                return "", err
+        }
+
+        return idToken, nil
+}
+
+func readPubsubMessage(w http.ResponseWriter, r *http.Request) string {
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

cmd/app-server/linterapp/linterapp.go
```diff
@@ -0,0 +1,109 @@
...
+        var m PubSubMessage
+        body, err := ioutil.ReadAll(r.Body)
+        if err != nil {
+                log.Printf("ioutil.ReadAll: %v", err)
+                http.Error(w, "Bad Request", http.StatusBadRequest)
+                return ""
+        }
+        if err := json.Unmarshal(body, &m); err != nil {
+                log.Printf("json.Unmarshal: %v", err)
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

shrutiparabgoogle:
Will create a new PR for the worker architecture.

PR #36, opened 2021-02-12
Optionally allow missing specs on UpdateSpec method

seaneganx:


server/actions_specs.go
```diff
@@ -38,53 +38,66 @@ func (s *RegistryServer) CreateSpec(ctx context.Context, request *rpc.CreateSpec
...
+	k := client.NewKey(models.SpecEntityName, spec.ResourceNameWithRevision())
+	if _, err := client.Put(ctx, k, spec); err != nil {
+		log.Printf("save spec error %+v", err)
+		return err
+	}
+
+	return nil
+}
+
+func saveBlob(ctx context.Context, client storage.Client, spec *models.Spec, contents []byte) error {
```
timburks, line 95: I'm a bit paranoid about name collisions - since we also save blobs for properties/artifacts, could we include "Spec" in this function name? Something like "saveSpecBlob" or "saveBlobForSpec"?

timburks (changes requested):
Looks good - see one question about a name below. I'm guessing we'll do this same change for Projects/Apis/Versions once this pattern is proven. What are your thoughts about testing?

server/actions_specs.go
seaneganx: Good idea, I didn't realize this file was part of a much wider shared package.

seaneganx:
> What are your thoughts about testing?

I've been following along with the status quo which seems to be mostly manual testing with some very high level functional tests. There could be value in developing an automated suite to cover at least critical API functionality, or all supported functionality.

We'll have to discuss this outside the current PR because making this easily testable (alongside the rest of the API) would be a decent sized story.

timburks (approved):

PR #37, opened 2021-02-16
registry/v1 protos

timburks:
Add protos for the registry/v1 API.


timburks:
instead of doin g this, let's start a v1 branch in the main repo

PR #49, opened 2021-02-22
V1 labels

timburks:
Initial support for new labels and annotations

### DIFF
--- a/google/cloud/apigee/registry/v1/registry_internal.proto
+++ b/google/cloud/apigee/registry/v1/registry_internal.proto
@@ -1,0 +1,41 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package google.cloud.apigee.registry.v1;
+
+option java_package = "com.google.cloud.apigee.registry.v1";
+option java_multiple_files = true;
+option java_outer_classname = "RegistryInternalProto";
+option go_package = "github.com/apigee/registry/rpc;rpc";
+
+// OrderedPair represents a name-value pair in which both are strings.
+// (-- api-linter: core::0123::resource-annotation=disabled
+//     aip.dev/not-precedent: This message is not currently used in an API. --)
+message OrderedPair {
+  // The name of the pair.
+  string name = 1;
+
+  // The value of the pair.
+  string value = 2;
+}
+
+// OrderedMap represents a map of strings with ordered keys.
+// (-- api-linter: core::0123::resource-annotation=disabled
+//     aip.dev/not-precedent: This message is not currently used in an API. --)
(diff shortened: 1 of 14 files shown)


google/cloud/apigee/registry/v1/registry_internal.proto
```diff
@@ -0,0 +1,41 @@
...
+  string name = 1;
+
+  // The value of the pair.
+  string value = 2;
+}
+
+// OrderedMap represents a map of strings with ordered keys.
+// (-- api-linter: core::0123::resource-annotation=disabled
+//     aip.dev/not-precedent: This message is not currently used in an API. --)
+message OrderedMap {
```
seaneganx, line 38: What's the purpose of ordering this for serialization? We don't seem to use the ordering anywhere since the blobs are deserialized into an unordered map. Using proto maps instead of custom types seems like a more direct approach.

google/cloud/apigee/registry/v1/registry_internal.proto
```diff
@@ -0,0 +1,41 @@
...
+
+option java_package = "com.google.cloud.apigee.registry.v1";
+option java_multiple_files = true;
+option java_outer_classname = "RegistryInternalProto";
+option go_package = "github.com/apigee/registry/rpc;rpc";
+
+// OrderedPair represents a name-value pair in which both are strings.
+// (-- api-linter: core::0123::resource-annotation=disabled
+//     aip.dev/not-precedent: This message is not currently used in an API. --)
+message OrderedPair {
```
seaneganx, line 27: Is this called OrderedPair because it's used in the OrderedMap? Having this message nested inside the OrderedMap message and named Pair or Entry seems more appropriate if it exists specifically for that use case.

seaneganx (commented):
Since the API allows filtering according to these labels I think we should consider an approach where they are stored without blobification so they can be indexed by the database. Were there any issues with that approach that made you decide on using blobs?

google/cloud/apigee/registry/v1/registry_internal.proto
timburks: I was thinking that a stable serialization would be much easier to compare, if we needed to do equality testing.

google/cloud/apigee/registry/v1/registry_internal.proto
timburks: Yes, and I agree - will change.

google/cloud/apigee/registry/v1/registry_internal.proto
timburks: to your top-level question, storing these as a blob puts fewer demands on the database, and at least for now (early on) we're relying on a simple connection that allows multiple backends - i.e. this is an approach that I think will meet our initial needs and that we can smoothly upgrade if needed.

google/cloud/apigee/registry/v1/registry_internal.proto
timburks: Deleted in the latest commit.

google/cloud/apigee/registry/v1/registry_internal.proto
seaneganx: I'm having trouble imagining when we'll need to compare the serialized blobs for equality, but if we need it in the future we can add it.

Makes sense that it's simpler to implement blob storage across multiple databases. I'm just concerned this storage format will limit us from some API functionality now or in the near future. Specifically listing with filters/ordering could be affected.

seaneganx (approved):

google/cloud/apigee/registry/v1/registry_internal.proto
timburks: My strongest supporting argument for that was making database records easier to compare manually (mainly for debugging), but agree that we can leave it out for now.

PR #56, opened 2021-02-25
Include RPC retry policy on all methods

seaneganx:


gapic/grpc_service_config.json
```diff
@@ -3,34 +3,16 @@
...
-            "method": "CreateVersion"
-          },
-          {
-            "service": "google.cloud.apigee.registry.v1alpha1.Registry",
-            "method": "GetSpec"
-          },
-          {
-            "service": "google.cloud.apigee.registry.v1alpha1.Registry",
-            "method": "CreateSpec"
+            "service": "google.cloud.apigee.registry.v1alpha1.Registry"
```
timburks, line 9: I think we can drop this. In this branch (and following), the gapics are generated for the v1 API only.

timburks (approved):
Just one minor comment that we can address here or later.

gapic/grpc_service_config.json
seaneganx: Makes sense. I'll leave it in for now as we still have the v1alpha1 protos, but it can be removed at any time. Even if we forget to remove it in the future it won't cause issues by having rules on a nonexistent service.

PR #58, opened 2021-02-25
Add label-based filtering to apis, versions, and specs.

timburks:
The first commit is for APIs only. Let's give that an initial review before adding version and spec filtering.

### DIFF
--- a/server/models/api.go
+++ b/server/models/api.go
@@ -138,3 +138,8 @@ func (api *Api) Update(message *rpc.Api, mask *fieldmaskpb.FieldMask) error {
 	api.UpdateTime = time.Now()
 	return nil
 }
+
+// LabelsMap returns a map representation of stored labels.
+func (api *Api) LabelsMap() (map[string]string, error) {
+	return mapForBytes(api.Labels)
+}
--- a/tests/filters_test.go
+++ b/tests/filters_test.go
@@ -114,9 +114,9 @@ func TestFilters(t *testing.T) {
 	// test some filters against their expected return counts
 	// see https://github.com/google/cel-spec/blob/master/doc/langdef.md#field-selection for CEL details.
 	pairs := []struct {
-		string // the filter pattern
-		int    // number of expected responses
-		bool   // true if we expect an error
+		filter        string // the filter pattern
+		expectedCount int    // number of expected responses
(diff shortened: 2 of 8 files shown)


tests/filters_test.go
```diff
@@ -0,0 +1,170 @@
...
+		return false
+	}
+	st, ok := status.FromError(err)
+	if !ok {
+		return false
+	}
+	return st.Code() == codes.Unavailable
+}
+
+func check(t *testing.T, message string, err error) {
```
seaneganx, line 41: Can this be checked in-line?

The check for unavailable doesn't seem to add much value compared to the actual error: `rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp [::1]:8080: connect: connection refused"`. If desired we could do one availability check at the start of the tests to make it explicitly clear that a local server should be running, but I think the message is quite clear.

Assertion helpers like these are used like domain-specific languages for test assertions, with drawbacks such as the indirection preventing linters from telling us if the message string we provide isn't valid. This also makes errors look like they occur at line 47 regardless of where the check was called (this can be fixed though).

server/models/api.go
```diff
@@ -138,3 +138,8 @@ func (api *Api) Update(message *rpc.Api, mask *fieldmaskpb.FieldMask) error {
 	api.UpdateTime = time.Now()
 	return nil
 }
+
+// LabelsMap returns a map representation of stored labels.
+func (api *Api) LabelsMap() (map[string]string, error) {
+	return mapForBytes(api.Labels)
+}
```
seaneganx, line 145: No change necessary, just thinking out loud. As more fields appear with serialized storage formats it seems increasingly appealing to have a data access package that insulates the RPC handlers from handling all the storage logic, making internal types usable immediately after retrieval.

tests/filters_test.go
timburks: This was reused from elsewhere - I think it would make sense to tighten this up and put it somewhere common, how about in a separate PR focused on test architecture?

server/models/api.go
timburks: To some extent the "models" package helps with this, but further improvement sgtm.

seaneganx (approved):

Issue #43, opened 2021-02-18
Registry v1 API: update hashing of ApiSpec and Artifact contents from SHA-1 to SHA-256


timburks:

timburks:
Addressed in #60

PR #59, opened 2021-02-25
V1 hashing

timburks:
Addresses #43 by updating hashing from SHA-1 to SHA-256 and ensuring that specs and artifacts returned by the API have correct hash and size values.


timburks:
Incorrect target branch

PR #63, opened 2021-02-27
Fix a bug and an inconsistency in the "registry get" command

timburks:
With this, "registry get" for specs and artifacts will print the message body by default and, if the --contents flag is provided, the contents of the spec or artifact.

### DIFF
--- a/cmd/registry/cmd/compute-details.go
+++ b/cmd/registry/cmd/compute-details.go
@@ -176,7 +176,7 @@ func (task *computeDetailsTask) Run() error {
 				Description: description,
 			},
 			UpdateMask: &field_mask.FieldMask{
-				Paths: []string{"owner", "display_name", "description"},
+				Paths: []string{"display_name", "description"},
 			},
 		}
 
--- a/cmd/registry/cmd/compute-lint.go
+++ b/cmd/registry/cmd/compute-lint.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"strings"
 
 	"github.com/apigee/registry/cmd/registry/core"
 	"github.com/apigee/registry/connection"
@@ -101,7 +100,7 @@ func (task *computeLintTask) Run() error {
 	}
 	var relation string
 	var lint *rpc.Lint
-	if strings.HasPrefix(spec.GetMimeType(), "openapi") {
+	if core.IsOpenAPIv2(spec.GetMimeType()) || core.IsOpenAPIv3(spec.GetMimeType()) {
 		// the default openapi linter is gnostic
 		if task.linter == "" {
 			task.linter = "gnostic"
--- a/cmd/registry/cmd/export-sheet.go
+++ b/cmd/registry/cmd/export-sheet.go
@@ -94,7 +94,7 @@ var exportSheetCmd = &cobra.Command{
 			saveSheetPath(ctx, client, path, sheetArtifactName)
 		} else if messageType == "gnostic.metrics.Complexity" {
 			path, err = core.ExportComplexityToSheet("Complexity", inputs)
-			log.Printf("exported complexity %+v to %s", inputs, path)
+			log.Printf("exported complexity to %s", path)
 			saveSheetPath(ctx, client, path, sheetArtifactName)
 		} else if messageType == "google.cloud.apigee.registry.applications.v1alpha1.Index" {
 			if len(inputs) != 1 {
--- a/cmd/registry/cmd/get.go
+++ b/cmd/registry/cmd/get.go
@@ -61,7 +61,11 @@ var getCmd = &cobra.Command{
 				_, err = core.GetSpec(ctx, client, m, getContents, core.PrintSpecDetail)
 			}
 		} else if m := names.ArtifactRegexp().FindStringSubmatch(name); m != nil {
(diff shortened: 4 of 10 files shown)


demos/protos.sh
```diff
@@ -73,15 +73,15 @@ registry list projects/protos/apis/-/versions/-/specs
 registry get projects/protos/apis/google-cloud-translate/versions/v3/specs/protos.zip
 
 # You can also get this with the automatically-generated `apg` command line tool:
-apg registry get-spec --name projects/protos/apis/google-cloud-translate/versions/v3/specs/protos.zip
+apg registry get-api-spec --name projects/protos/apis/google-cloud-translate/versions/v3/specs/protos.zip
```
timburks, line 76: These command names changed in the v1 API

cmd/registry/cmd/compute-details.go
```diff
@@ -176,7 +176,7 @@ func (task *computeDetailsTask) Run() error {
 				Description: description,
 			},
 			UpdateMask: &field_mask.FieldMask{
-				Paths: []string{"owner", "display_name", "description"},
+				Paths: []string{"display_name", "description"},
```
timburks, line 179: The "owner" field was removed from the v1 API - its definition was unclear and it can be represented by a label or annotation. Removing it here and in a few other places fixed a potential crash due to conflicts with the database schema.

cmd/registry/cmd/compute-lint.go
```diff
@@ -101,7 +100,7 @@ func (task *computeLintTask) Run() error {
 	}
 	var relation string
 	var lint *rpc.Lint
-	if strings.HasPrefix(spec.GetMimeType(), "openapi") {
+	if core.IsOpenAPIv2(spec.GetMimeType()) || core.IsOpenAPIv3(spec.GetMimeType()) {
```
timburks, line 103: An earlier update missed this change to the way we recognize API types.

cmd/registry/cmd/export-sheet.go
```diff
@@ -94,7 +94,7 @@ var exportSheetCmd = &cobra.Command{
 			saveSheetPath(ctx, client, path, sheetArtifactName)
 		} else if messageType == "gnostic.metrics.Complexity" {
 			path, err = core.ExportComplexityToSheet("Complexity", inputs)
-			log.Printf("exported complexity %+v to %s", inputs, path)
+			log.Printf("exported complexity to %s", path)
```
timburks, line 97: The +v output of inputs was super-verbose and didn't add any value.

cmd/registry/cmd/get.go
```diff
@@ -61,7 +61,11 @@ var getCmd = &cobra.Command{
 				_, err = core.GetSpec(ctx, client, m, getContents, core.PrintSpecDetail)
 			}
 		} else if m := names.ArtifactRegexp().FindStringSubmatch(name); m != nil {
-			_, err = core.GetArtifact(ctx, client, m, true, core.PrintArtifactDetail)
+			if getContents {
+				_, err = core.GetArtifact(ctx, client, m, getContents, core.PrintArtifactContents)
+			} else {
+				_, err = core.GetArtifact(ctx, client, m, getContents, core.PrintArtifactDetail)
+			}
```
timburks, line 68: This lets "registry get" display both the Artifact message and its contents, depending on whether the --contents flag is included.

cmd/registry/core/print.go
```diff
@@ -61,7 +61,7 @@ func PrintSpecDetail(message *rpc.ApiSpec) {
 
 func PrintSpecContents(message *rpc.ApiSpec) {
 	contents := message.GetContents()
-	if strings.HasSuffix(message.GetMimeType(), "+gzip") {
+	if strings.Contains(message.GetMimeType(), "+gzip") {
```
timburks, line 64: This is another fix to the way we recognize content types (this can be improved further in the future)

demos/disco.sh
```diff
@@ -67,15 +67,15 @@ registry list projects/disco/apis/-/versions/-/specs
 registry get projects/disco/apis/translate/versions/v3/specs/discovery.json
 
 # You can also get this with the automatically-generated `apg` command line tool:
-apg registry get-spec --name projects/disco/apis/translate/versions/v3/specs/discovery.json
+apg registry get-api-spec --name projects/disco/apis/translate/versions/v3/specs/discovery.json
```
timburks, line 70: These command names changed in the v1 API.

demos/protos.sh
```diff
@@ -136,3 +136,11 @@ registry export sheet projects/protos/artifacts/vocabulary
 # track changes across versions, and find unique terms in APIs that we are reviewing.
 # By storing these results and other artifacts in the Registry, we can build a
 # centralized store of API information that can help manage an API program.
+
+# We can also run analysis tools like linters and store the results in the Registry.
+# Here we run the Google api-linter and compile summary statistics.
+registry compute lint projects/protos/apis/-/versions/-/specs/-
+registry compute lintstats projects/protos/apis/-/versions/-/specs/- --linter aip
+registry compute lintstats projects/protos --linter aip
+
```
timburks, line 145: This adds linting to the protos demo. Protobuf linting requires the api-linter, which is usually available because it is also used in the build process.

seaneganx (approved):

PR #65, opened 2021-03-01
Bump Go version to 1.15 or higher

seaneganx:
* Include DB config in Makefile cleanup


Makefile
```diff
@@ -16,7 +16,8 @@ clean:
 	rm -rf \
 		cmd/apg/*.go gapic/*.go rpc/*.go \
 		third_party/api-common-protos third_party/gnostic \
-		envoy/proto.pb
+		envoy/proto.pb \
+		/tmp/registry.db
```
timburks, line 20: This fixes test instability, but I'm concerned that it would be an unpleasant surprise to someone who's building a registry in a local sqlite database. Perhaps we could instead have a dedicated config YAML for testing that uses a different sqlite .db that we delete during the tests (which seems big enough for a dedicated PR)

timburks (changes requested):

Makefile
seaneganx: Good point. This can be addressed later through other improvements to test isolation.

timburks (approved):

Issue #66, opened 2021-03-01
Documentation for custom ID validation does not match behavior


seaneganx:
Currently the implementation validates custom IDs according to the regex `([a-zA-Z0-9-_\\.]+)`. The API documentation says the following:
> This value should be 4-63 characters, and valid characters are /[a-z][0-9]-/."

[AIP-122](https://aip.dev/122#resource-id-segments) suggests custom IDs follow [RFC-1034](https://tools.ietf.org/html/rfc1034). The RFC mentions the following:

> They must start with a letter, end with a letter or digit, and have as interior characters only letters, digits, and hyphen. There are also some restrictions on the length.  Labels must be 63 characters or less."

The demo scripts currently create resources with names like "1.0.0" so we can't follow the AIP suggestion without modifying the demos. Also, the tests in `filter_test.go` create APIs with single digit IDs. There may be other dependencies on the current validation being more relaxed than the documentation.

seaneganx:
This issue also allows for case sensitive resource IDs, e.g. `projects/abc`, `projects/Abc`, `projects/aBc`, and so on can all be created as separate resources provided they have different case.

seaneganx:
Per discussion with @timburks the limitations will be slightly different than RFC-1034. Specifically, the API validation will differ from RFC-1034 by allowing numbers as the first character in IDs and by allowing `.` in addition to hyphen as interior characters in an ID.

PR #67, opened 2021-03-01
Refactor names package + Generate IDs when unspecified

seaneganx:
Originally I intended on adding validation changes to this PR to make our implementation match the documented proto behaviors. That change turned out to be difficult and high impact so it will need its own PR at a later time. I created an issue for it [here](https://github.com/apigee/registry/issues/66).


server/names/common.go
```diff
@@ -17,31 +17,29 @@ package names
...
-func ValidateID(id string) error {
-	r := regexp.MustCompile("^" + NameRegex + "$")
-	m := r.FindStringSubmatch(id)
-	if m == nil {
-		return fmt.Errorf("invalid id '%s'", id)
-	}
-	return nil
+// GenerateID generates a random resource ID.
+func GenerateID() string {
+	return uuid.New().String()[:8]
```
timburks, line 34: What happens if this produces an ID that's already in use? Given the low likelihood, I think it would be OK if the request failed and had to be retried but not OK if the request proceeded and an existing resource was clobbered. Could we verify this behavior by testing with shorter IDs?

timburks: We might retry ourselves (in the API request hander) since we know when we are using auto-generated IDs.

timburks (approved):
I had a comment/suggestion but I think it's outside the scope of this PR, so let's leave it for a followup issue.

server/names/common.go
seaneganx: That depends on the implementation. Currently it results in an "already exists" error which might be confusing in the unlikely case where it happens. Implementing retry would eliminate the issue.

Writing an automated test for this would require some refactoring to allow injection of a custom randomness source at server create time. Let me know if you want me to proceed with that.

timburks: Let's discuss it on PR#70 - I don't think this is an immediate priority but we'll probably want to eventually address it

PR #69, opened 2021-03-02
Projects testing + Handler updates

seaneganx:
Some tests are marked as skipped until larger changes are submitted to make them pass.

### DIFF
--- a/server/actions_projects.go
+++ b/server/actions_projects.go
@@ -110,10 +110,15 @@ func (s *RegistryServer) ListProjects(ctx context.Context, req *rpc.ListProjects
 		return nil, unavailableError(err)
 	}
 	defer s.releaseStorageClient(client)
+
+	if req.GetPageSize() < 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid page_size: must not be negative")
+	}
+
 	q := client.NewQuery(models.ProjectEntityName)
 	q, err = q.ApplyCursor(req.GetPageToken())
 	if err != nil {
-		return nil, internalError(err)
+		return nil, invalidArgumentError(err)
 	}
 	prg, err := createFilterOperator(req.GetFilter(),
 		[]filterArg{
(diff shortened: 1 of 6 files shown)


server/actions_projects.go
```diff
@@ -110,21 +115,27 @@ func (s *RegistryServer) ListProjects(ctx context.Context, req *rpc.ListProjects
...
+
+	if req.GetPageSize() < 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid page_size: must not be negative")
+	}
+
 	q := client.NewQuery(models.ProjectEntityName)
 	q, err = q.ApplyCursor(req.GetPageToken())
 	if err != nil {
-		return nil, internalError(err)
+		return nil, invalidArgumentError(err)
```
seaneganx, line 126: Is there any reason an error would occur while applying the cursor for a valid token?

server/actions_projects.go
```diff
@@ -110,21 +115,27 @@ func (s *RegistryServer) ListProjects(ctx context.Context, req *rpc.ListProjects
...
 		[]filterArg{
+			{"name", filterArgTypeString},
 			{"project_id", filterArgTypeString},
 			{"display_name", filterArgTypeString},
 			{"description", filterArgTypeString},
 			{"create_time", filterArgTypeTimestamp},
 			{"update_time", filterArgTypeTimestamp},
 		})
 	if err != nil {
-		return nil, internalError(err)
```
seaneganx, line 127: `createFilterOperator` already sets the status code for these errors, so I think it can be trusted right now. Moving forward it may be a good idea to only set response statuses in the RPC handlers, using other techniques for returning multiple types of errors internally.

timburks (changes requested):
Looks good. A couple of unit tests are broken by the change to delete methods - generally they just need to ignore errors from initial calls to DeleteProject - could you update those before merging?

server/actions_projects.go
seaneganx: Agreed. Cases where the cursor becomes stale seem to be more like a failed precondition or aborted operation anyway. We would want to indicate that the entire sequence should be restarted.

timburks (approved):

Issue #52, opened 2021-02-23
Generated apg client cannot be used with Envoy over insecure connections


seaneganx:
The `apg` tool doesn't forward credentials to insecure services, so the locally running authz-server is unable to receive the credentials needed to authorize a request. Either the `apg` tool should be updated (may be difficult) to forward the credentials, or a workaround should be made to enable use of `apg` with Envoy.

seaneganx:
One potential solution is to add a flag in the authz-server config to permit requests without authorization headers. This would allow Envoy to be used locally for HTTP traffic in cases where authz isn't desired.

timburks:
#72 includes an addition to tools/COMPILE-PROTOS.sh that patches the generated GAPIC with code that ensures that tokens are always sent. That's a bad (fragile) solution but seemed the lesser of multiple evils - better than having calls silently fail until the developer realizes that tokens aren't being sent over non-SSL connections (a silent "feature" of the GAPICs).

timburks:
cf. https://github.com/googleapis/gapic-generator-go/issues/542

timburks:
This concern also applies to the `registry` CLI and any other GAPIC-based clients (but all are addressed by the GAPIC patch described above)

timburks:
#146 adds a check to COMPILE-PROTOS.sh that will cause make builds to fail if the patch is not applied.

timburks:
Now that we can test transcoding with no-auth Envoy configurations (https://github.com/apigee/registry/pull/223), I think this is unlikely to be a problem for anyone and that we can probably remove the patch.

PR #72, opened 2021-03-03
Add support for arm64 containers

timburks:
This updates the Dockerfile and run scripts to allow the Registry server to be run in containers locally and in Kubernetes installations on both x86 and arm64 architectures. New README instructions document the process of building and running containers locally. Arm64 support required updating to Envoy 1.16 (which has published images for both x86 and arm64) and some small configuration changes related to the version change.

### DIFF
--- a/deployments/container/Dockerfile.envoy
+++ b/deployments/container/Dockerfile.envoy
@@ -5,9 +5,13 @@
 # This is based on Debian and sets the GOPATH to /go.
 # https://hub.docker.com/_/golang
 FROM golang:1.15 as builder
+
+# use aarch_64 for arm64
+ARG ARCH=x86_64
+
 RUN apt-get update
 RUN apt-get install unzip
-RUN curl -L https://github.com/protocolbuffers/protobuf/releases/download/v3.13.0/protoc-3.13.0-linux-x86_64.zip > /tmp/protoc.zip
+RUN curl -L https://github.com/protocolbuffers/protobuf/releases/download/v3.15.1/protoc-3.15.1-linux-$ARCH.zip > /tmp/protoc.zip
 RUN unzip /tmp/protoc.zip -d /usr/local
 
 # Create and change to the app directory.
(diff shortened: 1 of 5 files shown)


deployments/container/Dockerfile.envoy
```diff
@@ -31,13 +35,13 @@ RUN CGO_ENABLED=0 GOOS=linux go build -v -o registry-server ./cmd/registry-serve
...
 # Use an Envoy release image to get envoy in the image.
-FROM envoyproxy/envoy:v1.14.4
+FROM envoyproxy/envoy:v1.16.0
 
 ARG DB_CONFIG=registry
 
 COPY deployments/container/RUN-WITH-ENVOY.sh /RUN-WITH-ENVOY.sh
 COPY deployments/envoy/envoy.yaml /etc/envoy/envoy.yaml
-COPY deployments/envoy/proto.pb /proto.pb
+COPY --from=builder /app/deployments/envoy/proto.pb /proto.pb
```
timburks, line 44: Previously this was copying proto.pb from the local build. This gets it from the container build, which allows containers to be built from otherwise-clean checkouts.

tools/COMPILE-PROTOS.sh
```diff
@@ -79,6 +79,11 @@ func (c *RegistryClient) GrpcClient() rpcpb.RegistryClient {
 }
 END
 
+# patch the generated GAPIC to send Authorization tokens with insecure requests
+# (this allows the registry command line tools to test container builds)
+sed -i 's/return metadata.NewOutgoingContext(ctx, out)/insecure := os.Getenv("APG_REGISTRY_INSECURE")\ntoken := os.Getenv("APG_REGISTRY_TOKEN")\nif insecure == "1" \&\& token != "" \{ \nout["authorization"] = append(out["authorization"], "Bearer "+token) \n\}\nreturn metadata.NewOutgoingContext(ctx, out)/' gapic/doc.go
```
seaneganx, line 84: If this stops working will CI fail? My concern is that the generated code may change in the future, and whoever happens to bump the version won't manually verify this still works.

seaneganx (approved):
One concern to consider.

tools/COMPILE-PROTOS.sh
timburks: Good point. If this breaks compilation might fail but it's not guaranteed - or this could just silently not happen and stop working. I'll update #52 to note that this was done. Perhaps we should keep that open until we have a better solution?

seaneganx: That seems like a reasonable approach 👍

PR #73, opened 2021-03-04
APIs Testing + Updates

seaneganx:


server/actions_apis.go
```diff
@@ -174,10 +181,18 @@ func (s *RegistryServer) ListApis(ctx context.Context, req *rpc.ListApisRequest)
...
 	}
-	responses.NextPageToken, err = it.GetCursor(len(apiMessages))
+
+	nextToken, err := it.GetCursor(len(apiMessages))
 	if err != nil {
 		return nil, internalError(err)
 	}
+
+	if _, err := it.Next(&api); err == nil {
+		responses.NextPageToken = nextToken
```
timburks, line 199: This change (and the one already made to projects) breaks paginated iteration when filters are used. This only happens for gorm-based queries, and occurs because the gorm queries use a fixed page size (50) to limit the number of records read from the database before applying filters. For large collections, this can cause some pages of a filtered query to return empty arrays, which is allowed by AIP-158 ("The API may return fewer results than the number requested (including zero results), even if not at the end of the collection.")

I agree that we should find a way to return empty page tokens when the end of a collection is reached, but rather than arbitrarily increasing gorm query size or including another layer of iteration in these queries (both of these things could potentially make queries arbitrarily expensive), we might instead look at the iterator returned by client.Run() and (since it is unfiltered), detect the end of the collection by recognizing when that iterator returns nothing.

server/actions_apis.go
```diff
@@ -66,17 +66,25 @@ func (s *RegistryServer) DeleteApi(ctx context.Context, req *rpc.DeleteApiReques
 		return nil, unavailableError(err)
 	}
 	defer s.releaseStorageClient(client)
-	// Validate name and create dummy api (we just need the ID fields).
+
```
timburks, line 69: What's your criteria for deleting comments? This seems like an unintuitive use of NewApiFromResourceName() that I thought was worth explaining.

server/actions_apis.go
timburks: (sorry this wasn't covered by existing tests - I can imagine testing this by creating a largish collection and using filters to select items from the end of the collection, but that also depends on knowing the order of iteration, which we currently do when we specify ids in create calls)

timburks (changes requested):

server/actions_apis.go
seaneganx: It wasn't intentionally removed, I should pay more attention to my diffs. Fixed.

server/actions_apis.go
```diff
@@ -66,17 +66,25 @@ func (s *RegistryServer) DeleteApi(ctx context.Context, req *rpc.DeleteApiReques
...
+		return nil, notFoundError(err)
+	} else if err != nil {
+		return nil, internalError(err)
+	}
+
+	if err := client.DeleteChildrenOfApi(ctx, api); err != nil {
+		return nil, internalError(err)
+	}
+
 	s.notify(rpc.Notification_DELETED, req.GetName())
```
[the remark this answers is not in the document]
seaneganx: Added an error check before sending the notification. It wasn't a style guide issue, I just wanted to reorder it to be more direct about the notification being sent regardless of the error.

seaneganx (commented):
PTAL. I'm making a different PR for modifications to the ListProjects method and tests.

timburks (approved):

PR #76, opened 2021-03-05
Versions Testing + Updates

seaneganx:
* Support filtering by version resource name
* Return NOT_FOUND when deleting nonexistent resources
* Return INVALID_ARGUMENT on negative page size
* Return INVALID_ARGUMENT on invalid page token
* Add server tests for version actions


timburks (approved):
Looks great! The patterns are becoming clear - looking ahead it might help to have a high-level summary of everything that is being tested, maybe as a README or wiki page.

PR #77, opened 2021-03-05
Specs Testing + Handler Updates

seaneganx:
This change doesn't include all the testing needed for the specs API. It just covers the main CRUD operations as we've covered them for projects/apis/versions. Some examples of behaviors that still need testing:
- Revisions
- Tagging
- UpdateSpec with allow_missing

### DIFF
--- a/server/gorm/deletions.go
+++ b/server/gorm/deletions.go
@@ -61,7 +61,7 @@ func (c *Client) DeleteChildrenOfProject(ctx context.Context, project *models.Pr
 		q = q.Require("ProjectID", project.ProjectID)
 		err := c.DeleteAllMatches(ctx, q)
 		if err != nil {
-			//return err
+			return err
 		}
 	}
 	return nil
@@ -79,7 +79,7 @@ func (c *Client) DeleteChildrenOfApi(ctx context.Context, api *models.Api) error
 		q = q.Require("ApiID", api.ApiID)
 		err := c.DeleteAllMatches(ctx, q)
 		if err != nil {
-			//return err
+			return err
 		}
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


server/gorm/deletions.go
```diff
@@ -61,7 +61,7 @@ func (c *Client) DeleteChildrenOfProject(ctx context.Context, project *models.Pr
 		q = q.Require("ProjectID", project.ProjectID)
 		err := c.DeleteAllMatches(ctx, q)
 		if err != nil {
-			//return err
+			return err
```
seaneganx, line 64: If the intention is to ignore errors when we're deleting children of resources, I think it would be more appropriate for the caller of this function to make that decision. For now I have the errors going back in the response because it appears to be consistent with the datastore client.

timburks: I don't recall why I did this, so I think it's fine to always return errors until we see a problem caused by that.

server/gorm/deletions.go
```diff
@@ -97,8 +97,18 @@ func (c *Client) DeleteChildrenOfVersion(ctx context.Context, version *models.Ve
...
 		}
 	}
 	return nil
 }
+
+// DeleteChildrenOfSpec deletes all the children of a spec.
+func (c *Client) DeleteChildrenOfSpec(ctx context.Context, spec *models.Spec) error {
+	q := c.NewQuery(models.BlobEntityName)
```
timburks, line 108: This should probably also delete artifacts associated with the spec. I'll create an issue that we can address later (same for APIs and versions).

timburks (approved):
Looks great! Two general comments below:

server/gorm/deletions.go
seaneganx: 👍

PR #84, opened 2021-03-10
Refactor API service

seaneganx:
* Add type for API resource names, refactor interfaces where applicable.
* Move API storage entity name into storage package.
* Refactor common API operations into helper functions.
* Check additional errors before returning
* Consistently use error helper functions.
* Remove redundant import names.

### DIFF
--- a/server/models/api.go
+++ b/server/models/api.go
@@ -66,38 +66,38 @@ func NewApi(name names.Api, body *rpc.Api) (api *Api, err error) {
 	return api, nil
 }
 
-// ResourceName generates the resource name of a api.
-func (a *Api) ResourceName() string {
-	return fmt.Sprintf("projects/%s/apis/%s", a.ProjectID, a.ApiID)
+// Name returns the resource name of the api.
+func (api *Api) Name() string {
+	return fmt.Sprintf("projects/%s/apis/%s", api.ProjectID, api.ApiID)
 }
 
 // Message returns a message representing an api.
-func (a *Api) Message(view rpc.View) (message *rpc.Api, err error) {
+func (api *Api) Message(view rpc.View) (message *rpc.Api, err error) {
 	message = &rpc.Api{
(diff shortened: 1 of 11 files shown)


server/models/api.go
```diff
@@ -42,103 +39,116 @@ type Api struct {
...
-		ProjectID: m[1],
-		ApiID:     m[2],
-	}, nil
+	return api, nil
 }
 
 // ResourceName generates the resource name of a api.
-func (api *Api) ResourceName() string {
-	return fmt.Sprintf("projects/%s/apis/%s", api.ProjectID, api.ApiID)
+func (a *Api) ResourceName() string {
```
timburks, line 70: "a *Api" looks weird to me - I know the Go preference for short variable names, but is it required? It seems inconsistent to so often use "api" as the name for *Api variables and to not do it on these method receivers. "api" is already an abbreviation, so this is abbreviating an abbreviation, which (I just think) is overly terse, so how strict is the style guide on this?

seaneganx: That's fair. In this case it's less about sticking to short variable names in general and more about receiver names in particular. It's a very common pattern for receiver names to be single characters, so I copied that style here.

Out of curiosity I picked some packages from the standard library. [Most](https://golang.org/pkg/bytes/) [seemed](https://golang.org/pkg/io/) to [stick](https://golang.org/pkg/os/) to [single](https://golang.org/pkg/testing/) [characters](https://golang.org/pkg/time/), but notably I found [regexp](https://golang.org/pkg/regexp/) using `re` and [net/http](https://golang.org/pkg/net/http/) using `srv` and `mux`.

My primary complaint about using `api` for the receiver name is that it reserves that valuable name when the same information (i.e. that we're acting on a receiver) can be conveyed by adopting the common pattern. We can go either way though.

timburks (approved):

PR #85, opened 2021-03-10
fix crashes caused by brittle type cast in filter evaluation

timburks:
This fixes crashes that occurred when filter expressions did not evaluate to bool. Such expressions are now treated as invalid arguments.

### DIFF
--- a/server/actions_apis.go
+++ b/server/actions_apis.go
@@ -16,7 +16,7 @@ package server
 
 import (
 	"context"
-	"errors"
+	"fmt"
 
 	"github.com/apigee/registry/rpc"
 	"github.com/apigee/registry/server/models"
@@ -179,11 +179,9 @@ func (s *RegistryServer) ListApis(ctx context.Context, req *rpc.ListApisRequest)
 			if err != nil {
 				return nil, invalidArgumentError(err)
 			}
-			v, ok := out.Value().(bool)
-			if !ok {
-				return nil, invalidArgumentError(errors.New("expression does not evaluate to a boolean"))
-			}
-			if !v {
+			if v, ok := out.Value().(bool); !ok {
+				return nil, invalidArgumentError(fmt.Errorf("expression does not evaluate to a boolean (instead yielding %T)", out.Value()))
+			} else if !v {
 				continue
 			}
 		}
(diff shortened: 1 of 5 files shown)


server/actions_artifacts.go
```diff
@@ -187,7 +188,11 @@ func (s *RegistryServer) ListArtifacts(ctx context.Context, req *rpc.ListArtifac
 			if err != nil {
 				return nil, invalidArgumentError(err)
 			}
-			if !out.Value().(bool) {
+			v, ok := out.Value().(bool)
+			if !ok {
+				return nil, invalidArgumentError(errors.New("expression does not evaluate to a boolean"))
+			}
+			if !v {
 				continue
```
timburks, line 196: It's noteworthy that this code is duplicated five times - I'd support factoring this out into common filtering code, but wanted to get this working first before formalizing that structure.

server/actions_apis.go
```diff
@@ -178,7 +179,11 @@ func (s *RegistryServer) ListApis(ctx context.Context, req *rpc.ListApisRequest)
 			if err != nil {
 				return nil, invalidArgumentError(err)
 			}
-			if !out.Value().(bool) {
+			v, ok := out.Value().(bool)
+			if !ok {
+				return nil, invalidArgumentError(errors.New("expression does not evaluate to a boolean"))
```
seaneganx, line 184: Can we offer any information about the expression in this error message? Would printing out `prg` provide any useful information for understanding the error?

server/actions_apis.go
```diff
@@ -178,7 +179,11 @@ func (s *RegistryServer) ListApis(ctx context.Context, req *rpc.ListApisRequest)
...
+			}
+			if !v {
 				continue
 			}
```
seaneganx, line 188: Nit: Both of these if statements are related and can be combined as the only use of `v` and `ok`.
```
if v, ok := out.Value().(bool); !ok {
...
} else if !v {
...
}
```

seaneganx (approved):
LGTM with two small changes to consider.

server/actions_apis.go
timburks: Good idea, I've added the expression's type to the error.

server/actions_apis.go
timburks: sgtm, updated

seaneganx (approved):

PR #86, opened 2021-03-10
Expand testing for API methods

seaneganx:
* Check that List only returns APIs from a specified parent project
* Check that List works across multiple collections

### DIFF
--- a/server/actions_apis.go
+++ b/server/actions_apis.go
@@ -68,7 +68,7 @@ func (s *RegistryServer) CreateApi(ctx context.Context, req *rpc.CreateApiReques
 		return nil, err
 	}
 
-	message, err := api.Message(rpc.View_BASIC)
+	message, err := api.Message(rpc.View_FULL)
 	if err != nil {
 		return nil, internalError(err)
 	}
--- a/server/actions_apis.go
+++ b/server/actions_apis.go
@@ -245,7 +245,7 @@ func (s *RegistryServer) UpdateApi(ctx context.Context, req *rpc.UpdateApiReques
 		return nil, err
 	}
 
-	message, err := api.Message(rpc.View_BASIC)
+	message, err := api.Message(rpc.View_FULL)
 	if err != nil {
 		return nil, internalError(err)
 	}
--- a/server/actions_apis.go
+++ b/server/actions_apis.go
@@ -152,6 +152,9 @@ func (s *RegistryServer) ListApis(ctx context.Context, req *rpc.ListApisRequest)
 	}
 	if name.ProjectID != "-" {
 		q = q.Require("ProjectID", name.ProjectID)
+		if _, err := getProject(ctx, client, name); err != nil {
+			return nil, err
+		}
 	}
 	prg, err := createFilterOperator(req.GetFilter(),
 		[]filterArg{
--- a/server/actions_apis.go
+++ b/server/actions_apis.go
@@ -68,7 +68,7 @@ func (s *RegistryServer) CreateApi(ctx context.Context, req *rpc.CreateApiReques
 		return nil, err
 	}
 
-	message, err := api.Message(rpc.View_FULL)
+	message, err := api.Message(rpc.View_BASIC)
 	if err != nil {
 		return nil, internalError(err)
 	}
@@ -248,7 +248,7 @@ func (s *RegistryServer) UpdateApi(ctx context.Context, req *rpc.UpdateApiReques
 		return nil, err
 	}
 
-	message, err := api.Message(rpc.View_FULL)
+	message, err := api.Message(rpc.View_BASIC)
 	if err != nil {
 		return nil, internalError(err)
 	}
--- a/server/actions_apis_test.go
+++ b/server/actions_apis_test.go
@@ -637,7 +637,17 @@ func TestUpdateApi(t *testing.T) {
 		want *rpc.Api
 	}{
 		{
-			desc: "default parameters",
+			desc: "populated resource with default parameters",
+			seed: fullApi,
+			req: &rpc.UpdateApiRequest{
+				Api: &rpc.Api{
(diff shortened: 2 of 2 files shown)


server/actions_apis.go
```diff
@@ -68,7 +68,7 @@ func (s *RegistryServer) CreateApi(ctx context.Context, req *rpc.CreateApiReques
 		return nil, err
 	}
 
-	message, err := api.Message(rpc.View_BASIC)
+	message, err := api.Message(rpc.View_FULL)
```
timburks, line 71: If we make this change, I think for consistency we should also do it for versions and specs, and if we do this for specs, then Create and Update methods will return the full spec. I think that's too much data to return - uploading a spec shouldn't also re-download it.

We might separately discuss whether annotations should be returned in BASIC views. They were originally kept out of BASIC so they could be used for larger objects than labels.

I guess this change was made to simplify testing of Create/Update methods... were there other reasons? (it seems out of scope for this PR)

server/actions_apis.go
```diff
@@ -245,7 +248,7 @@ func (s *RegistryServer) UpdateApi(ctx context.Context, req *rpc.UpdateApiReques
...
```
timburks, line 251: As above.

timburks (changes requested):

server/actions_apis.go
timburks: The new testing is great!

server/actions_apis.go
seaneganx: Correct, my intention is to return the full resource for Create/Update across the entire API. I agree that it's strange to have spec contents returned when we just uploaded them. I don't think testing is simplified by this change. My reason was that I wrote tests based on the AIPs and our protos which led to the discovery that this case doesn't pass. 

[AIP-133](https://google.aip.dev/133#guidance) is pretty direct in requiring a fully populated response with the only exception being INPUT_ONLY fields. I agree with the purpose behind the guidance because as an API consumer mutating a resource, I expect the response to provide the resulting resource state. 

Here are some of my thoughts about the conflict:
- Returning partial responses on Create/Update isn't great because it goes against AIP-133 and gives the consumer no confirmation that the operation succeeded as intended.
- Returning full responses on Create/Update isn't great because spec contents can be large and can be unreadable if compressed. This makes an otherwise lightweight and human readable API response heavy and potentially clunky to use manually.
- Perhaps spec contents would make more sense as an OUTPUT_ONLY field, uploaded using a custom method after the spec resource is created.
- Taking that idea one step farther, perhaps spec contents would make more sense as an artifact of the spec resource. Artifacts and spec contents already have the same `mime_type`, `size_bytes`, and `hash` fields. This might introduce an unusual resource association where one artifact is special compared to the others.
- We can't make either of these changes without a full version bump.
- There doesn't seem to be any information about file uploading or working with binary data in the AIPs, but this seems like a generally useful topic. Maybe it could be good to discuss with a wider range of folks.

timburks: Great points, particularly in highlighting the conflict with API-133.

I think adding additional methods to upload spec and artifact bodies would complicate the API and reduce usability since it would take two calls to do something that fits well in a single API call.

Since both artifacts and specs have contents fields, we would still have this problem for artifacts even if we changed specs to use artifacts to hold spec bodies. That also would require multiple API calls to upload a spec, which seems less usable.

I don't think a user would find it intuitive or beneficial for a create or update call for specs or artifacts to require them to re-download the full resource in the returned value. These calls return a hash which can be used to verify the upload.

Annotations seem to be a middle ground. Not wanting to constrain how customers might use these fields, I originally preferred to err on the side of caution, treating annotations as larger collections of data which might even be stored separately in future API implementations. This is inconsistent with AIP-133, but as above, so is the treatment of spec and artifact contents. A difference for annotations is that there is nothing returned by the Create/Update methods that can be used to confirm that they were set (other than the API status code, but I think it might be a good idea to define and return a hash for annotations). 

Also, annotations are mentioned in the [AIP-128 draft](https://google.aip.dev/128#annotations) where they are constrained to follow [Kubernetes limits](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) which say that "metadata in an annotation can be small or large, structured or unstructured." If this metadata is (always) "small" then I think there's no problem with returning artifacts in BASIC views and from Create/Update methods. But if it is (sometimes) "large", then I think the same concerns that exist for spec and artifact contents would apply.

The working assumption so far has been that Create and Update methods should return BASIC views that exclude artifacts and contents fields (making this easy to change if we decide to change it). But since it has broader implications, I think we should not make that change in this PR. Can we keep the tests and reported conflicts, marking them "skipped" for now?

seaneganx: Good points. I pushed a new commit with the changes back to basic view. We can discuss the details of this further offline.

server/actions_apis_test.go
```diff
@@ -16,6 +16,34 @@ import (
...
+		RecommendedVersion: "v1",
+		Labels: map[string]string{
+			"label-key": "label-value",
+		},
+		Annotations: map[string]string{
+			"annotation-key": "annotation-value",
+		},
+	}
+)
+
```
timburks, line 46: I think this structure is great, and just wonder - do you see an advantage to building both views explicitly instead of deriving one from a copy of the other?

timburks (approved):

server/actions_apis_test.go
seaneganx: I think simplicity is the primary benefit. Ideally the data for each test case shouldn't require any code execution to produce because it makes it more difficult to understand the test.

It would be better if we could define these in-line so readers don't have to jump around the file to understand the test, but fully populated structs across several test cases would be difficult to maintain if fields are added. This was the best compromise I could come up with.

PR #88, opened 2021-03-11
Expand testing for API Version methods

seaneganx:
* Check that List only returns versions from a specified parent API
* Check that List works across multiple collections

### DIFF
--- a/server/actions_versions.go
+++ b/server/actions_versions.go
@@ -160,7 +160,7 @@ func (s *RegistryServer) ListApiVersions(ctx context.Context, req *rpc.ListApiVe
 		if _, err := getApi(ctx, client, parent); err != nil {
 			return nil, err
 		}
-	} else if parent.ProjectID != "-" {
+	} else if parent.ProjectID != "-" && parent.ApiID == "-" {
 		if _, err := getProject(ctx, client, parent.Project()); err != nil {
 			return nil, err
 		}
--- a/server/actions_versions.go
+++ b/server/actions_versions.go
@@ -156,6 +156,15 @@ func (s *RegistryServer) ListApiVersions(ctx context.Context, req *rpc.ListApiVe
 	if parent.ApiID != "-" {
 		q = q.Require("ApiID", parent.ApiID)
 	}
+	if parent.ProjectID != "-" && parent.ApiID != "-" {
+		if _, err := getApi(ctx, client, parent); err != nil {
+			return nil, err
+		}
+	} else if parent.ProjectID != "-" {
+		if _, err := getProject(ctx, client, parent.Project()); err != nil {
+			return nil, err
+		}
+	}
 	prg, err := createFilterOperator(req.GetFilter(),
 		[]filterArg{
 			{"name", filterArgTypeString},
--- a/server/actions_versions_test.go
+++ b/server/actions_versions_test.go
@@ -361,6 +361,7 @@ func TestListApiVersions(t *testing.T) {
 				{Name: "projects/my-project/apis/my-api/versions/v1"},
 				{Name: "projects/my-project/apis/my-api/versions/v2"},
 				{Name: "projects/my-project/apis/my-api/versions/v3"},
+				{Name: "projects/my-project/apis/other-api/versions/v1"},
 			},
 			req: &rpc.ListApiVersionsRequest{
 				Parent: "projects/my-project/apis/my-api",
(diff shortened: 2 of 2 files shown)


server/actions_versions.go
```diff
@@ -156,6 +156,15 @@ func (s *RegistryServer) ListApiVersions(ctx context.Context, req *rpc.ListApiVe
...
 	}
+	if parent.ProjectID != "-" && parent.ApiID != "-" {
+		if _, err := getApi(ctx, client, parent); err != nil {
+			return nil, err
+		}
+	} else if parent.ProjectID != "-" {
+		if _, err := getProject(ctx, client, parent.Project()); err != nil {
+			return nil, err
+		}
+	}
```
timburks, line 167: These seem like fundamental problems that we should check and report earlier - either on line 139 or 143. Also, the current structure suggests that when both the API and project are missing, we report the error as the missing API. Subjectively, I think the missing project is a bigger problem and is what we should report. Is there a reason that I might have missed behind this particular ordering of checks?

seaneganx: We could move it back but I usually consider invalid argument errors more fundamental than ones that are determined by the state of the system like not found errors. The idea of providing an error that isn't "invalid argument" in response to an invalid request seems strange to me. Could you elaborate on why missing resources seem more fundamental to you?

Good catch about the resource priority when multiple errors could be returned. I agree the most helpful messages for consumers would point out the highest level resource that's missing, and I think this would be true for all our API methods. That isn't what I was attempting to do here though.

This code performs at most one database read to check existence of the lowest level parent that's fully specified in the path. The reason there are two possible existence checks is that we want to be able to return not found for a missing API (`projects/p/apis/missing`) and also for missing projects when reading across API collections (`projects/missing/apis/-`). Other methods have the same behavior but only ever need to check their immediate parent because they don't support unspecified collections.

If we want to provide the better errors (on this or any other method) we will have to sequentially check existence of all the parent resources to identify which one (if any) is the highest level missing resource. This wouldn't be difficult to implement, but it does increase the database interaction quite a bit. Particularly on resources like spec artifacts which would require an existence check of the project, api, version, spec, and artifact before continuing.

server/actions_versions.go
```diff
@@ -156,6 +156,15 @@ func (s *RegistryServer) ListApiVersions(ctx context.Context, req *rpc.ListApiVe
 	if parent.ApiID != "-" {
 		q = q.Require("ApiID", parent.ApiID)
 	}
+	if parent.ProjectID != "-" && parent.ApiID != "-" {
+		if _, err := getApi(ctx, client, parent); err != nil {
+			return nil, err
+		}
+	} else if parent.ProjectID != "-" {
```
timburks, line 163: I think some of my confusion reading this came from not realizing that parent.ApiID == "-" for this test (based on the condition above). It might help to restructure this so there's a single parent.ProjectID != "-" test wrapping the if condition on ApiId. 

Have you written this code for the more complicated case (specs)? It might be easier to work back from that.

server/actions_versions.go
timburks: Thanks, I'm OK with this. I think the case for "invalid arguments" being reported first is fine, particularly since these are quick checks that can potentially prevent needless database accesses. The existence checks might be clarified with the suggestion above, and I'd rather wait and see if anyone is confused about these errors before complicating them further.

timburks (approved):

server/actions_versions.go
seaneganx: I think that works out fine for this case, but in specs I find it confusing with the different levels of nesting.
```
	if parent.ProjectID != "-" {
		if parent.ApiID != "-" {
			if parent.VersionID != "-" {
				if _, err := getVersion(ctx, client, parent); err != nil {
					return nil, err
				}
			} else {
				if _, err := getApi(ctx, client, parent.Api()); err != nil {
					return nil, err
				}
			}
		} else {
			if _, err := getProject(ctx, client, parent.Project()); err != nil {
				return nil, err
			}
		}
	}
```

I think my current approach with some additional details can increase clarity. Again with specs as an example.
```
	if parent.ProjectID != "-" && parent.ApiID != "-" && parent.VersionID != "-" {
		if _, err := getVersion(ctx, client, parent); err != nil {
			return nil, err
		}
	} else if parent.ProjectID != "-" && parent.ApiID != "-" && parent.VersionID == "-" {
		if _, err := getApi(ctx, client, parent.Api()); err != nil {
			return nil, err
		}
	} else if parent.ProjectID != "-" && parent.ApiID == "-" && parent.VersionID == "-" {
		if _, err := getProject(ctx, client, parent.Project()); err != nil {
			return nil, err
		}
	}
```

timburks (approved):

PR #94, opened 2021-03-22
Fix for sed command on MacOS

shrutiparabgoogle:
This command was previously. failing on MacOS leading to the insecure mode not working properly for different CLI tools.


timburks (approved):
Also manually verified on Ubuntu

PR #101, opened 2021-03-29
Update API Spec and SpecRevision methods 

seaneganx:
* Add tests for method behaviors
* Move revision methods into separate file
* Order revisions by revision creation time
* Reject negative page sizes in List requests
* Return the correct page size and token on revision listing with sqlite
* Refactor existing methods to use common helper methods

### DIFF
--- a/server/actions_spec_revisions.go
+++ b/server/actions_spec_revisions.go
@@ -1,0 +1,294 @@
+package server
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/models"
+	"github.com/apigee/registry/server/names"
+	"github.com/apigee/registry/server/storage"
+	"github.com/golang/protobuf/ptypes/empty"
+	"google.golang.org/api/iterator"
+)
+
+// ListApiSpecRevisions handles the corresponding API request.
+func (s *RegistryServer) ListApiSpecRevisions(ctx context.Context, req *rpc.ListApiSpecRevisionsRequest) (*rpc.ListApiSpecRevisionsResponse, error) {
+	client, err := s.getStorageClient(ctx)
+	if err != nil {
+		return nil, unavailableError(err)
+	}
+	defer s.releaseStorageClient(client)
+
+	if req.GetPageSize() < 0 {
+		return nil, invalidArgumentError(fmt.Errorf("invalid page_size %d: must not be negative", req.GetPageSize()))
+	}
+
+	name, err := names.ParseSpec(req.GetName())
+	if err != nil {
+		return nil, invalidArgumentError(err)
+	}
+
+	q := client.NewQuery(storage.SpecEntityName)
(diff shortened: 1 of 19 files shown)


server/actions_spec_revisions.go
```diff
@@ -0,0 +1,294 @@
+package server
```
timburks, line 1: We need to include license headers in all new files (sorry I'm just now noticing this)

server/actions_versions.go
```diff
@@ -252,7 +252,7 @@ func (s *RegistryServer) UpdateApiVersion(ctx context.Context, req *rpc.UpdateAp
 	}
 
 	if err := saveVersion(ctx, client, version); err != nil {
-		return nil, internalError(err)
+		return nil, err
```
timburks, line 255: Why is this not wrapping the err with internalError()? ... is this being caught at a higher level and handled there?

server/storage/storage.go
```diff
@@ -30,6 +29,10 @@ const (
 	ApiEntityName = "Api"
 	// VersionEntityName is the storage entity name for API version resources.
 	VersionEntityName = "Version"
+	// SpecEntityName is the storage entity name for API spec resources.
```
timburks, line 32: let's talk offline about the decision to move these names from the models package to the storage package - it seems to be part of sanitizing the models layer to be independent of the storage details, but there are other things in the models package that are storage-dependent (the datastore and gorm annotations on struct fields), so this is a fuzzy boundary. 

I originally added the storage package as a way to abstract the interface to different storage backends (datastore, sql) and used the models package to describe most entity-specific details. A few things broke that (deletions seem like the biggest), but it was still fairly thin and minimal, making it relatively easy to consider adding other backends if needed.

timburks: since this is a multi-pr change that's already in progress, I don't think we should stop it, but it's an internal detail that we may want to revisit.

timburks (approved):
Please add license headers to any files that you've added that lack them; otherwise this is good to go. I had another comment that we can discuss offline.

server/actions_versions.go
seaneganx: The helper method already wraps errors as needed. In this case the only error code returned is internal so this would just be a double-wrap. Some other helpers (notably ones that read data) need to return several error codes depending on the error, so I thought it would make more sense for the helpers to be responsible for setting error codes.

server/actions_spec_revisions.go
seaneganx: Will do.

server/storage/storage.go
seaneganx: I agree, let's revisit this. My reasoning for the move was actually that these entity names solely used as arguments to storage.Client methods, so it made sense to locate them in the same package as the Client interface.

server/actions_versions.go
timburks: SGTM, thanks!

PR #105, opened 2021-04-01
Compute external references and provided files of protobuf API specs.

timburks:


cmd/registry/cmd/compute-references.go
```diff
@@ -0,0 +1,144 @@
...
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/names"
+	openapi_v2 "github.com/googleapis/gnostic/openapiv2"
+	openapi_v3 "github.com/googleapis/gnostic/openapiv3"
```
seaneganx, line 27: What's the purpose behind these renamed imports? The original package name on its own seems preferable because it doesn't have underscores. Same comment applies for other files.

cmd/registry/cmd/compute-references.go
```diff
@@ -0,0 +1,144 @@
...
+					ctx:      ctx,
+					client:   client,
+					specName: spec.Name,
+				}
+			})
+			if err != nil {
+				log.Fatalf("%s", err.Error())
+			}
+			close(taskQueue)
+			core.WaitGroup().Wait()
```
seaneganx, line 68: Should the workers be cleaned up when the regex pattern doesn't match?

cmd/registry/cmd/compute-references.go
```diff
@@ -0,0 +1,144 @@
...
+	},
+}
+
+type computeReferencesTask struct {
+	ctx      context.Context
+	client   connection.Client
+	specName string
+}
+
+func (task *computeReferencesTask) Name() string {
```
seaneganx, line 79: This name might be an established pattern already, but FYI this function seems like it might be useful as `String() string` where it's recognized and used by formatters and other utilities for printing the string version of a task.

cmd/registry/cmd/compute-references.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			return err
+		}
+	} else if core.IsProto(spec.GetMimeType()) && core.IsZipArchive(spec.GetMimeType()) {
+		references, err = core.NewReferencesFromZippedProtos(spec.GetContents())
+		if err != nil {
+			return fmt.Errorf("error processing protos: %s", spec.Name)
+		}
+	} else {
+		return fmt.Errorf("we don't know how to compute references for %s", spec.Name)
```
seaneganx, line 127: Could this error indicate that the MIME type is the issue?

cmd/registry/cmd/compute-references.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if references == nil {
+		return fmt.Errorf("we don't know how to compute references for %s", spec.Name)
+	}
```
seaneganx, line 131: Do we intend to have `nil` be a valid non-error return value from NewReferencesFromX methods? This seems unusual to me. IMO those methods should be responsible for checking this and return an error when they cannot compute the references.

cmd/registry/core/references.go
```diff
@@ -0,0 +1,124 @@
...
+	// unzip the protos to the temp directory
+	_, err = UnzipArchiveToPath(b, dname)
+	if err != nil {
+		return nil, err
+	}
+	// process the directory
+	internals, externals, err := internalsAndExternalsForPath(dname)
+	externals = removeInternalsAndDuplicatesFromExternals(internals, externals)
+	sort.Strings(internals)
+	sort.Strings(externals)
```
seaneganx, line 59: Nit: Consider changing these methods to `referencesForPath(path string) (*rpc.References, error)`, `deduplicate(refs *rpc.References)`, `filterExternals(refs *rpc.References, exclusions []string)`, and `sort(refs *rpc.References)`. A slightly heavier but perhaps cleaner approach would be to have a separate type that could have methods for these and one to convert to RPC.

The current directory processing code makes sense but I think if we modify it or reuse these functions in the future there are some bugs that could be easily introduced, e.g. forgetting the "remove" helper doesn't mutate the slice.

cmd/registry/core/references.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return externals, err
+	}
+
+	for _, x := range p.ProtoBody {
+		switch m := x.(type) {
+		case *parser.Import:
+			externals = append(externals, strings.Trim(m.Location, "\""))
+		default:
+		}
```
seaneganx, line 106: Do we expect many different types here? Should we log anything or return an error when the type is unrecognized? Depending on the answers to these questions, consider using
```
if import, ok := x.(*parser.Import); ok {
    ...
}
```

cmd/registry/cmd/compute-references.go
timburks: Good point. I have a commit to push that removes this anyway since OpenAPI specs aren't actually being handled.

cmd/registry/cmd/compute-references.go
timburks: I agree. Let's change them all in a single (future) PR.

cmd/registry/cmd/compute-references.go
timburks: Right. I'll update this to report the spec's MIME type.

cmd/registry/cmd/compute-references.go
timburks: I think that's a reasonable expectation - I'd say this was just defensive programming.

cmd/registry/core/references.go
timburks: That would be fine. This evolved from more general code that traversed the protobody and handled various other types that correspond to different parts of the proto file. I'll make the change you suggested.

cmd/registry/core/references.go
timburks: OK - I'm happy to do the renaming (is there a style guide reference that could help me get this right in the future?) but don't want to overbuild this because I think it is still somewhat experimental.

cmd/registry/cmd/compute-references.go
timburks: Do you mean should we shut down the goroutines? I haven't done that because these command handlers immediately exit. But I guess we could move the close and wait outside of the test to do that.

cmd/registry/core/references.go
seaneganx: There isn't any specific style guide for function naming. My suggestion comes from the single responsibility principle, having separate functions `foo()` and `bar()` instead of `doFooAndBar()`. My name choices were best effort from my reading of your code, but could be different.

cmd/registry/cmd/compute-references.go
seaneganx: I was just observing that the task queue and worker goroutines will be created and not closed if the name arg is invalid, but will be closed if the arg is valid. I'm curious why the name arg is relevant at all in whether or not we clean up those resources explicitly.

seaneganx (approved):

PR #106, opened 2021-04-01
Core API Test Suite + Bug Fixes

seaneganx:
This branch includes incrementally reviewed changes that introduce a suite of unit tests for the server. Along with the tests, some bug fixes are included for some situations that tests detected an issue and could be easily resolved. Other known bugs have their tests marked as skipped until they can be addressed on the main branch.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -4,11 +4,11 @@ on:
   push:
     branches:
     - main
-    - v1
+    - v1.1
   pull_request:
     branches:
     - main
-    - v1
+    - v1.1
 jobs:
 
   build:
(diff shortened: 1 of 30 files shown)


.github/workflows/go.yml
```diff
@@ -4,11 +4,11 @@ on:
   push:
     branches:
     - main
-    - v1
+    - v1.1
```
timburks, line 7: I think we can remove this now (and the tests will still run on this PR); I think it's also fine to proceed with merging this and removing the v1.1 tests in a subsequent PR.

timburks (approved):
Thanks!  This is big/significant! I'm mostly relying on our incremental reviews and the tests, but based on those I think this is ready to merge.

PR #107, opened 2021-04-06
Refactor worker implementation

shrutiparabgoogle:
Addressing comments from the previous PR:

**1. Package utils in cmd/capabilities directory**
I think this is good advice about "utils" packages: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common.

**2. Topic name for registry's pub/sub channel**
Maybe having a more specific name for pub/sub topic is better. My understanding is this will be used to publish registry events, so maybe something like registry-events?
This will require a refactor on the registry-backend side too.

**3. For struct Dispatcher defined in dispatcher.go**
Do we intend on having this struct provide any functionality beyond setting up and starting a server? Requiring a call to setUp before its other methods indicates we're putting too much responsibility on the user of the type. It seems more appropriate to create a package dispatcher with exported method StartServer(context.Context) that handles any necessary setup internally. The main package would become very lightweight with its only responsibility being starting the server (possibly parsing configuration info in the future).


seaneganx (approved):

timburks (approved):

wangyh28 (approved):
This LGTM

PR #114, opened 2021-04-09
Update datastore indexes for use of RevisionCreateTime in spec queries.

timburks:
This fixes datastore test failures introduced in recent changes.

### DIFF
--- a/server/datastore/index.yaml
+++ b/server/datastore/index.yaml
@@ -28,6 +28,5 @@ indexes:
   - name: ProjectID
   - name: SpecID
   - name: VersionID
-  - name: Currency
-  - name: CreateTime
+  - name: RevisionCreateTime
     direction: desc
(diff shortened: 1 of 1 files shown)


seaneganx (approved):

server/datastore/index.yaml
```diff
@@ -28,6 +28,5 @@ indexes:
   - name: ProjectID
   - name: SpecID
   - name: VersionID
-  - name: Currency
```
seaneganx, line 31: Just realized we might need this still. I haven't fully eliminated currency from ListSpecs because some changes are necessary to list only the most recent revisions of specs without it. Locally everything seems to work fine, but without the previous index (with currency) there might be issues.

seaneganx (changes requested):

server/datastore/index.yaml
timburks: Right - thanks for noticing. I restored the original index and added the new one

seaneganx (approved):

PR #116, opened 2021-04-12
Delete backup sed file in proto compile script

seaneganx:
This file isn't necessary (the backup version is invalid) and results in
an untracked file being created that should not be checked into changes.
It should be removed instead of ignored because of its invalidity.


seaneganx:
Alternatively, maybe we can use the sed command without creating the backup file. I think last time I tried that I ran into compatibility issues though, so removing the file seems like a reasonable approach.

timburks (approved):

shrutiparabgoogle (approved):
Looks good

Issue #92, opened 2021-03-18
Existence of parent is not checked during resource creation


seaneganx:
Creating resources with a non-existent parent does not result in an error. The parent is considered valid if it has the correct format. Create methods should check existence of the parent and return NOT_FOUND status if it does not exist.

This issue is present on all resource creation and listing methods including spec revision listing.

seaneganx:
This was fixed by #119.

Issue #90, opened 2021-03-18
Update method responses for missing resource bodies are inaccurate


seaneganx:
The response reports a missing resource name when the required resource body is missing. An error should be returned indicating the body is required. This issue is present on all update methods.

Issue #61, opened 2021-02-26
Generate IDs for resources when unspecified in create requests


seaneganx:
As a user of the CreateProject, CreateApi, CreateApiVersion, CreateApiSpec, and CreateArtifact methods, I would like to be able to create resources without providing a specific ID. My expectation is that a valid ID will be generated for me and returned as part of the response.

PR #119, opened 2021-04-19
Refactor Artifacts service

seaneganx:
* Move model storage code into DAO
* Use names package for parsing and representing resource names
* Add test suite

### DIFF
--- a/server/actions_artifacts_test.go
+++ b/server/actions_artifacts_test.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC. All Rights Reserved.
+// Copyright 2021 Google LLC. All Rights Reserved.
 //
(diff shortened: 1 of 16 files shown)


seaneganx:
Fixes #92 #90 #61

server/actions_artifacts_test.go
```diff
@@ -0,0 +1,894 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
```
timburks, line 1: nit: 2021

timburks (approved):
This looks great! Thanks for staging it in multiple commits, that made it much easier to digest.

server/actions_artifacts_test.go
seaneganx: Fixed.

PR #125, opened 2021-04-20
Write all fields including zero-value in gorm

### LINKED ISSUE #124: Update methods cannot clear field data with gorm storage backend
This bug does not exist when using datastore as the storage backend. It is confirmed to exist for sqlite3 and postgres backends, likely meaning the issue has to do with our use of gorm. I confirmed the updated model reaches the gorm client layer correctly, but any fields that are empty do not persist to the database.

The response from update methods is correct because the model itself is correctly updated and used to create the response. This bug reveals itself on a future request to retrieve the updated resource, where fields have not changed when they should have been cleared.

seaneganx:
Fixes #124

### DIFF
--- a/server/gorm/client.go
+++ b/server/gorm/client.go
@@ -211,6 +211,7 @@ func (c *Client) Put(ctx context.Context, k storage.Key, v interface{}) (storage
 	}
 	c.db.Transaction(
 		func(tx *gorm.DB) error {
+			// Update all fields from model: https://gorm.io/docs/update.html#Update-Selected-Fields
 			rowsAffected := tx.Model(v).Select("*").Where("key = ?", k.(*Key).Name).Updates(v).RowsAffected
 			if rowsAffected == 0 {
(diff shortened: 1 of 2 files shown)


server/gorm/client.go
```diff
@@ -213,7 +211,7 @@ func (c *Client) Put(ctx context.Context, k storage.Key, v interface{}) (storage
 	}
 	c.db.Transaction(
 		func(tx *gorm.DB) error {
-			rowsAffected := tx.Model(v).Where("key = ?", k.(*Key).Name).Updates(v).RowsAffected
+			rowsAffected := tx.Model(v).Select("*").Where("key = ?", k.(*Key).Name).Updates(v).RowsAffected
```
timburks, line 214: Looks good. It might help to link to https://gorm.io/docs/update.html#Update-Selected-Fields for background on this.

timburks (approved):

PR #129, opened 2021-04-22
Normalize resource names for case insensitivity

### LINKED ISSUE #91: Resource names are case sensitive
Created resources are case sensitive, allowing creation of independent resources like specs that differ only be changing the case of their resource IDs.

[AIP-122](https://google.aip.dev/122#guidance) recommends against allowing uppercase letters in resource IDs. Allowing case sensitivity in resource names can result in confusing API behavior, e.g. resources not being found when accessed using a case insensitive identifier match.

One possible solution would be to reject uppercase characters during resource creation as recommended by AIP-122, and additionally attempt to canonicalize resource names to lowercase for other operations. This would make consumers aware of the restriction at resource creation time and reduce confusion from case sensitivity mismatch during future accesses.

seaneganx:
Fixes #91.


timburks (approved):
We might want to clarify that we're lowercasing resource names and not always case-insensitive.  If nothing else was case-sensitive, I think filtering would still be case-sensitive and that we wouldn't want to lowercase filter expressions because they might refer to mixed-case text in titles and descriptions (unless we lowercase everything that we pass to the filtering implementation, which is outside the scope of this PR).

seaneganx:
> We might want to clarify that we're lowercasing resource names and not always case-insensitive. 

What do you think about making lowercase identifiers part of request validation? That would eliminate normalization issues but make the API more restrictive and require changes to our other tools like bulk upload to ensure they lowercase directory/file names that are used as custom IDs.

>If nothing else was case-sensitive, I think filtering would still be case-sensitive and that we wouldn't want to lowercase filter expressions because they might refer to mixed-case text in titles and descriptions (unless we lowercase everything that we pass to the filtering implementation, which is outside the scope of this PR).

Good catch, filtering case sensitivity is an issue that might be difficult to handle with mixed case content. Enforcing lowercase IDs at resource creation time would remove this issue because all filtering will be case sensitive still, but we should be careful not to make the API less usable for sake of easier engineering.

timburks:
Requiring lowercase identifiers as part of request validation sounds good to me, and is probably better than automatically lowercasing because we are informing users of the requirement up front and not later when they might be surprised to find that we've changed the case on identifiers they specified. It's also consistent with AIP-122: "resource IDs should not use upper-case letters."

seaneganx:
Pushed up some changes that separate parsing validation from custom ID validation, allowing consumers to provide mixed case resource names when referencing existing resources but still preventing the creation of resources with mixed case IDs. PTAL.

timburks (approved):

PR #132, opened 2021-05-03
Add "contents" singleton for direct access to API spec contents

timburks:
For discussion - and possible inclusion - this adds a new API method that allows API specs to be directly downloaded as the response message bodies. It uses the [httpbody.HttpBody](https://github.com/googleapis/googleapis/blob/5a04154e7c7c0e98e0e4085f6e2c67bd5bff6ff8/google/api/httpbody.proto#L28) type for responses, which allows transcoded gRPC methods to return arbitrary data. This allows specs to be directly downloaded from the service with HTTP requests, e.g.
```
curl "https://$HOST/v1/projects/googleapis/apis/googleapis.com-youtube/versions/v3/specs/discovery.json/contents" \
  -H "Authorization: Bearer $TOKEN"
```
This uses the [singleton](https://google.aip.dev/156) pattern. This seems preferable (and more clear) than an alternate [custom method](https://google.aip.dev/136), which was originally considered using `:getContents` as the name of the custom method.

This seems also appropriate to add for Artifacts.

Because this is primarily to support easy access over gRPC transcoding, I don't think this necessarily means that other means of accessing API spec contents should be removed.


server/actions_specs.go
```diff
@@ -198,6 +203,46 @@ func (s *RegistryServer) getApiSpecRevision(ctx context.Context, name names.Spec
...
+	var err error
+	if name, err := names.ParseSpec(req.GetName()); err == nil {
+		spec, err = s.getApiSpec(ctx, name, rpc.View_FULL)
+	} else if name, err := names.ParseSpecRevision(req.GetName()); err == nil {
+		spec, err = s.getApiSpecRevision(ctx, name, rpc.View_FULL)
+	}
+	if err != nil {
+		return nil, err
+	}
+	if spec != nil {
```
seaneganx, line 228: Nit: Negating this to make it an error check (rather than a success check) might read better.

server/actions_specs.go
```diff
@@ -198,6 +203,46 @@ func (s *RegistryServer) getApiSpecRevision(ctx context.Context, name names.Spec
...
+		contents := spec.Contents
+		contentType := spec.MimeType
```
seaneganx, line 230: Nit: These seem to slightly obscure the data we're operating on in the next few lines. Could we just reference MimeType and Contents directly?

server/actions_specs.go
```diff
@@ -198,6 +203,46 @@ func (s *RegistryServer) getApiSpecRevision(ctx context.Context, name names.Spec
...
+			if err != nil {
+				return nil, err
+			}
+		}
+		return &httpbody.HttpBody{
+			ContentType: contentType,
+			Data: contents,
+		}, nil
+	}
+	return nil, invalidArgumentError(fmt.Errorf("invalid resource name %q, must be an API spec or revision", req.GetName()))
```
seaneganx, line 243: Is this error message correct? This only happens if getApiSpec or getApiSpecRevision return a nil spec and nil error, which seems to me more like an unknown internal error occurred.

server/actions_specs.go
```diff
@@ -198,6 +203,46 @@ func (s *RegistryServer) getApiSpecRevision(ctx context.Context, name names.Spec
...
+	}
+	if err != nil {
+		return nil, err
+	}
+	if spec != nil {
+		contents := spec.Contents
+		contentType := spec.MimeType
+		if (strings.Contains(contentType, "+gzip")) {
+			contents, err = GUnzippedBytes(contents)
+			contentType = strings.Replace(contentType, "+gzip", "", 1)
```
seaneganx, line 233: Do we want to replace +gzip even if it's not a suffix of the content type? If it exists multiple times, do we only want to remove the first instance of it? These both seem like very unusual and maybe invalid cases.

Consider using strings.HasSuffix instead of strings.Contains above, and strings.TrimSuffix here if covers everything we want. If not, maybe a comment would be useful to point out an example of an unusual case where this would function differently.

seaneganx (commented):
Looks good to me! Regarding other means of accessing contents, I think we should consider how the API is likely to be used in practice.

With CreateApiSpec it doesn't make sense to remove the contents field from ApiSpec because it would force spec creation to take two steps, so ApiSpec should certainly have the field. What I think is more reasonable to consider is whether that field should be marked as input only.

Currently we avoid returning that field unless VIEW_FULL is specifically requested. For our mutation methods that don't support views, this is at odds with [AIP-133](https://google.aip.dev/133#guidance) and [AIP-134](https://google.aip.dev/134#guidance) guidance. They suggest a fully populated ApiSpec should be returned from CreatApiSpec and UpdateApiSpec. With an input only field we could more closely align here while still allowing mutation as we currently do.

The main downside I see for making spec contents input only is that GetApiSpec, ListApiSpecs, and ListApiSpecRevisions will not be able to list the contents for a collection of specs. I'm not sure whether this use case is actually needed, but it reasonably could be. One compromise could be to provide a batch get method as described in [AIP-231](https://google.aip.dev/231), so consumers can list a page and fetch all the full view specs using the batch method.

From what I understand right now I think either solution should work fine. Having two separate calls to fetch Spec/Contents or ManySpecs/ManyContents should work well in a registry viewer interface because the spec metadata is already likely displayed in a different location than the spec contents. Having separate methods would also encourage a potentially good practice of only downloading spec contents if they will actually be displayed.

server/actions_specs.go
timburks: I originally used HasSuffix and discovered that this broke when mime types had version parameters as I'd proposed in https://apigee.github.io/registry/media_types.html - for example `application/x.openapi+gzip;version=3.0.0`. I agree that more careful parsing would be good to consider/plan for the future.

server/actions_specs.go
timburks: This is rearranged in the latest version.

server/actions_specs.go
timburks: Revised in the latest version.

server/actions_specs.go
timburks: Done, thanks

server/actions_specs.go
```diff
@@ -198,6 +203,49 @@ func (s *RegistryServer) getApiSpecRevision(ctx context.Context, name names.Spec
...
+	if name, err := names.ParseSpec(specName); err == nil {
+		spec, err = s.getApiSpec(ctx, name, rpc.View_FULL)
+	} else if name, err := names.ParseSpecRevision(specName); err == nil {
+		spec, err = s.getApiSpecRevision(ctx, name, rpc.View_FULL)
+	}
+	if err != nil {
+		return nil, err
+	}
+	if spec == nil {
+		return nil, invalidArgumentError(fmt.Errorf("failed to locate %q", req.GetName()))
```
seaneganx, line 230: If we failed to locate the spec, a "not found" error would have been returned at line 227. Receiving a `nil` spec with `nil` error isn't defined behavior and means we introduced a bug. Given we don't know what happened, I think "internal" or "unknown" error would make more sense.

Am I misunderstanding the purpose of this being an "invalid argument" error or the reason for its message?

server/actions_specs.go
```diff
@@ -198,6 +203,49 @@ func (s *RegistryServer) getApiSpecRevision(ctx context.Context, name names.Spec
...
+	if strings.Contains(spec.MimeType, "+gzip") {
+		contents, err := GUnzippedBytes(spec.Contents);
+		if err != nil {
+			return nil, err
+		}
+		return &httpbody.HttpBody{
+			ContentType: strings.Replace(spec.MimeType, "+gzip", "", 1),
+			Data:        contents,
+		}, nil
+	} else {
```
seaneganx, line 241: Nit: This nesting isn't necessary.

seaneganx (commented):
Those MIME type version parameters do make things interesting. As long as we expect only one `+gzip` tag to exist anywhere in the string I think this solution is fine. Based on your description in the link you provided I can't imagine why there would ever be several. It might still be worth adding a comment stating our assumption that only one exists.

server/actions_specs.go
timburks: removed

server/actions_specs.go
timburks: changed to internalError()

seaneganx (approved):

PR #141, opened 2021-05-13
Update openapi.yaml to match current protos.

timburks:


seaneganx (approved):
Is this generated or manually updated?

timburks:
Generated with https://github.com/apigee/registry/blob/main/tools/GENERATE-OPENAPI.sh - see https://github.com/google/gnostic/pull/247 for some necessary changes to the generator

PR #146, opened 2021-05-14
Minor build fixes

kenigeer:
- Another fix for `sed` on Mac
- .gitignore for IntelliJ

### DIFF
--- a/tools/COMPILE-PROTOS.sh
+++ b/tools/COMPILE-PROTOS.sh
@@ -82,9 +82,22 @@ END
 
 # patch the generated GAPIC to send Authorization tokens with insecure requests
 # (this allows the registry command line tools to test container builds)
-sed -i.bak 's/return metadata.NewOutgoingContext(ctx, out)/insecure := os.Getenv("APG_REGISTRY_INSECURE")\ntoken := os.Getenv("APG_REGISTRY_TOKEN")\nif insecure == "1" \&\& token != "" \{ \nout["authorization"] = append(out["authorization"], "Bearer "+token) \n\}\nreturn metadata.NewOutgoingContext(ctx, out)/' gapic/doc.go
-rm gapic/doc.go.bak
-gofmt -w gapic/doc.go
+FILE=gapic/doc.go
+PATCH='\
+insecure := os.Getenv("APG_REGISTRY_INSECURE")\
+token := os.Getenv("APG_REGISTRY_TOKEN")\
+if insecure == "1" \&\& token != "" {\
+  out["authorization"] = append(out["authorization"], "Bearer "+token)\
+}\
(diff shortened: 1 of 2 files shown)


tools/COMPILE-PROTOS.sh
```diff
@@ -82,7 +82,15 @@ END
...
 # (this allows the registry command line tools to test container builds)
-sed -i.bak 's/return metadata.NewOutgoingContext(ctx, out)/insecure := os.Getenv("APG_REGISTRY_INSECURE")\ntoken := os.Getenv("APG_REGISTRY_TOKEN")\nif insecure == "1" \&\& token != "" \{ \nout["authorization"] = append(out["authorization"], "Bearer "+token) \n\}\nreturn metadata.NewOutgoingContext(ctx, out)/' gapic/doc.go
+PATCH='\
+insecure := os.Getenv("APG_REGISTRY_INSECURE")\
+token := os.Getenv("APG_REGISTRY_TOKEN")\
+if insecure == "1" \&\& token != "" {\
+  out["authorization"] = append(out["authorization"], "Bearer "+token)\
+}\
+return metadata.NewOutgoingContext(ctx, out)\
+' # no ' or / in the patch; escape & and newlines
```
timburks, line 92: Just curious: what was breaking here? The prior code works for me on Big Sur (11.3).

tools/COMPILE-PROTOS.sh
```diff
@@ -82,7 +82,15 @@ END
...
+sed -i.bak "s/return metadata.NewOutgoingContext(ctx, out)/${PATCH}/" gapic/doc.go
```
timburks, line 93: Did you check this on Linux? We're not checking this in CI but if we did, that would clarify the safety of this - I think a simple grep for "APG_REGISTRY_INSECURE" in the generated file would verify that the patch was applied.

timburks: Could you add this to COMPILE-PROTOS.sh right after the `gofmt` call?
```
grep -q APG_REGISTRY_INSECURE gapic/doc.go
if [ $? -ne 0 ]
then
  echo "GAPIC patch failed"
  exit -1
fi
```

This will cause the make to fail, which will show up in CI.

tools/COMPILE-PROTOS.sh
kenigeer: `\n` is just escaped-n, `n`, not a newline, in 10.15

tools/COMPILE-PROTOS.sh
kenigeer: Also tested on Linux. Added grep check.

timburks (approved):

PR #155, opened 2021-05-21
Add a Dockerfile for a container containing the registry CLI tools.

timburks:
This adds a Dockerfile for building a "registry-tools" container containing the `registry` and `apg` CLIs.


seaneganx (approved):
LGTM - I'm fairly inexperienced with Docker though so it may be worth a second review depending on your confidence level about it.

PR #156, opened 2021-05-21
Decrease default timeout + additional error checking

seaneganx:
Updates to address some issues I've ran into while working on development of the CSV import command.


cmd/registry/core/projects.go
```diff
@@ -35,5 +35,7 @@ func EnsureProjectExists(ctx context.Context, client *gapic.RegistryClient, proj
 		if err != nil {
 			log.Fatalf("%s", err.Error())
 		}
+	} else if err != nil {
+		log.Fatalf("Failed to check project existence: %s", err.Error())
```
timburks, line 39: I misinterpreted this at first - I was thinking "failed to check" meant that a user had forgotten/neglected to make this check. I'm not sure if others would make that same mistake, but something like "project existence check failed" might have been more clear to me.

timburks (approved):

cmd/registry/core/projects.go
seaneganx: I agree in isolation, but with the error message included I think it's clear that an error occurred rather than a lack of action from the user. For example, `Failed to check project existence: context deadline exceeded`. Thoughts?

seaneganx: Added more info to the message.

timburks: Thanks (and good point). Apart from linter checks, we haven't put any style guidelines on our error messages. It might be worth an issue to review our messages and settle on expectations.

Issue #162, opened 2021-05-26
Add "registry label" and "registry annotate" subcommands to more easily manipulate labels and annotations.


timburks:
Currently the API supports adding and removing labels and annotations to resources, but at the API level, this requires dealing with the entire current collection of labels or annotations on a particular resource. Labels and annotations in the registry are modeled on Kubernetes, and thus we can find a good pattern for their manipulation in the `kubectl label` and `kubectl annotate` commands, documented [here](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#label) and [here](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#annotate). Let's consider adding something similar as subcommands of the `registry` tool.

timburks:
#166 addresses this for labels. We can support annotations with the exact same mechanism.

timburks:
Completed in #173

PR #166, opened 2021-06-01
New "registry label" subcommand to set labels on apis, versions, and specs

timburks:
Partially addresses #162 and includes usage information following the style proposed in #167.


cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (task *labelApiTask) Run() error {
+	if task.api.Labels == nil {
+		task.api.Labels = make(map[string]string)
+	}
+	if len(labelPairs) > 0 {
+		log.Printf("label %s with %v", task.api.Name, labelPairs)
+		for k, v := range labelPairs {
+			if v == "" {
```
seaneganx, line 112: Can you explain your reasoning for this behavior? If I'm understanding correctly, it means this command will ignore any keys set to empty values such as `someKey=`.

I think our API has different behavior by allowing empty label values, so this seems odd to me.

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,258 @@
...
+		_, err := task.client.UpdateApiSpec(task.ctx, req)
+		return err
+	}
+	return nil
+}
+
+func matchAndHandleLabelCmd(
+	ctx context.Context,
+	client connection.Client,
+	taskQueue chan core.Task,
```
seaneganx, line 191: Nit: Since we only want this function to add tasks to the queue (not read them) we can indicate that using directions, e.g.`taskQueue chan<- core.Task,`. Same for the functions below.

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	})
+}
+
+func labelSpecs(
+	ctx context.Context,
+	client *gapic.RegistryClient,
+	segments []string,
+	filterFlag string,
+	taskQueue chan core.Task) error {
```
seaneganx, line 250: Nit: I don't think this is better than a single line, but that's subjective of course. Is there a particular line length you're formatting to?

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,258 @@
...
+
+		// Initialize task queue.
+		taskQueue := make(chan core.Task, 1024)
+		workerCount := 64
+		for i := 0; i < workerCount; i++ {
+			core.WaitGroup().Add(1)
+			go core.Worker(ctx, taskQueue)
+		}
+
+		if len(args) > 1 {
```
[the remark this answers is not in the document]
timburks: right, probably just a legacy - I'll remove this check

cmd/registry/cmd/label.go
timburks: This would clear labels, but I'm looking more closely at kubectl patterns (which are different) and will update this to match those.

cmd/registry/cmd/label.go
timburks: Thanks, I'll change these. This usage exists throughout the registry tool, so it will be inconsistent without an overall fix.

cmd/registry/cmd/label.go
timburks: Not specifically - my editor window is usually 80-100 characters wide, and these lines would be around 140 chars.

cmd/registry/cmd/label.go
seaneganx: Okay. Our API behavior currently considers the labels field an atomic map that gets replaced fully if modified, so deleting is equivalent to not providing the entry. I'm interested in hearing if the kubectl patterns differ in any way.

timburks: Right (about the API behavior). kubectl takes care of that when users add labels (with "key=value") or remove them (with "key-") - that's the pattern that I'm following along with also requiring "--overwrite" to modify existing labels (as a safety precaution).

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,260 @@
...
+
+		taskQueue := make(chan core.Task, 1024)
+		workerCount := 64
+		for i := 0; i < workerCount; i++ {
+			core.WaitGroup().Add(1)
+			go core.Worker(ctx, taskQueue)
+		}
+
+		labelsToClear = make([]string, 0)
+		labelsToSet = make(map[string]string)
```
seaneganx, line 62: We aren't validating that these are disjoint sets. If a label is set and cleared in the same command it will arbitrarily have one behavior chosen.

seaneganx (approved):
LGTM - Just some comments about edge cases that are likely rare and some code clarity feedback.

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,260 @@
...
+		for k, _ := range labelsToSet {
+			if v, ok := task.api.Labels[k]; ok {
+				return fmt.Errorf("%q already has a value (%s), and --overwrite is false", k, v)
+			}
+		}
+	}
+	for _, k := range labelsToClear {
+		delete(task.api.Labels, k)
+	}
+	if len(labelsToSet) > 0 {
```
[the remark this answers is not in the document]
timburks: Thanks, removing three instances of this.

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,260 @@
...
+		Api: task.api,
+		UpdateMask: &field_mask.FieldMask{
+			Paths: []string{"labels"},
+		},
+	}
+	_, err := task.client.UpdateApi(task.ctx, req)
+	return err
+}
+
+func (task *labelVersionTask) Name() string {
```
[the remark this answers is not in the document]
timburks: Good point - and I think this has come up before. We should act on this in a separate PR because this is determined by the [Tasks interface](https://github.com/apigee/registry/blob/92b14e9ac410da3d87b1b1e2a3da5c77e121e7c0/cmd/registry/core/tasks.go#L26). #170

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,260 @@
...
+		for i := 0; i < workerCount; i++ {
+			core.WaitGroup().Add(1)
+			go core.Worker(ctx, taskQueue)
+		}
+
+		labelsToClear = make([]string, 0)
+		labelsToSet = make(map[string]string)
+		for _, labeling := range args[1:] {
+			if len(labeling) > 1 && strings.HasSuffix(labeling, "-") {
+				labelsToClear = append(labelsToClear, labeling[0:len(labeling)-1])
```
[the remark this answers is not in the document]
timburks: Done, thanks!

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,260 @@
...
+		workerCount := 64
+		for i := 0; i < workerCount; i++ {
+			core.WaitGroup().Add(1)
+			go core.Worker(ctx, taskQueue)
+		}
+
+		labelsToClear = make([]string, 0)
+		labelsToSet = make(map[string]string)
+		for _, labeling := range args[1:] {
+			if len(labeling) > 1 && strings.HasSuffix(labeling, "-") {
```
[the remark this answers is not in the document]
timburks: Informally I found that I could create a label named `a-` with `registry label <resource> a-=` and remove it with `registry label <resource> a--`. This is weird but not broken - happy to reconsider this case later, though.

cmd/registry/cmd/label.go
timburks: True that we aren't checking for disjoint sets. The behavior would be consistent though, always removing first and then setting, effectively overwriting an existing label.

cmd/registry/cmd/label.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (task *labelApiTask) Name() string {
+	return "label " + task.api.Name
+}
+
+func (task *labelApiTask) Run() error {
+	if task.api.Labels == nil {
+		task.api.Labels = make(map[string]string)
+	}
+	if !labelOverwrite {
```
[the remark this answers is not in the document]
timburks: Putting this in the loop would increase the number of times that we make this check. I don't think efficiency is a huge concern here, but that tips me toward leaving things as-is for now. Happy to reconsider.

Issue #170, opened 2021-06-03
Cleanup: rename Name to String in Task interface


timburks:
As pointed out by @seaneganx:
```
Using String() string will give this value when the struct is printed using common formatters like fmt.Println or Printf("%s", ...).
```

The code to change is [here](https://github.com/apigee/registry/blob/92b14e9ac410da3d87b1b1e2a3da5c77e121e7c0/cmd/registry/core/tasks.go#L26).

PR #171, opened 2021-06-03
Rename "Name()" to "String()" in Task interface.

timburks:
Addresses #170

### DIFF
--- a/cmd/registry/core/tasks.go
+++ b/cmd/registry/core/tasks.go
@@ -37,7 +37,7 @@ func Worker(ctx context.Context, taskChan <-chan Task) {
 	for task := range taskChan {
 		err := task.Run()
 		if err != nil {
-			log.Printf("%s: %s", task.String(), err.Error())
+			log.Printf("%s: %s", task, err)
 		}
 	}
 }
--- a/cmd/registry/core/tasks.go
+++ b/cmd/registry/core/tasks.go
@@ -23,7 +23,7 @@ import (
 // Task is a generic interface for a runnable operation
 type Task interface {
 	Run() error
-	Name() string
+	String() string
 }
 
 var wg sync.WaitGroup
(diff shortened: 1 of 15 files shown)


cmd/registry/core/tasks.go
```diff
@@ -37,7 +37,7 @@ func Worker(ctx context.Context, taskChan <-chan Task) {
 	for task := range taskChan {
 		err := task.Run()
 		if err != nil {
-			log.Printf("%s: %s", task.Name(), err.Error())
+			log.Printf("%s: %s", task.String(), err.Error())
```
seaneganx, line 40: FYI this is equivalent to `log.Printf("%s: %s", task, err)`. 

From the [fmt docs](https://golang.org/pkg/fmt/#hdr-Printing): 
> If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %v %x %X), the following two rules apply:
>
>4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).
>
>5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

seaneganx: Also, not relevant to this PR in particular but still worth mentioning: the log should have a newline character since Printf doesn't include one by default. Providing more info could also be nice since it's not ending execution, e.g. `log.Printf("Error running task %q: %s\n", task, err)`.

seaneganx (approved):

cmd/registry/core/tasks.go
timburks: Thanks! I made the simplification change that you suggested above, but want to address the error messaging recommendation separately because I found some other error messages being produced by tasks and think we should handle them consistently. Also in my usage of log.Printf, I've found that it automatically adds a newline at the end of log messages, which seems related to [this](https://golang.org/src/log/log.go?s=10139:10169#L373).

PR #173, opened 2021-06-07
Modify "registry label" command to also implement "registry annotate"

timburks:
These two commands provide equivalent capability for labels and annotations. This adds annotation support by generalizing the "registry label" command implementation to also provide "registry annotate".


cmd/registry/cmd/label.go
```diff
@@ -29,223 +29,262 @@ import (
...
+			}
+
+			close(taskQueue)
+			core.WaitGroup().Wait()
+		},
+	}
 }
 
-type labelSpecTask struct {
+type labelOrAnnotateApiTask struct {
```
seaneganx, line 134: Consider having separate tasks for labelling and annotating, or having a generic "updateApiTask" type. This particular task feels "specifically generic" where we don't get the benefits of simplicity or reusability. 

Is there a purpose for making this handle both labelling and annotating other than code reuse? The amount of helper functions and reliance on global state required to have it handle two behaviors makes me think we're taking two steps back for one step forward.

seaneganx (changes requested):

cmd/registry/cmd/label.go
timburks: I didn't like the replication but have split these into handler-specific tasks along with the command implementations.

cmd/registry/cmd/label.go
```diff
@@ -29,157 +29,67 @@ import (
...
 )
 
-var labelFilter string
-var labelOverwrite bool
-var labelsToSet map[string]string
-var labelsToClear []string
+var labelKeyFilter string
+var labelKeyOverwrite bool
+var labelKeysToSet map[string]string
+var labelKeysToClear []string
```
seaneganx, line 35: For the same reason as above, consider declaring these in the command's run function.

cmd/registry/cmd/label_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// Add some labels to the test api.
+	rootCmd.SetArgs([]string{"label", apiName, "a=1", "b=2"})
+	if err := labelCmd.Execute(); err != nil {
+		t.Fatalf("Execute() with args %v returned error: %s", rootCmd.Args, err)
+	}
+	api, err := registryClient.GetApi(ctx, &rpc.GetApiRequest{
+		Name: apiName,
+	})
+	if !reflect.DeepEqual(api.Labels, map[string]string{"a": "1", "b": "2"}) {
```
seaneganx, line 104: The `cmp` package is preferred for comparison in tests. Using the `Diff` function we can perform the same check but provide informative error messages indicating what was expected but missing, correct, or unexpected but present. [Example](https://github.com/apigee/registry/blob/78e8f84c190e1d42f00441281f99bb434c96532f/cmd/control_loop/controller/controller_test.go#L187).

cmd/registry/cmd/label_test.go
```diff
@@ -0,0 +1,199 @@
...
+		t.Errorf("labels incorrectly set %+v", api.Labels)
+	}
+	// Remove one label and overwrite the other.
```
seaneganx, line 107: These seem like separate test cases. Consider using table-driven tests for cases that have the same parameters and verification (probably one table per resource type). Alternatively but less extensible, these should at least be wrapped in different subtests using `t.Run("the comment description", func(t *testing.T) { ... })`. More info [here](https://blog.golang.org/subtests).

seaneganx (approved):
Added some comments on the label files, but the feedback can be applied more generally to annotations as well. LGTM, but I don't think the BUILD.sh modifications or reflect.DeepEqual calls should be included.

cmd/registry/cmd/label.go
```diff
@@ -29,157 +29,67 @@ import (
 	"google.golang.org/genproto/protobuf/field_mask"
 )
 
-var labelFilter string
-var labelOverwrite bool
-var labelsToSet map[string]string
-var labelsToClear []string
+var labelKeyFilter string
+var labelKeyOverwrite bool
```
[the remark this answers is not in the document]
timburks: Fixed, and as suggested, flags are parsed in the command Run function.

cmd/registry/cmd/label.go
timburks: done

cmd/registry/cmd/label.go
```diff
@@ -29,157 +29,67 @@ import (
...
+var labelKeysToSet map[string]string
+var labelKeysToClear []string
+
+const labelFieldName = "labels"
+const labelCommandName = "label"
+
+var labelCmd = labelCommand()
```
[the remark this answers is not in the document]
timburks: It didn't at the time, but I think it does now - see the latest iteration of the label.go and annotate.go declarations and the corresponding tests - I found during testing that values of certain flags would "stick" when the command was invoked multiple times from tests (particularly for the optional boolean "overwrite" flag). A handler constructor allows us to construct multiple instances of a command handler for use in tests and also moves us closer to being able to have a constructor for the root command (which would allow us to handle multiple commands in parallel, once all the global flags were removed)

cmd/registry/cmd/label_test.go
timburks: thanks, I fixed this in the crud tests as well

cmd/registry/cmd/label_test.go
timburks: I've restructured these to follow the table-driven style

containers/BUILD.sh
```diff
@@ -21,17 +21,18 @@
...
+TARGET=dev
 
 if [[ $TARGET == "dev" ]]
 then
   # If TARGET is specified as "dev", a minimal set of containers are built.
   # Container names have the suffix "-dev".
   SUFFIX="-$TARGET"
-  CONTAINERS=("registry-server" "authz-server" "registry-tools")
+  #CONTAINERS=("registry-server" "authz-server" "registry-tools")
+  CONTAINERS=("authz-server")
```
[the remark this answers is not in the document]
timburks: no - thanks

cmd/registry/cmd/annotate_test.go
```diff
@@ -0,0 +1,180 @@
...
+			map[string]string{"a": "3"}},
+		{"changing an annotation without --overwrite should be ignored",
+			[]string{"a=4"},
+			map[string]string{"a": "3"}},
+	}
+	// test annotations for APIs.
+	for _, tc := range testCases {
+		cmd := &cobra.Command{}
+		cmd.SetArgs(append([]string{"annotate", apiName}, tc.args...))
+		annotateCmd(cmd)
```
timburks, line 116: @seaneganx see these three lines (114-116) for an alternate way to construct command handlers. I don't think this would be the final form - eventually I think it would be good to have a rootCmd() function that explicitly builds the full command tree instead of relying on init methods. (The init methods make commands easy to add locally but they don't allow us to construct multiple identical rootCmd() trees, which would help in testing and possibly in the implementation of the controller (suppose the controller used worker goroutines to run multiple commands in parallel).

A future revision might replace these three lines with 
```
cmd := rootCmd(append([]string{"annotate", apiName}, tc.args...)
```
where rootCmd automatically includes the annotateCmd (and all others).

seaneganx: Agreed, I actually started poking around with that idea yesterday. I didn't think about accepting arguments in the constructors. It might be convenient, but we still have cmd.SetArgs to achieve the same thing more clearly?

The pattern I came up with yesterday took advantage of packages more than our existing solution, so creating the annotate command was a function called `annotate.Command()` rather than `cmd.annotateCmd(parent *cobra.Command)`. With that approach the child is isolated until a parent command attaches it, allowing us to avoid the mandatory "parent" attachment if we want to test in isolation.

I can put up a draft PR if you're interested.

cmd/registry/cmd/annotate_test.go
```diff
@@ -0,0 +1,180 @@
...
+	}{
+		{"add some annotations",
+			[]string{"a=1", "b=2"},
+			map[string]string{"a": "1", "b": "2"}},
+		{"remove one annotation and overwrite the other",
+			[]string{"a=3", "b-", "--overwrite"},
+			map[string]string{"a": "3"}},
+		{"changing an annotation without --overwrite should be ignored",
+			[]string{"a=4"},
+			map[string]string{"a": "3"}},
```
seaneganx, line 110: Nit: Adding field names to table-driven test cases can aid readability quite a bit.

cmd/registry/cmd/annotate_test.go
```diff
@@ -0,0 +1,180 @@
...
+	var err error
+
+	projectId := "annotate-test"
+	projectName := "projects/" + projectId
+	apiId := "sample"
+	apiName := projectName + "/apis/" + apiId
+	versionId := "1.0.0"
+	versionName := apiName + "/versions/" + versionId
+	specId := "openapi.json"
+	specName := versionName + "/specs/" + specId
```
seaneganx, line 39: Consider using a const block for the IDs and other "irrelevant" test data where the contents aren't a key piece of what's being tested. 

Style for initialisms should be to [maintain uppercase](https://github.com/golang/go/wiki/CodeReviewComments#initialisms). Does `golint` warn you about these? It usually does for me.
```
const (
    projectID = "annotate-test"
    apiID = "sample"
    versionID = "1.0.0"
    specID = "openapi.json"
    openAPIv3 = "application/x.openapi;version=3.0.0"
    specContents = `{"openapi": "3.0.0", "info": {"title": "test", "version": "v1"}, "paths": {}}`
)
```

Instead of discarding the created project, api, version, and spec you can keep the result and use GetName() in the tests. You might not need custom resource IDs at all actually. At a glance it looks like the tests will work even with generated IDs.

seaneganx (approved):

cmd/registry/cmd/annotate_test.go
timburks: done - I think it's arguably helpful at this small scale, but I'd support this as an general style for all our tests, and we might also want to standardize on some field names (optional todo)

cmd/registry/cmd/annotate_test.go
timburks: Done (partly) - I've updated to the const style and initialisms but kept the custom ids (will ping you about that)

cmd/registry/cmd/annotate_test.go
```diff
@@ -0,0 +1,178 @@
...
+			args:     []string{"a=3", "b-", "--overwrite"},
+			expected: map[string]string{"a": "3"}},
+		{comment: "changing an annotation without --overwrite should be ignored",
+			args:     []string{"a=4"},
+			expected: map[string]string{"a": "3"}},
+	}
+	// test annotations for APIs.
+	for _, tc := range testCases {
+		cmd := annotateCmd()
+		cmd.SetArgs(append([]string{apiName}, tc.args...))
```
timburks, line 116: @seaneganx Notice this got simpler when I found that we don't need a rootCmd to invoke the subcommand. (This assumes that we're not using any inherited options, which might be a problem for some current or future commands, but we could address that by constructing the handler with `rootCmd()` instead.

cmd/registry/cmd/annotate.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/spf13/cobra"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+const annotateFieldName = "annotations"
+const annotateCommandName = "annotate"
+
+func init() {
+	rootCmd.AddCommand(annotateCmd())
+}
```
timburks, line 37: @seaneganx Just reiterating that I think it would be good for future refactoring to eliminate use of init to build the command hierarchy. It's a bit more verbose, but I think it would be better for the parent command to explicitly add its children, e.g. the `rootCmd()` constructor would include line 36 to add the annotate subcommand.

PR #174, opened 2021-06-07
Add CSV import command to registry tool

seaneganx:
* Add file seeding package to test directory

### DIFF
--- a/cmd/registry/cmd/upload-csv.go
+++ b/cmd/registry/cmd/upload-csv.go
@@ -1,0 +1,256 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"encoding/csv"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	rpcpb "github.com/apigee/registry/rpc"
+	"github.com/spf13/cobra"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func init() {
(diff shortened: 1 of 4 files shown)


cmd/registry/cmd/upload-csv.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return err
+	}
+
+	spec, err := t.client.CreateApiSpec(t.ctx, &rpcpb.CreateApiSpecRequest{
+		Parent:    version.GetName(),
+		ApiSpecId: t.specID,
+		ApiSpec: &rpcpb.ApiSpec{
+			// TODO: How do we choose a mime type?
+			MimeType: core.OpenAPIMimeType("+gzip", "3.0.0"),
```
seaneganx, line 237: I'm not sure how to approach this. Should we make a guess by checking if the file is "openapi.yaml" or ends with ".proto" and leave it blank if we're not sure?

cmd/registry/cmd/upload-csv_test.go
```diff
@@ -0,0 +1,232 @@
...
+					Path: fmt.Sprintf("%s/specs.csv", root),
+					Contents: []byte(strings.Join([]string{
+						"api_id,version_id,spec_id,filepath",
+						fmt.Sprintf("cloudtasks,v2beta2,openapi.yaml,%s/cloudtasks/v2beta2/openapi.yaml", root),
+						fmt.Sprintf("cloudtasks,v2,openapi.yaml,%s/cloudtasks/v2/openapi.yaml", root),
+						fmt.Sprintf("datastore,v1beta1,openapi.yaml,%s/datastore/v1beta1/openapi.yaml", root),
+						fmt.Sprintf("datastore,v1,openapi.yaml,%s/datastore/v1/openapi.yaml", root),
+					}, "\n")),
+				},
+			},
```
seaneganx, line 84: After setting up these tests to write the file contents to a temporary directory as part of their setup I think I prefer the option of having fixed files in a test data directory.

If we're taking that approach I think it would be good to reconsider the location of cobra command tests. Since we're able to execute against the root command with provided arguments I think we can have an external test package that lives in a sibling directory like `cmd/registry/test`.

timburks (approved):

cmd/registry/cmd/upload-csv_test.go
seaneganx: I think reorganizing in a later PR is the way to go. That way it's one non-functional change instead of a mix of reorganizing and functional additions.

cmd/registry/cmd/upload-csv_test.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"google.golang.org/api/iterator"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+const (
+	gzipOpenAPIv3 = "application/x.openapi+gzip;version=3.0.0"
+	testProject   = "googleapis.com"
```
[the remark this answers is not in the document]
seaneganx: Done

PR #175, opened 2021-06-07
Revisit pointers in controller.go

shrutiparabgoogle:
Refactored the usage of pointers in `controller.go`


timburks (approved):
Reviewing this as a refactoring that removed some unnecessary pointers => LGTM

PR #181, opened 2021-06-09
add full text search when using Postgres

kenigeer:
Search is primarily on descriptions for Apis and Discovery specs (including parameters, nested resource methods, and schemas including enums).

Plain text query returns matching keys with excerpt

    apg registry search-all --q "draft mode"

### DIFF
--- a/cmd/registry/core/gzip.go
+++ b/cmd/registry/core/gzip.go
@@ -34,7 +34,7 @@ func GZippedBytes(input []byte) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-// GUnzippedBytes uncompresses a slice of bytes.
+// GUnzippedBytes decompresses a slice of bytes.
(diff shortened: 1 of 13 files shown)


timburks (commented):
Thanks for this prototype - this is interesting (and in some ways elegant) in that it tightly integrates search into the core API. I had been thinking about search as more of an add-on application, for a few reasons:
1. building indexes might be too slow to put on the main API path
2. for various reasons, we might want the ability to rebuild indexes and would want a global index-builder (which we might later refactor into an incremental one) - such reasons might include changes to the indexing details including user-specified index configuration
3. we might want an expansion capability to allow indexing of currently-unknown spec formats, particularly for users that have their own internal API description formats

So I wonder if it might be optimizing for elegance too soon to index directly in the CRUD operations - let's discuss!

timburks:
Thanks for this, @kenigeer! Moving an updated version of this to https://github.com/apigee/registry-experimental/pull/15

PR #183, opened 2021-06-10
Move controller to the registry tool and make use of cobra framework

shrutiparabgoogle:


cmd/registry/cmd/controller-update.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+var controllerUpdateCmd = &cobra.Command{
+	Use:   "update",
+	Short: "Generate a list of commands to update the registry state (experimental)",
+	Run: func(cmd *cobra.Command, args []string) {
+		var err error
+
+		flagset := cmd.LocalFlags()
+		manifestPath, err := flagset.GetString("manifest_path")
```
timburks, line 36: If this is always required, could we make it a positional argument? I'd like to follow kubectl conventions for style; that suggests using FILENAME in the usage string. You can add that by changing 
```
Use: "update"
```
to 
```
Use: "update FILENAME"
```
on line 30 above.

timburks (approved):

cmd/registry/cmd/controller-update.go
shrutiparabgoogle: Done

PR #186, opened 2021-06-15
Refactor cmd/ directory

seaneganx:
Experimenting with different code organization for the cmd/ directory to increase isolation and decrease reliance on the `core` package for shared functionality. Related commands are grouped as packages so they have their own namespace and ability to share utility functions. In most cases this means subcommands are grouped into packages, but some situations like `registry label` and `registry annotate` might make sense to share a package.

These changes are probably easier to follow by looking at individual commits.

Initial draft:
* Restructure registry CLI into subcommand packages
* Fix lint errors + import names …
* Move sheet export functionality out of core
* Move YAML export functionality out of core


timburks:
As we've discussed, I generally agree with this restructuring, particularly looking forward to a `rootCmd()` constructor for reentrant command handlers. I don't think it's necessary to move everything out of `core` for this to be successful, so wherever something is used by more than one command, I think it would be fine (better?) to leave it in `core`, at least in this PR, and emptying `core` is a relatively easy later-stage tweak that we can do after the major rework of `cmd`.

Issue #147, opened 2021-05-18
Remove storage layer GORM abstractions 


seaneganx:
We have custom-built interfaces to support runtime configurable usage of different storage frameworks, but no longer support multiple storage frameworks since #130 removed datastore from the project. These abstractions only obscure how GORM is being used for storage now.

Some of the types that implement these abstractions are still useful, e.g. Iterator is a helpful abstraction that makes it easier to handle listing operations. 

Action items:
1. Remove the interfaces defined in `storage.go`.
2. Use GORM query builder directly - Remove the custom `Query` struct.
3. Relocate the `dao` and `gorm` packages into the `storage` directory and consider merging them.

PR #191, opened 2021-06-17
Follow AIP-134 guidance on Update methods

seaneganx:
The field mask functionality doesn't seem like it's very relevant to the models package anymore, but I think it's best to move that at a later date if desired. The models package in general feels to me like it should live behind the abstraction of the DAO, allowing the API handlers to work with protos without also handling conversion to the storage format.


timburks:
LGTM. I believe we agreed when the `dao` package was introduced that it would be temporary and that we would merge `models` into it, delete `models`, and rename `dao` to `models`. Regardless, I think further refactoring is fine (but not necessarily the highest priority).

timburks (approved):

seaneganx:
I recall having a similar discussion, but I don't remember discussing the `models` package in particular. There's an open issue #147 that I think may have came from that discussion, and it suggests merging `storage` with `dao` as you're describing. I agree with all of the above though, having `models`, `storage`, and `dao` together as a single package would be nice but doesn't seem urgent.

PR #192, opened 2021-06-17
Differ

Jamaalbwells:
Added Stability Protobuf
Added Differ Module to Registry


google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,88 @@
...
+/* This represents the stability info for a API */
+message APIStability {
+  Stats stats = 1;
+  Metrics metrics = 2;
+  // the filter used on the API
+  string filter_used = 3;
+}
+
+/* This represents the stability info for a Project (Catalog)*/
+message ProjectStability {
```
timburks, line 83: We've separately discussed the question of whether we should have four messages or just one for this.  At this stage, there's no difference between these messages.

timburks (changes requested):
Could we merge this to a branch other than main? I've got some reservations about this (noted below) but don't want that to block progress.

cmd/stability/differ.go
```diff
@@ -0,0 +1,77 @@
...
+		PathFilter:         diffOptions.filter,
+		PathPrefix:         diffOptions.prefix,
+	}, spec1, spec2)
+	if err != nil {
+		fmt.Printf("diff failed with %v", err)
+		return nil, err
+	}
+	var yaml_bytes = getYAML(diffReport)
+	//Convert Yaml Diff to string.
+	stringYamlDiff := string(yaml_bytes)
```
seaneganx, line 56: This comment seems to just restate what the code is doing. Code comments are great for explaining _why_ something is happening, while the code itself should be understandable enough to explain _what_ is happening without commentary.

Since the line of code is so clear already and the variable is only used in one location, consider doing the conversion in-line when you set the struct field instead.

cmd/stability/differ_test.go
```diff
@@ -0,0 +1,118 @@
...
+                required:
+                    deleted:
+                        - name
+                properties:
+                    added:
+                        - added
+                        - age
+`
+	expected_name_1 = "Test"
+	expected_name_2 = "Test"
```
seaneganx, line 30: It's difficult to understand what these constants are used for given their current names. Can the names be more descriptive, or can they be defined in the tests instead so we have context to help understand?

Also, recall that names should be [camelCase or MixedCase](https://golang.org/doc/effective_go#mixed-caps). These tests have lots of underscores that make it harder to write the variable names.

cmd/stability/differ_test.go
```diff
@@ -0,0 +1,118 @@
...
+		t.FailNow()
+	}
+	diff_proto := &pb.TextDiff{}
+	err = proto.Unmarshal(seralized_diff_proto, diff_proto)
+	if err != nil {
+		t.Logf("Marshalling error: %v", err)
+		t.FailNow()
+	}
+	if diff_proto.Diff != expected_diff_proto.Diff {
+		t.Errorf("The Diffs were not equal |\n" +
```
seaneganx, line 62: Instead of saying what they were, the test failure can be more descriptive by saying what was expected. For example, `t.Errorf("GetDiff(%q, %q) returned unexpected diff:\nwant: %s\ngot: %s", yamlFile, yamlFile2, expected_diff_proto.Diff, diff_proto.Diff)`.

cmd/stability/differ_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if diff_proto.Spec != expected_diff_proto.Spec {
+		t.Errorf("The Spec names were not equal |\n" +
+				"want: %v \ngot: %v", expected_diff_proto.Spec, diff_proto.Spec)
+	}
+	if diff_proto.SpecRevision != expected_diff_proto.SpecRevision {
+		t.Errorf("The Revision names were not equal |\n" +
+				"want: %v \ngot: %v", expected_diff_proto.SpecRevision,
+				diff_proto.SpecRevision)
+	}
```
seaneganx, line 73: It looks like the three equality checks being done here can use the [go-cmp/cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) `Diff` function instead. It will provide better formatting for this kind of test verification.

seaneganx (commented):
Lots of comments about Go specific stuff. Take a look at these and address what you can or what you think is best, then I'll take another pass to discuss "bigger picture" items.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,88 @@
...
+    bool breaking_change = 3;
+  }
+  repeated ChangeLine change_line = 1;
+  int64 change_count = 2;
+  int64 breaking_change_count = 3;
+}
+
+/* This represents the total stats of a resource.
+These stats are computed for all resources. */
+message Stats {
```
timburks, line 41: This is a very generic (too generic) name to use in an "applications" package that contains protos associated with other analysis capabilities. I don't think there's enough here to create a new package, so I'd like to look for message names that include "Stability"

Looking through the messages in this file, I think that "Stats" and "Metrics" could be merged into a single message and if a "filter" field was added, this merged message could replace the last six protos in this file.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,88 @@
...
+  int64 changes = 1;
+  int64 breaking_changes = 2;
+  int64 revisions = 3;
+  int64 featurers_added = 4;
+  int64 features_removed = 5;
+}
+/* This represents computed metrics for a resource.
+These metrics are computed for all resources */
+message Metrics {
```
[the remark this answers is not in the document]
seaneganx: This seems like it depends on the components and their APIs that don't exist yet. IIRC from the design, stats are computed in a different component (package?) of the system, then metrics are computed from the stats at a later stage. Something like `func ComputeStats(d DiffAnalysis) Stats` and `func ComputeMetrics(s Stats) Metrics` might exist.

If these were merged into a single message, would we go from DiffAnalysis directly to Stats+Metrics in one call? If that's the plan we might not need any of the stats fields at all since they're just intermediate values to derive the metrics.

seaneganx (commented):
Took another pass to look at the proto. Much of the discussion seems to depend on how the components of the system will be organized. It might be best to only add the TextDiff message in this PR because there's a lot of unknowns around how the remaining messages will be used.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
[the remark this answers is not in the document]
timburks: I could see wanting to keep these messages distinct so they could be attached as separate artifacts, which makes it possible to use the controller (with an appropriate manifest) to go from "Stats" to "Metrics", but if it's possible to compute this in one step, I think one step is probably better. 

Stepping back, I don't think "stats" and "metrics" will be well-understood descriptive terms for these chunks of data. Here "stats" seems to be the more basic (primitive) values, but "statistics" generally refers to analysis of populations of data. I think people would have trouble intuitively understanding that metrics are derived from stats and not the other way around. Bundling these together is one way to resolve that problem.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
timburks: The "applications" package itself is arbitrary and in some ways temporary (v1alpha1) - I would expect that some applications might grow out of it, but (very subjectively) I think that stability might be one of many factors that could be used to compute an "API Score" and that having these factors in a single package might aid clarity. (I guess I'm saying that I'm not sure that stability belongs in its own package and that even though "applications" probably isn't right either, at least it forces us to think about stability alongside other quantities that might be of interest)

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,88 @@
+syntax = "proto3";
+package google.cloud.apigee.registry.applications.v1;
+// [START go_declaration]
+option go_package = "github.com/apigee/registry/rpc;rpc";
```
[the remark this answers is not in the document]
timburks: I'm open to changing that - this was originally just for convenience. (But if we do that, I'd like to make the change in a separate PR and follow convention until we do)

Also, I just today noticed that this file is in a new applications/v1 directory (the other protos are in applications/v1alpha1). We might want to move all the application protos to v1 later, but for consistency, I think we should keep them together in v1alpha1

PR #204, opened 2021-06-23
Proto definition for manifest

shrutiparabgoogle:
Initial proto definition for controller manifest.

### DIFF
--- a/google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
+++ b/google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
@@ -1,0 +1,47 @@
+syntax = "proto3";
+
+package google.cloud.apigee.registry.applications.v1alpha1;
+
+option java_package = "com.google.cloud.apigee.registry.applications.v1alpha1";
+option java_multiple_files = true;
+option java_outer_classname = "RegistryManifestProto";
+option go_package = "github.com/apigee/registry/rpc;rpc";
+
+// Manifest represents the dependencies of a list of generated resources that are stored
+// in a registry and is used by the controller to keep these resources up-to-date.
+message Manifest {
+  // List of Generated resources.
+  repeated GeneratedResource generated_resources = 1 [
+		(google.api.field_behavior) = REQUIRED
+	];
+}
+
+message GeneratedResource {
+	// Pattern for one particular or a group of api/version/spec/artifact
+	// Format: 
+	// apis/{api}/versions/{version}/specs/{spec}/artifacts/{artifact}
+	// apis/-/versions/-/specs/-/artifacts/-
+	string pattern = 1 [
+		(google.api.field_behavior) = REQUIRED
+	];
+	string filter = 2;
+	repeated Dependency dependencies = 3;
+	// action Can contain references to both the resource and dependencies
+	// Example: "compute complexity $dependency0 $resource"
+	string action = 4 [
+		(google.api.field_behavior) = REQUIRED
+	];
+}
+
+message Dependency {
+	// pattern for one particular or a group of api/version/spec/artifact
+	// pattern in a Dependency can contain references to the original resource
+	// Format: 
+	// $resource.api/versions/-/specs/-
+	// $resource.version/specs/-/artifacts/-
+	string pattern = 1 [
+		(google.api.field_behavior) = REQUIRED
+	];
+	string filter = 2;
+}
+


google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
```diff
@@ -0,0 +1,36 @@
...
+// is used by the controller to keep the resources up-to-date.
+message Manifest {
+  // List of dependency relations.
+  repeated Resource resources = 1 [
+		(google.api.field_behavior) = REQUIRED
+	];
+}
+
+message Resource {
+	string pattern = 1 [
```
shrutiparabgoogle, line 20: I was wondering if it makes sense to add some validation annotation here to make sure the pattern is a valid query for the registry.

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
```diff
@@ -0,0 +1,36 @@
...
+	];
+	string filter = 2;
+	repeated SourcePattern dependsOn = 3;
+	string updateAction = 4 [
+		(google.api.field_behavior) = REQUIRED
+	];
+}
+
+message SourcePattern {
+	string sourcePattern = 1 [
```
shrutiparabgoogle, line 31: Same here. Validation annotation.

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
```diff
@@ -0,0 +1,36 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "RegistryManifestProto";
+option go_package = "github.com/apigee/registry/rpc;rpc";
+
+// Manifest represents the dependency relation between resources, which 
+// is used by the controller to keep the resources up-to-date.
+message Manifest {
+  // List of dependency relations.
+  repeated Resource resources = 1 [
+		(google.api.field_behavior) = REQUIRED
```
timburks, line 15: There are two different descriptions here in the comment and field name/type: "dependency relations" and "resources". I think we'd get more conceptual clarity if they were consistent. "Resource" seems too generic, because not all resources have dependencies. "Dependency" seems too specific because any of the things listed here might have multiple dependencies. "GeneratedResource" is more descriptive but verbose. I like "Product" but think that it might be dismissed for the same reason that "Target" was. I've also seen "Result" considered, but that seems unclear (a result of what?).

From this, I think "generated resource" might be best - it's verbose but descriptive and accurate.
```
  // List of generated resources.
  repeated GeneratedResource generated_resources = 1;
```

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
timburks: Could we start with a comment showing the expected format? Like [this](https://github.com/apigee/registry/blob/21587ed876e8bd4073334eb8c24eba0fa2a9a871/google/cloud/apigee/registry/v1/registry_models.proto#L88)?

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
```diff
@@ -0,0 +1,36 @@
...
+	];
+}
+
+message Resource {
+	string pattern = 1 [
+		(google.api.field_behavior) = REQUIRED
+	];
+	string filter = 2;
+	repeated SourcePattern dependsOn = 3;
```
timburks, line 24: how about `repeated Dependency dependencies = 3;` ?

timburks: (also FYI here's [style guidance](https://developers.google.com/protocol-buffers/docs/style#message_and_field_names) on message and field names)

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
timburks: I think `pattern` would be better here, and it would be consistent with the definition of `pattern` in the `Resource` (or `GeneratedResource`) message.

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
timburks: Also, I agree that `pattern` is the right name for this. I briefly considered `name_pattern` after finding it [here](https://github.com/googleapis/googleapis/blob/7e91915c247a5c84ab5fcdf11ce2bdf7404f2fcc/google/cloud/binaryauthorization/v1beta1/resources.proto#L101), but that's an isolated instance and `pattern` is used throughout the AIPs and [here](https://github.com/googleapis/googleapis/blob/7cde5d0df08801e00fc45df1546942aa2692d5c3/google/api/resource.proto#L218), which seems more authoritative.

timburks (changes requested):
Thanks for this draft - I have comments about naming but not structure (the structure looks good).

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
shrutiparabgoogle: Yes, renamed to `pattern`

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
shrutiparabgoogle: Added some examples in the comment.

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
shrutiparabgoogle: I like `GeneratedResource`.

The only reason I was avoiding multiple words in the name is because we have to reference it in the `update_action` or dependency `pattern` strings. I have used `$resource` and `$dependency` in the examples below. We can revisit if `GeneratedResource` and `$resource` feel a little disconnected.

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
shrutiparabgoogle: Thanks for that link. Renamed the field.

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+
+package google.cloud.apigee.registry.applications.v1alpha1;
+
+option java_package = "com.google.cloud.apigee.registry.applications.v1alpha1";
+option java_multiple_files = true;
+option java_outer_classname = "RegistryManifestProto";
+option go_package = "github.com/apigee/registry/rpc;rpc";
+
+// Manifest represents the dependency relation between resources, which
```
timburks, line 10: minor tweak, could we say this? "Manifest represents the dependencies of a list of generated resources that are stored in a registry and is used by the controller to keep these resources up-to-date."

timburks (approved):
Just suggesting a couple of minor changes - if you agree with them, this all LGTM

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
```diff
@@ -0,0 +1,47 @@
...
+	// apis/{api}/versions/{version}/specs/{spec}/artifacts/{artifact}
+	// apis/-/versions/-/specs/-/artifacts/-
+	string pattern = 1 [
+		(google.api.field_behavior) = REQUIRED
+	];
+	string filter = 2;
+	repeated Dependency dependencies = 3;
+	// update_action Can contain references to both the resource and dependencies
+	// Example: "compute complexity $dependency0 $resource"
+	string update_action = 4 [
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
shrutiparabgoogle: Done

PR #206, opened 2021-06-24
Refactor server setup

seaneganx:
The main purpose of this change is to transfer control of the server networking to the user of the server package instead of the RunServer function. For testing it's useful because it allows listeners to be created on port 0 which automatically assigns an open port and allows the code running the server to inspect what that port is. This enables multiple servers to run simultaneously without assigning a pool of fixed ports to use.


server/server.go
```diff
@@ -53,19 +53,19 @@ const (
 
 // RegistryServer implements a Registry server.
 type RegistryServer struct {
-	database            string
-	dbConfig            string
-	enableNotifications bool
```
timburks: Please link to a style guide or precedent for this (and similar name changes)

timburks (commented):
I'm neutral about these changes but not ready to agree with the justification: "This enables multiple servers to run simultaneously without assigning a pool of fixed ports to use." I'd like to find ways to satisfy our test isolation goals without requiring multiple concurrent servers or even multiple backend database files - see #210 for a related experiment/suggestion.

seaneganx (commented):
>I'm neutral about these changes but not ready to agree with the justification: "This enables multiple servers to run simultaneously without assigning a pool of fixed ports to use."

I didn't intend for that to be the justification (i.e. reason to merge) for this change. My justification is that this PR improves our server by increasing modularity and detecting configuration errors before starting the server instead of while handling requests, among other benefits.

>I'd like to find ways to satisfy our test isolation goals without requiring multiple concurrent servers or even multiple backend database files - see #210 for a related experiment/suggestion.

I think we should have this discussion, but I think this PR is worth merging independent of that discussion. Let's move forward with this and keep the discussion about potential next steps on #210 and #208.

server/server.go
timburks: SGTM, thanks for adding this

timburks (approved):

PR #209, opened 2021-06-28
Add registry upload manifest command

shrutiparabgoogle:


cmd/registry/controller/manifest-reader.go
```diff
@@ -18,6 +18,9 @@ import (
 	"io/ioutil"
 	"fmt"
 	"gopkg.in/yaml.v2"
+	yaml2 "github.com/ghodss/yaml"
```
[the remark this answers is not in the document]
timburks: (since there's no direct YAML-to-proto unmarshaling, we're probably going to be using this pattern often)

cmd/registry/cmd/upload-manifest.go
```diff
@@ -0,0 +1,63 @@
...
+
+)
+
+func init() {
+	uploadCmd.AddCommand(uploadManifestCmd)
+	uploadManifestCmd.Flags().String("project", "", "ProjectID this manifest should be associated with.")
+}
+
+var uploadManifestCmd = &cobra.Command{
+	Use:   "manifest",
```
timburks, line 21: We can add more information to the usage message to indicate that manifest_path is a required argument. See https://github.com/apigee/registry/blob/bf3525dcc6b7637453ca28035f2a9f1fc85e6d36/cmd/registry/cmd/label.go#L41 and https://github.com/apigee/registry/blob/bf3525dcc6b7637453ca28035f2a9f1fc85e6d36/cmd/registry/cmd/upload-csv.go#L44 for examples (also note that we have some style inconsistencies to resolve)

cmd/registry/cmd/upload-manifest.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/apigee/registry/cmd/registry/controller"
+	"github.com/apigee/registry/rpc"
+	"google.golang.org/protobuf/proto"
+	"github.com/apigee/registry/cmd/registry/core"
+
+)
+
+func init() {
+	uploadCmd.AddCommand(uploadManifestCmd)
+	uploadManifestCmd.Flags().String("project", "", "ProjectID this manifest should be associated with.")
```
timburks, line 17: For consistency across the registry command, please use `project` when a full project resource name is required (`projects/PROJECT_ID`) and `project_id` when we are only expecting the project ID.

timburks (commented):
Generally LGTM, just a few questions and observations.

cmd/registry/controller/test/manifest.yaml
```diff
@@ -12,32 +12,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-project: "controller-demo"
-manifest:
-- resource: apis/-/versions/-/specs/-/artifacts/lint-gnostic
+name: "test-manifest"
```
seaneganx, line 15: If we're changing the structure of the manifest, should the other test files change too? Or what's the difference between those manifests with "project" fields and these ones with "name" fields?

cmd/registry/cmd/upload-manifest.go
seaneganx: It seems like `project_id` is the more common pattern right now, but it seems rare that we would ever refer to an API, Version, Spec, or Artifact by its ID alone. That said, maybe we should lean toward using full resource names.

seaneganx (commented):
I recommend using `gofmt` to format your code before creating PRs. It's likely your editor has some kind of plugin that can be set up to do this automatically.

cmd/registry/controller/manifest-reader.go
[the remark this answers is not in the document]
shrutiparabgoogle: Yes, this is a workaround for yaml to proto conversion.

cmd/registry/cmd/upload-manifest.go
shrutiparabgoogle: Done.

cmd/registry/cmd/upload-manifest.go
shrutiparabgoogle: Done, I changed this to `project_id` to be consistent.

cmd/registry/controller/test/manifest.yaml
shrutiparabgoogle: Yes, the other manifests should also change, the controller tests are currently dependent on those manifests. I am going to cleanup and update those as part of the `controller` command cleanup. I wanted to keep this change short and not include too many cleanup changes.

shrutiparabgoogle:
> I recommend using `gofmt` to format your code before creating PRs. It's likely your editor has some kind of plugin that can be set up to do this automatically.

Done.

seaneganx (approved):

PR #210, opened 2021-06-29
Experimenting with server test setup (for discussion only)

timburks:
*Not for merging*

While playing around with the helper code in server_test.go, I found/confirmed that tests run sequentially and can share a common database if we clean up between tests, either by deleting the sqlite database file or by deleting all of the projects in the test database. If we maintain this discipline, it would be easy to test against a Postgres database (tests wouldn't require CREATE DATABASE access) and it probably also wouldn't be hard to write a proxy server that sent requests to a remote server. Let's discuss here or in #206

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -15,15 +15,65 @@
 package server
 
 import (
+	"context"
 	"fmt"
+	"io/ioutil"
 	"testing"
+
+	"github.com/apigee/registry/rpc"
 )
+
+var defaultServer *RegistryServer
+var dname string
+
+func init() {
+	var err error
+	dname, err = ioutil.TempDir("", "testdb-")
+	if err != nil {
+		panic(err)
+	}
+
+	defaultServer = &RegistryServer{
+		// to allow the server tests to run with Postgres
+		// (and other backends including CloudSQL Postgres),
+		// these can be set using the environment variables
+		// that are defined in config/registry.yaml and
+		// (if undefined), default to the following values:
+		database:     "sqlite3",
+		dbConfig:     fmt.Sprintf("%s/registry.db", dname),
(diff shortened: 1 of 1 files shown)


seaneganx:
After our discussion yesterday I came up with a similar experiment, although intentionally using a mutex to control access to the PostgreSQL backed server. Tests within the same package are definitely executed serially unless marked with `t.Parallel()`, but IIRC tests across multiple packages are executed in parallel.

How configurable do we want the test server to be? Should tests accept a config path similar to the flag `-c` on `registry-server`? Maybe just some option to use PostgreSQL instead of SQLite? Do we want to put in the effort to make our `cmd/` tests run against multiple server types?

timburks:
If the tests run sequentially, I don't think the mutex is needed for the Postgres DB - wouldn't that only be preventing tests from concurrently interfering with each other? Tests across multiple packages should use project isolation (imo) and this is a similar issue for SQLite.

To simplify certain container deployments (particularly CloudRun), we [defined some environment variables in the default registry.yaml](https://github.com/apigee/registry/blob/3e01861de1645d91966ab5be1020b277ac9ee315/config/registry.yaml#L17) that I think would be fine to also use in the tests.

For the cmd tests, it's a bit redundant to run them for both backends, but I think testing with different backends is similar to testing on different platforms (amd64/arm64 or OSX/iOS/Linux) - typically CI is set up to run tests on each supported platform. Developers typically test on one before committing but have the ability to test on any when problems show up in CI. 

So I'd suggest that we maintain our present expectation to test with SQLite pre-commit (or Postgres when needed, for example for search) and just set up CI to run both. I think that also minimizes the changes needed.

seaneganx:
Closing because #234 implements this idea.

PR #211, opened 2021-06-29
Remove retry on unknown errors

seaneganx:
Errors that don't have a specified status code are considered "unknown"
and shouldn't be retried because we don't have reason to believe they
will succeed in future attempts. By retrying until timeout we lose the
information provided by the unknown error, and instead see the result of
a timeout.


seaneganx:
Fixes #211

timburks (approved):
Thanks, this should help a lot with debugging

PR #214, opened 2021-06-30
Add suport for 'registry resolve' command

shrutiparabgoogle:
Summary of changes:
- support for `registry resolve` command
- Cleanup `registry controller` commands
- start including filter attribute in `generated_resource` definition


cmd/registry/cmd/resolve/resolve.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"log"
+
+	"github.com/apigee/registry/cmd/registry/controller"
+	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/connection"
+	"github.com/spf13/cobra"
+)
+
+func Command() *cobra.Command {
```
seaneganx, line 28: When you merge the main branch this should use the new pattern of including context as an argument here.

cmd/registry/cmd/resolve/resolve.go
```diff
@@ -0,0 +1,88 @@
...
+	cmd := &cobra.Command{
+		Use:   "resolve MANIFEST_RESOURCE",
+		Short: "resolve the dependencies and update the registry state (experimental)",
+		Args:  cobra.MinimumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			var err error
```
seaneganx, line 34: Nit: I don't think this is needed since it gets redeclared at line 42.

cmd/registry/controller/controller.go
```diff
@@ -30,40 +31,48 @@ type ResourceCollection struct {
...
-	dependencyMaps := make([]map[string]ResourceCollection, 0, len(entry.Dependencies))
-	for _, d := range entry.Dependencies {
-		dMap, err := generateDependencyMap(ctx, client, resourcePattern, d.Source, d.Filter)
+	resourcePattern := fmt.Sprintf("projects/%s/%s", projectID, resource.Pattern)
+	dependencyMaps := make([]map[string]ResourceCollection, 0, len(resource.Dependencies))
+	for _, d := range resource.Dependencies {
+		dMap, err := generateDependencyMap(ctx, client, resourcePattern, d.Pattern, d.Filter)
 		if err != nil {
-			log.Printf("Encountered error during updateResources().\n Error: %s\n Skipping entry %+v", err.Error(), entry)
+			log.Printf("Encountered error during generateDependencyMap().\n Error: %s\n Skipping resource %+v", err.Error(), resource)
```
seaneganx, line 61: To avoid mismatched messages it might be a better to describe the situation, e.g. "error while generating dependency map" instead of using a specific function name that can be renamed globally without the message being noticed.

seaneganx (approved):

cmd/registry/core/resources.go
```diff
@@ -24,3 +26,11 @@ func ParentAndIdOfResourceNamed(name string) (string, string) {
 	last := len(parts) - 1
 	return strings.Join(parts[0:last-1], "/"), parts[last]
 }
+
+func ProjectIdOfResource(name string) (string, error) {
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

I agree, there are a lot of regex functions too in the controller which do a lot of parsing on the resource names.  That can also go int the common name functionality.

cmd/registry/core/resources.go
```diff
@@ -24,3 +26,11 @@ func ParentAndIdOfResourceNamed(name string) (string, string) {
...
+	parts := strings.Split(name, "/")
+	if len(parts) > 1 {
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

cmd/registry/cmd/resolve/resolve.go
shrutiparabgoogle: Done

cmd/registry/cmd/resolve/resolve.go
shrutiparabgoogle: Done

cmd/registry/controller/controller.go
shrutiparabgoogle: Thanks for this. Done.

PR #215, opened 2021-07-01
Delete children artifacts when ancestor is deleted

### LINKED ISSUE #78: Deleting resources should delete any artifacts associated with them.
For example, deleting a spec (projects/PROJECT_ID/apis/API_ID/versions/VERSION_ID/specs/SPEC_ID) should delete any artifacts under that spec (projects/PROJECT_ID/apis/API_ID/versions/VERSION_ID/specs/SPEC_ID/artifacts/ARTIFACT_ID).

Currently we (crudely) delete all artifacts only when their containing project is deleted.

seaneganx:
Fixes #78

### DIFF
--- a/server/gorm/deletions.go
+++ b/server/gorm/deletions.go
@@ -71,6 +71,7 @@ func (c *Client) DeleteChildrenOfProject(ctx context.Context, project names.Proj
 // DeleteChildrenOfApi deletes all the children of a api.
 func (c *Client) DeleteChildrenOfApi(ctx context.Context, api names.Api) error {
 	for _, entityName := range []string{
+		storage.ArtifactEntityName,
 		models.BlobEntityName,
 		storage.SpecEntityName,
 		storage.VersionEntityName,
@@ -89,6 +90,7 @@ func (c *Client) DeleteChildrenOfApi(ctx context.Context, api names.Api) error {
 // DeleteChildrenOfVersion deletes all the children of a version.
 func (c *Client) DeleteChildrenOfVersion(ctx context.Context, version names.Version) error {
 	for _, entityName := range []string{
+		storage.ArtifactEntityName,
 		models.BlobEntityName,
 		storage.SpecEntityName,
 	} {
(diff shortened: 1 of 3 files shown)


server/gorm/deletions.go
```diff
@@ -71,6 +71,7 @@ func (c *Client) DeleteChildrenOfProject(ctx context.Context, project names.Proj
 // DeleteChildrenOfApi deletes all the children of a api.
 func (c *Client) DeleteChildrenOfApi(ctx context.Context, api names.Api) error {
 	for _, entityName := range []string{
+		storage.ArtifactEntityName,
 		models.BlobEntityName,
```
timburks, line 75: aside: it seems odd that we didn't also move this name into the "storage" package, not urgent

server/gorm/deletions.go
```diff
@@ -71,6 +71,7 @@ func (c *Client) DeleteChildrenOfProject(ctx context.Context, project names.Proj
 // DeleteChildrenOfApi deletes all the children of a api.
 func (c *Client) DeleteChildrenOfApi(ctx context.Context, api names.Api) error {
 	for _, entityName := range []string{
+		storage.ArtifactEntityName,
```
timburks, line 74: aside: I don't think we have a way (in testing) to verify that the blobs associated with specs and artifacts are actually getting deleted - this is less urgent than testing the API but worth noting

timburks (approved):

server/gorm/deletions.go
seaneganx: I think if we want to test this we can do the following.
1. Create an API and child artifact. 
2. Delete the API.
3. Recreate the API. 
4. Read the artifact again.
5. Assert it doesn't exist.

I didn't include this test because it seems like a bug that's at low risk of regression, and has quite low cost even if we did reintroduce the bug.

server/gorm/deletions.go
seaneganx: Just noticed I missed this comment after checking my email. I agree, I missed this one when I moved the others over.

#147 seems like it will cover this.

server/gorm/deletions.go
timburks: That mostly gets it - since artifacts (like specs) store contents in blobs, we might want to also verify that the blob was deleted. Blobs are implementation details and not exposed by the API, so this would be like looking for a memory leak.

server/gorm/deletions.go
timburks: SGTM

server/gorm/deletions.go
seaneganx: Functionally it should be invisible unless it causes issues recreating the artifact with the leftover blob still existing. If we tested directly from the gorm layer this can be easily verified though.

PR #217, opened 2021-07-02
Require custom IDs on resource creation

### LINKED ISSUE #70: Resource creation methods using automatically generated ids treat name collisions as user errors
In the current v1 implementation, unspecified resource ids are replaced with automatically generated ones using substrings of UUIDs: `uuid.New().String()[:8]`.

Name collisions are unlikely but possible - according to [birthday attack analysis](https://en.wikipedia.org/wiki/Birthday_attack), they tend to follow the square root of the number of states, i.e. with 2^32-state ids, we could expect a collision after approximately 2^16 (65536) ids were generated. This is consistent with an experiment (below) using our actual id generation expression.

Currently collisions are reported as "already exists" errors, which suggest that the user specified an invalid ID, when simply retrying is likely to fix the problem.

If we explicitly detected and handled collisions, we might also be able to reduce the length of our automatically-generated identifiers to more user-friendly values.

```
package main

import (
	"fmt"
	"github.com/google/uuid"
)

func findDuplicate() int {
	ids := make(map[string]bool)
	for i := 1; true; i++ {
		x := uuid.New().String()[:8]
		if ids[x] {
			return i
		}
		ids[x] = true
	}
	return 0
}

func main() {
	for i := 0; i < 10; i++ {
		c := findDuplicate()
		fmt.Printf("found duplicate in %d ids\n", c)
	}
}
```

Running the above code once produced:
```
found duplicate in 104690 ids
found duplicate in 67448 ids
found duplicate in 41598 ids
found duplicate in 93047 ids
found duplicate in 97930 ids
found duplicate in 72057 ids
found duplicate in 85917 ids
found duplicate in 98528 ids
found duplicate in 110139 ids
found duplicate in 100775 ids
```

seaneganx:
Fixes #70 by removing the feature entirely. If we don't want to remove the feature, much of this PR is still useful. I think it's a good idea for us to prefer using fixed IDs in our tests so we have the flexibility of removing generated IDs in the future.

### DIFF
--- a/cmd/registry/cmd/export-csv_test.go
+++ b/cmd/registry/cmd/export-csv_test.go
@@ -35,14 +35,23 @@ func TestExportCSV(t *testing.T) {
 	}
 
 	const (
+		projectID = "export-csv-test-project"
 		apiID     = "my-api"
 		versionID = "v1"
 		specID    = "my-spec"
 	)
 
 	// Setup
+	err = client.DeleteProject(ctx, &rpc.DeleteProjectRequest{
+		Name: "projects/" + projectID,
+	})
+	if err != nil {
+		t.Fatalf("Setup: Failed to delete project: %s", err)
+	}
+
 	project, err := client.CreateProject(ctx, &rpc.CreateProjectRequest{
-		Project: &rpc.Project{},
+		ProjectId: projectID,
+		Project:   &rpc.Project{},
 	})
 	if err != nil {
 		t.Fatalf("Setup: Failed to create project: %s", err)
--- a/cmd/registry/cmd/export-csv_test.go
+++ b/cmd/registry/cmd/export-csv_test.go
@@ -25,6 +25,8 @@ import (
 	"github.com/apigee/registry/rpc"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 func TestExportCSV(t *testing.T) {
@@ -45,8 +47,8 @@ func TestExportCSV(t *testing.T) {
 	err = client.DeleteProject(ctx, &rpc.DeleteProjectRequest{
 		Name: "projects/" + projectID,
 	})
-	if err != nil {
-		t.Fatalf("Setup: Failed to delete project: %s", err)
+	if err != nil && status.Code(err) != codes.NotFound {
+		t.Fatalf("Setup: Failed to delete test project: %s", err)
 	}
 
 	project, err := client.CreateProject(ctx, &rpc.CreateProjectRequest{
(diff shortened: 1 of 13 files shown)


cmd/registry/cmd/export-csv_test.go
```diff
@@ -35,14 +35,23 @@ func TestExportCSV(t *testing.T) {
 	}
 
 	const (
+		projectID = "export-csv-test-project"
```
timburks, line 40: Could we choose and document a convention for naming projects in tests? A declared namespace (perhaps using a prefix) would help us (occasionally) run tests against running installations and might also improve predictability and help us reduce project name collisions.

I'd suggest calling this "test-export-csv" or (possibly better) "test-registry-export-csv" since it's testing a subcommand of the registry tool.

timburks (approved):

cmd/registry/cmd/export-csv_test.go
```diff
@@ -45,8 +47,8 @@ func TestExportCSV(t *testing.T) {
 	err = client.DeleteProject(ctx, &rpc.DeleteProjectRequest{
 		Name: "projects/" + projectID,
 	})
-	if err != nil {
-		t.Fatalf("Setup: Failed to delete project: %s", err)
+	if err != nil && status.Code(err) != codes.NotFound {
```
timburks, line 50: LGTM - were tests sometimes failing because these deletes failed to complete?

cmd/registry/cmd/export-csv_test.go
seaneganx: As long as the project names are specific enough not to collide I don't see any additional benefit to documenting and enforcing a common prefix. The project has to be deleted before/after the test run regardless.

If I'm missing the point, can you elaborate on how predictability would increase with a common prefix?

cmd/registry/cmd/export-csv_test.go
seaneganx: They were passing locally because the test doesn't delete the project after it finishes, so this wasn't returning "Not found" on my local runs. When I pushed it up and CI ran it against an empty database, it started to fail.

This happens to me frequently and is only possible because our tests are non-hermetic. As much as I try to stop the server, rebuild, restart it, delete the databases, etc. it's difficult to avoid these issues.

cmd/registry/cmd/export-csv_test.go
timburks: Partly it's an aesthetic preference to add some order and predictability by using a naming convention to help future test authors follow a consistent pattern.

It could also help with automation - we might write helpers that delete all "test-" projects, for example.

If we documented that user projects shouldn't use our test prefix, this would allow people to run tests on running instances, and we might further enforce this by explicitly blocking creation of "test-" projects unless a header was set.

seaneganx: Makes sense. Having an established pattern makes it easier to follow.

For automation I think it's safer for us to use explicit test/non-prod databases that can be safely wiped out instead of relying on a prefix to indicate which rows are safe to delete. Having the tests connect to a database named `registry_test` makes it unlikely somebody will be upset that its contents are affected by a test run.

This would add a one-time cost to setting up tests to run against Postgres, but it would make this data safety issue extremely obvious to users. I'll think about this idea a bit more because it might solve some of our existing issues with testing too.

PR #218, opened 2021-07-08
Reorganize cmd/ directory structure

seaneganx:
This PR adjusts the directory structure of the cmd/ directory to represent how the subcommands are organized. Other changes will follow in a different PR for ease of review and to maintain git movement history.


cmd/registry/cmd/annotate/annotate.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package cmd
+package annotate
```
timburks, line 15: I don't want to change this in this PR - but for predictability, I think it would be helpful to always give the same name to the file that contains a package's Command() function - "command.go" seems like a good choice for this. Following that pattern, we would probably also rename "root.go" at the top to "command.go". (the current pattern names these files to match the package name, which is consistent but makes it more effort (mental and programmatic) to find all of these files - if they were all named "command.go" for example, we could list them all with "find . -name command.go"

timburks (approved):

cmd/registry/cmd/annotate/annotate.go
seaneganx: Agreed, that sounds like a good way to organize this. Expanding on that, the command file might contain only the very basics of the command, e.g. the Command() function, flag parsing, argument validation, etc. with one or more other files containing the task definitions and other functionality.

timburks: I'd be open to that, but I also like the current pattern where the package contains one file for the main command and one for each (if any) subcommands. I'm sure there will be some examples with additional files containing common code, but I don't feel strongly that we should factor out things that are only used once (by the main command)

PR #219, opened 2021-07-08
Refactor cobra commands into functional constructors

seaneganx:
Blocked by #218 to maintain git file history as much as possible. The diff will be hard to understand until that PR is merged. This change removes package-global variables, updates flag descriptions, addresses lint + staticcheck issues, and moves commands into functional constructors instead of package-level singletons with init function configuration.


timburks (approved):
Looks generally ok - running the demo scripts, I found that some options were marked as required that weren't actually - I think we should drop these requirements and address them individually if needed.

timburks:
LGTM, please merge this

PR #226, opened 2021-07-13
Rename keywords in parser to align with the proto definition

shrutiparabgoogle:


cmd/registry/controller/test/manifest.yaml
```diff
@@ -18,26 +18,26 @@ generated_resources:
   dependencies:
   - pattern: $resource.spec
     filter: "mime_type.contains('openapi')"
-  action: "compute lint $source0 --linter gnostic"
+  action: "compute lint $dependency0 --linter gnostic"
```
timburks, line 21: "$dependency0" is verbose and might seem awkward to users. Would it be clear enough from the context to use $0 here?

timburks: Related: is there a list of all of the $variable definitions? ($dependencyN, $resourceN, etc)

shrutiparabgoogle: I agree "`$dependency` is a bit too much. `$0, $1` should work, the only reason I wanted to mention dependency is because there can be a reference to `$resource` as well in the action string, so `$0` should not create any confusion on exactly what it is referring to. What do you think of the following manifest?

```
# A hypothetical case where we will need resource name in the action command
generated_resources:
-  pattern: artifacts/summary
    dependencies:
    - pattern: apis/-
    - pattern: apis/-/versions/-
    - pattern: apis/-/versions/-/specs/-
    action: "compute summary $resource"
-  pattern: apis/-/versions/-/specs/-/artifacts/score
    dependencies:
    - pattern: $resource.spec/artifacts/complexity
    - pattern: $resource.spec/artifacts/lintstats-gnostic
    action: "compute score $0 $1"
```

As for the list of $variable s, currently there are only two $dependency(0..N) and $resource.

shrutiparabgoogle:
Lost track of this PR somehow. Bumping it up for review/approval.

timburks (approved):

PR #227, opened 2021-07-13
Add SpecDiff Module

Jamaalbwells:
Add Diff Proto for Open API 3 spec diff
Add Functionality for Diffing and Parsing Tuffin OSA diff structs into diff proto format
Add Testing for Diff Parsing
Modify make to build new Diff_Analytics Proto


cmd/specdiff/differ_test.go
```diff
@@ -0,0 +1,282 @@
...
+					"components.schemas.Pet.properties.tag.type": {
+						To:   "integer",
+						From: "string",
+					},
+					"components.schemas.Pet.properties.tag.format": {
+						To: "int64",
+					},
+				},
+			},
+		},
```
seaneganx, line 73: Consider adding a case to test what happens when you rearrange the ordering of various parts of the spec.

I expect that fields are likely able to appear in any order without the diff changing, but what happens if we change the order of entries in a list of required fields? Requiring `id` and `name` is the same as requiring `name` and `id` so I don't think there should be anything in the diff about it.

cmd/specdiff/differ_test.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			yamlFile, err := ioutil.ReadFile(test.baseSpec)
+			if err != nil {
+				t.Fatalf("Failed to get base spec yaml: %s", err)
+			}
+			yamlFile2, err := ioutil.ReadFile(test.revisionSpec)
```
seaneganx, line 81: These can be `base` and `revision` or other similar descriptive names.

cmd/specdiff/differ_test.go
```diff
@@ -0,0 +1,282 @@
...
+			diffProto, err := GetDiff(yamlFile, yamlFile2)
+			if err != nil {
+				t.Fatalf("Failed to get diff.Diff: %s", err)
+			}
+			opts := cmp.Options{
+				protocmp.Transform(),
+				cmpopts.SortSlices(func(a, b string) bool { return a < b }),
+			}
+			if !cmp.Equal(test.wantProto, diffProto, opts) {
+				t.Errorf("Test %+v returned unexpected diff (-want +got):\n%s", test.desc, cmp.Diff(test.wantProto, diffProto, opts))
```
seaneganx, line 94: On line 76 the description is already included in the name of the subtest, so printing it again isn't necessary. It's a good idea to call out the specific function that had an unexpected return value, i.e. GetDiff returned unexpected diff...

cmd/specdiff/differ_test.go
```diff
@@ -0,0 +1,282 @@
...
+				changeType: "added",
+			},
+			wantProto: &rpc.Diff{
+				Added: []string{
+					"input1.result1",
+					"input2.result2",
+				},
+			},
+		}, {
+			desc: "Map Diff Recursive Key",
```
seaneganx, line 124: I think "Struct Key" is a better description of it. Recursion is an implementation detail.

cmd/specdiff/differ_test.go
```diff
@@ -0,0 +1,282 @@
...
+			testMap: reflect.ValueOf(map[diff.Endpoint]testStruct{
+				{Method: "Test-Method",
```
seaneganx, line 126: Formatting can be improved here by moving Method to the next line.

timburks (commented):
What's the intended top-level command and its syntax? These seem to be deferred to a future PR but at least seem to be influencing the directory name (my initial reaction is that "specdiff" is more specific than necessary - I think "diff" would be fine and that a smart "registry diff" would handle different types of arguments, including specs).

cmd/specdiff/differ.go
```diff
@@ -0,0 +1,283 @@
...
+	case reflect.Bool:
+		return true
+	case reflect.Int:
+		return true
+	default:
+		return false
+	}
+}
+
+func getAtomicType(node reflect.Value) string {
```
seaneganx, line 270: Until now I couldn't find the right words for these, but I think these types are "scalar types" and types like maps/slices/arrays are "composite" or "compound" types.

I think this function can have a more descriptive name. Something like `stringifyScalar(scalar reflect.Value) string`, `scalarString(scalar reflect.Value) string`, or similar. It should match the naming we use for stringifying the endpoint struct.

cmd/specdiff/differ.go
```diff
@@ -0,0 +1,283 @@
...
+	return nil
+}
+
+func handleEndpointStruct(ed diff.Endpoint) string {
+	Method := getAtomicType(reflect.ValueOf(ed.Method))
+	Path := getAtomicType(reflect.ValueOf(ed.Path))
+	return fmt.Sprintf("{method.%s path.%s}", Method, Path)
+}
+
+func handleValueDiffStruct(vd diff.ValueDiff, diffProto *rpc.Diff, changePath *change) {
```
seaneganx, line 245: I think the code might be clearer if this function is removed. It's only used in one place, and can be handled in one line instead:
```
diffProto.Modification[changePath.fieldPath.String()] = &rpc.Diff_ValueChange{
	From: getAtomicType(reflect.ValueOf(vd.From)),
	To:   getAtomicType(reflect.ValueOf(vd.To)),
}
```

cmd/specdiff/differ.go
```diff
@@ -0,0 +1,283 @@
...
+		panic("searchMapType called with invalid type")
+	}
+
+	for _, childNodeKey := range mapNode.MapKeys() {
+		childNode := mapNode.MapIndex(childNodeKey)
+		if childNode.IsZero() {
+			continue
+		}
+
+		if isAtomicType(childNodeKey) {
```
seaneganx, line 132: Since both of these checks are looking at the type of the key, it might make more sense to do all the checking here if it works out nicely.
```
	switch childNodeKey.Interface().(type) {
	case float64, string, bool, int:
		// Handle scalar type.
	case diff.Endpoint:
		// Handle endpoint.
	default:
		// Return error.
	}
```

cmd/specdiff/differ.go
```diff
@@ -0,0 +1,283 @@
...
+
+	err := searchNode(value, diffProto, changePath)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func handleEndpointStruct(ed diff.Endpoint) string {
```
seaneganx, line 239: I think this function is equivalent to `fmt.Sprintf("{method.%s path.%s}", ed.Method, ed.Path)`. Similar to getAtomicType, I think it can have a more descriptive name that indicates we're turning the struct into a string.

I'm hesitant about using dots in this result because it might be confusing as an intermediate value in a path. For example I think `something.{method.GET path./users/{id}}.something` is more difficult to read than `something.{GET /users/{id}}.something`.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,23 @@
...
+  repeated string added = 1;
+  // Change will be displayed in format foo.bar.x
+  repeated string deleted = 2;
+
+  message ValueChange {
+    string From = 1;
+    string To = 2;
+  }
+  // Change will be displayed in format foo.bar.x {to.x from.x}
+  map <string, ValueChange> modification= 3;
```
seaneganx, line 22: With the fields `added` and `deleted` I expect this to be `modified`. Alternatively `additions`, `deletions`, and `modifications` would work.

cmd/specdiff/differ_test.go
```diff
@@ -0,0 +1,282 @@
...
+	tests := []struct {
+		desc          string
+		testValueDiff reflect.Value
+		change        change
+		wantProto     *rpc.Diff
+	}{
+		{
+			desc: "Value Diff Test",
+			testValueDiff: reflect.ValueOf(diff.ValueDiff{From: 66,
+				To: true}),
```
seaneganx, line 248: Nit: Formatting this on multiple lines would be clearer.

cmd/specdiff/differ.go
```diff
@@ -0,0 +1,283 @@
...
+	if arrayNode.Kind() != reflect.Slice && arrayNode.Kind() != reflect.Array {
+		panic("searchArrayAndSliceType called with invalid type")
+	}
+
+	for i := 0; i < (arrayNode.Len()); i++ {
+		childNode := arrayNode.Index(i)
+		if childNode.IsZero() {
+			continue
+		}
+		if childNode.Kind() == reflect.String {
```
seaneganx, line 167: Similar to the type checking in searchMapType you might be able to type check this using a switch instead.

seaneganx:
> What's the intended top-level command and its syntax? These seem to be deferred to a future PR but at least seem to be influencing the directory name

This package seems like it might be used by multiple commands, so currently it might make the most sense under cmd/registry/ or cmd/registry/core, but we haven't established much of a pattern for this yet.

I imagine `registry compute diff` might be one command to output or save a diff, then something like `registry compute breaking-changes` or `registry compute stability` would use this package but not output the diff.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+package google.cloud.apigee.registry.applications.v1;
+// [START go_declaration]
+option go_package = "github.com/apigee/registry/rpc;rpc";
+// [END go_declaration]
+
+/* This proto contains the diff of a spec and its revision. Changes that are
+additions fall in the added field. Changes that are Deletions fall in the
+deleted field, and changes that modify values fall in the value_modify map
+field. */
```
seaneganx, line 10: This documentation refers to field names that no longer exist. That being said, I think the sentences that explain the purpose of each field should be included in the actual field documentation instead.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,23 @@
...
+  // Change will be displayed in format foo.bar.x
+  repeated string additions = 1;
+  // Change will be displayed in format foo.bar.x
+  repeated string deletions = 2;
+
+  message ValueChange {
+    string From = 1;
+    string To = 2;
+  }
+  // Change will be displayed in format foo.bar.x {to.x from.x}
```
seaneganx, line 21: I don't think we should document how this message "will be displayed" since the code that displays it may change without knowing to update this comment. We might decide to serialize ValueChange using `:` to separate its fields from its values instead of `.`, and shouldn't need to update this documentation. 

This comment should stick to explaining _what_ is included in the map and the intended format of the string key. The `ValueChange` message can have its own documentation explaining the meaning of the `to` and `from` fields.

google/cloud/apigee/registry/applications/v1/diff_analytics.proto
```diff
@@ -0,0 +1,23 @@
...
+deleted field, and changes that modify values fall in the value_modify map
+field. */
+message Diff {
+  // Change will be displayed in format foo.bar.x
+  repeated string additions = 1;
+  // Change will be displayed in format foo.bar.x
+  repeated string deletions = 2;
+
+  message ValueChange {
+    string From = 1;
```
seaneganx, line 18: Is there a reason for these to be capitalized? Other fields in our protos are generally lowercase.

seaneganx (commented):
A few comments about the proto documentation, but otherwise LGTM!

seaneganx (approved):
Deferring to @timburks for final approval.

timburks (approved):
Looks great, I'd just like to adjust the location of the proto (as @seaneganx suggested)

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package google.cloud.apigee.registry.applications.v1;
+package google.cloud.apigee.registry.v1alpha1;
```
timburks, line 2: Let's make this package name "google.cloud.apigee.registry.applications.v1alpha1"

timburks (changes requested):
See my suggested change to the package name

timburks (approved):

PR #229, opened 2021-07-15
update demo scripts to the current API

timburks:


seaneganx (approved):
Consider adding a defensive check for the existence of `jq` if we're going to use it in the demo scripts. Bash doesn't exactly fail gracefully when a program is missing, so a check at the start of the script might provide value by failing fast.

timburks:
Updated - thanks @seaneganx !

Issue #230, opened 2021-07-20
Use spf13/viper for configuration


seaneganx:
Currently the configuration for the registry tool, registry server, and other applications comes from a mix of [environment variables](https://github.com/apigee/registry/blob/eba674418c534d1d125d74012051c2b69c4bf8d2/cmd/registry-server/main.go#L48), [configuration files](https://github.com/apigee/registry/blob/eba674418c534d1d125d74012051c2b69c4bf8d2/cmd/registry-server/main.go#L35), and [command-line flags](https://github.com/apigee/registry/blob/eba674418c534d1d125d74012051c2b69c4bf8d2/cmd/registry/cmd/compute/lint.go#L74). In some cases we [allow configuration files to reference environment variables](https://github.com/apigee/registry/blob/eba674418c534d1d125d74012051c2b69c4bf8d2/cmd/registry-server/main.go#L92).

Switching to the [viper](https://github.com/spf13/viper) configuration library would give us the flexibility of offering several ways to set configuration values while ensuring consistent precedence in each case. It also offers additional functionality like live reloading.

PR #234, opened 2021-07-21
Add server testing with PostgreSQL

### LINKED ISSUE #208: Expand CI testing to include Postgres storage backends
CI tests currently verify the service using the SQLite storage backend. This is useful for verifying the “demo experience” which is an important tool for educating people about the API, but our expected deployments will primarily use Postgres. To verify that they work correctly with Postgres, we should update our CI tests to run a second set of tests in parallel using an API server that is configured to use a Postgres backend.

[This post](https://dev.to/techschoolguru/how-to-setup-github-actions-for-go-postgres-to-run-automated-tests-81o) describes a way to set up a Postgres database inside a GitHub Action. 

It seems likely that we’ll also want to do something similar to test API search features that are based on Postgres.

seaneganx:
Fixes #208

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -15,15 +15,89 @@
 package server
 
 import (
+	"flag"
 	"fmt"
+	"sync"
 	"testing"
+
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
 )
+
+const (
+	postgresDriver   = "postgres"
+	postgresTestUser = "registry_tester"
+	postgresDBConfig = "host=localhost port=5432 user=registry_tester dbname=registry_test sslmode=disable"
+)
+
+var (
+	sharedStorage sync.Mutex
+	usePostgres   = false
+)
+
+func init() {
+	flag.BoolVar(&usePostgres, "postgresql", false, "perform server tests using postgresql")
+}
 
 func defaultTestServer(t *testing.T) *RegistryServer {
 	t.Helper()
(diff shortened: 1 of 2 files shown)


server/server_test.go
```diff
@@ -15,15 +15,89 @@
...
+
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+)
+
+const (
+	postgresDriver   = "postgres"
+	postgresTestUser = "registry_tester"
+	postgresDBConfig = "host=localhost port=5432 user=registry_tester dbname=registry_test sslmode=disable"
```
timburks, line 31: I don't see a way to configure these credentials for testing against a postgres database that has a different test user and configuration; I'd like to add that later. It would be good to also explain how to run the tests against postgres in the main README, possibly in [this section](https://github.com/apigee/registry#optional-running-with-a-postgresql-backend).

timburks (approved):

server/server_test.go
seaneganx: Agreed. I had a few ideas on how to handle configurability.
1. After addressing #230 this would be like any other configurable value: settable by file, environment, or flag.
2. We can borrow the standard [environment variables](https://www.postgresql.org/docs/9.3/libpq-envars.html) from `psql`.
3. We can add environment variables for this specific testing situation.

How important do you think it is to document this in the README? I think we should wait and see how often developers have a reason to run these locally before investing in ease of setup.

PR #235, opened 2021-07-21
Breakingchange

Jamaalbwells:
Add Breaking Change Module to Registry
Add Changes Proto Message
Add Test for BreakingChange Module


cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if _, err := strconv.ParseBool(valueOne); err == nil {
+		if _, err := strconv.ParseBool(valueOne); err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+func isNonStringValue(value string) bool {
```
seaneganx, line 204: I think we've discussed this as a possible bug due to `rpc.Diff_ValueChange` only storing strings, regardless of the type used in the original spec file. How will this affect our ability to detect breaking changes?

We should document any caveats about this and write tests that target this detail. If the tests will fail due to the bug we can mark them as skipped, but they should still be checked into the suite.

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,237 @@
...
+// GetBreakingChanges compares each change in a diff Proto to the relavant change type detection Patterns.
+// Each change is then catgorized as breaking, nonbreaking, or unknown.
+func GetBreakingChanges(diff *rpc.Diff) *rpc.ClassifiedChanges {
+	changePatterns := getChangePatterns()
+	changesProto := compareChangesToPatterns(changePatterns, diff)
+	return changesProto
+}
+
+// getChangePatterns Intalizes the Breaking and NonBreaking change Patterns.
+func getChangePatterns() detectionTypeList {
```
seaneganx, line 35: This function seems like it should be a variable. Looking ahead, it might be best as several variables that aren't contained in a `detectionList` struct.

I see this function being called once, then the result being passed through a few other functions as parameters, then its fields being accessed once each. Instead we can define variables for `d.breakingChanges.additions`, `d.nonBreakingChanges.additions`, `d.breakingChanges.deletions`, `d.nonBreakingChanges.deletions`, `d.breakingChanges.modifications`, and `d.nonBreakingChanges.modifications`.

Something like the following.
```
var (
  unsafeAdds = []regexp.Regexp{ ... }
  unsafeDeletes = []regexp.Regexp{ ... }
  unsafeMods = []regexp.Regexp{ ... }
  safeAdds = []regexp.Regexp{ ... }
  safeDeletes = []regexp.Regexp{ ... }
  safeMods = []regexp.Regexp{ ... }
)
```

seaneganx (commented):
Some suggestions for refactoring and questions about other behaviors. I recommend looking at turning those pattern lists into package variables first, then approaching other changes to processing.

cmd/registry/breaking-change-detector/breaking_change_detector.go
Jamaalbwells: Done

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,237 @@
...
+		},
+		UnknownChanges: &rpc.Diff{
+			Additions:     []string{},
+			Deletions:     []string{},
+			Modifications: make(map[string]*rpc.Diff_ValueChange),
+		},
+	}
+	updateChangeStatusAdditions(allChanges, d, diff)
+	updateChangeStatusDeletions(allChanges, d, diff)
+	updateChangeStatusModifications(allChanges, d, diff)
```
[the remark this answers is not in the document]
Jamaalbwells: Done

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,237 @@
...
+	for _, addition := range additions {
+		if fitsAnyPattern(d.breakingChanges.additions, addition) {
+			allChanges.BreakingChanges.Additions = append(allChanges.BreakingChanges.Additions, addition)
+			continue
+		}
+		if fitsAnyPattern(d.nonBreakingChanges.additions, addition) {
+			allChanges.NonBreakingChanges.Additions = append(allChanges.NonBreakingChanges.Additions, addition)
+			continue
+		}
+		allChanges.UnknownChanges.Additions = append(allChanges.UnknownChanges.Additions, addition)
```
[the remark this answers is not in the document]
Jamaalbwells: Done

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	updateChangeStatusAdditions(allChanges, d, diff)
+	updateChangeStatusDeletions(allChanges, d, diff)
+	updateChangeStatusModifications(allChanges, d, diff)
+
+	return allChanges
+}
+
+func updateChangeStatusAdditions(allChanges *rpc.ClassifiedChanges, d detectionTypeList, diff *rpc.Diff) {
+	additions := diff.GetAdditions()
```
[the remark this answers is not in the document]
Jamaalbwells: Done

cmd/registry/breaking-change-detector/breaking_change_detector_test.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+func TestChanges(t *testing.T) {
+	tests := []struct {
+		desc      string
+		diffProto *rpc.Diff
+		wantProto *rpc.ClassifiedChanges
+	}{
+		{
+			desc: "basic Addition Breaking Test",
```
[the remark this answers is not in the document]
Jamaalbwells: Done

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,212 @@
...
+		UnknownChanges: &rpc.Diff{
+			Additions:     []string{},
+			Deletions:     []string{},
+			Modifications: make(map[string]*rpc.Diff_ValueChange),
+		},
+	}
+	allChanges.BreakingChanges = getBreakingChanges(diff)
+	allChanges.NonBreakingChanges = getNonBreakingChanges(diff)
+	allChanges.UnknownChanges = getUnknownChanges(diff)
+	return allChanges
```
seaneganx, line 110: Now this function can be reduced to the following. 

```
return &rpc.ClassifiedChanges{
	BreakingChanges: getBreakingChanges(diff),
	NonBreakingChanges: getNonBreakingChanges(diff),
	UnknownChanges: getUnknownChanges(diff),
}
```

Since the function is only used in `GetBreakingChanges` you can actually remove the function entirely and return it all directly.

```
func GetBreakingChanges(diff *rpc.Diff) *rpc.ClassifiedChanges {
	return &rpc.ClassifiedChanges{
		BreakingChanges: getBreakingChanges(diff),
		NonBreakingChanges: getNonBreakingChanges(diff),
		UnknownChanges: getUnknownChanges(diff),
	}
}
```

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,212 @@
...
+		{
+			detectionWeight:    1,
+			PositiveMatchRegex: regexp.MustCompile("(tags.)+(.)"),
+		},
+	}
+)
+
+// GetBreakingChanges compares each change in a diff Proto to the relavant change type detection Patterns.
+// Each change is then catgorized as breaking, nonbreaking, or unknown.
+func GetBreakingChanges(diff *rpc.Diff) *rpc.ClassifiedChanges {
```
seaneganx, line 84: Should this be `GetClassifiedChanges` now?

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func fitsAnyPattern(patterns []detectionPattern, change string) bool {
+	for _, pattern := range patterns {
+		if pattern.PositiveMatchRegex == nil {
+			if !pattern.NegativeMatchRegex.MatchString(change) {
+				return true
+			}
+			continue
+		}
```
seaneganx, line 197: This case doesn't seem to exist with the current set of patterns, since a positive pattern is always provided. Is it safe to assume the negative match regex is non-nil here, or what happens in that case?

How do these checks differ from the following code?
```
		if p := pattern.NegativeMatchRegex; p != nil && p.MatchString(change) {
			continue
		}
		if p := pattern.PositiveMatchRegex; p != nil && p.MatchString(change) {
			return true
		}
```

cmd/registry/breaking-change-detector/breaking_change_detector.go
```diff
@@ -0,0 +1,212 @@
+package breakingchangedetector
+
+import (
+	"regexp"
+
+	"github.com/apigee/registry/rpc"
+)
+
+type detectionPattern struct {
+	detectionWeight    int
```
seaneganx, line 10: Do you have plans for this field in a follow up PR? If not I think we should leave it out.

seaneganx (commented):
Looks good! Just a few small comments.

cmd/registry/breaking-change-detector/breaking_change_detector.go
Jamaalbwells: Done

cmd/registry/breaking-change-detector/breaking_change_detector.go
Jamaalbwells: Removed

cmd/registry/breaking-change-detector/breaking_change_detector.go
Jamaalbwells: Done

cmd/registry/breaking-change-detector/breaking_change_detector.go
Jamaalbwells: Done

seaneganx (approved):

Issue #239, opened 2021-07-27
Refactor WaitGroup implementation in tasks


shrutiparabgoogle:
In the current implementation of taskQueues, there is a global [WaitGroup()](https://github.com/shrutiparabgoogle/registry/blob/c91582f76714975e4c19a745af9350bcfd475323/cmd/registry/core/tasks.go#L31). While this works for all the other registry commands, issues were faced while implementing the `registry resolve` command.

Registry resolve creates a taskQueue and submit execution tasks to it. The execution tasks are nothing but `cmd.Execute()` calls to the. subcommands like `compute lint`, `compute complexity`, etc. While not all of these subcommand implement taskQueues, most of them do.

The issue faced here was as follows:
- WaitGroup (wg) is global
- `registry resolve` creates a bunch of workers in the taskQueue, adds entries to global wg. Then waits on these workers to finish execution.
- the workers execute the sub commands. `compute lint` for example. ComputeLint command creates another taskQueue and adds more entries in the global wg. Once ComputeLint is done executing all it's commands, it is still waiting on the wg because it contains entries from `resolve` command. Hence, the subcommand worker keeps waiting, never returns and `resolve` command also keeps waiting because the subcommand workers have not returned.

In order to avoid this deadlock, we need to refactor the WaitGroup's global implementation.

PR #240, opened 2021-07-27
Restructure resolve command

shrutiparabgoogle:
Implementation of resolve command now calls the subcommands programmatically and not through `os.exec` calls.

Two limitations or follow up tasks:
* refactor WaitGroup implementation https://github.com/apigee/registry/issues/239
* Handle logging in a more structured way.

Both of the above tasks will change implementations of all the registry commands and hence would be best to handle them separately. In this change, the resolve command is calling the sub-tasks programmatically and doing it in a serially. We can switch back to `taskQueues` once the `WaitGroup`  refactor is done.

### DIFF
--- a/cmd/registry/cmd/resolve/resolve.go
+++ b/cmd/registry/cmd/resolve/resolve.go
@@ -16,8 +16,8 @@ package resolve
 
 import (
 	"context"
-	"fmt"
 	"log"
+	"strings"
 
 	"github.com/apigee/registry/cmd/registry/controller"
 	"github.com/apigee/registry/cmd/registry/core"
@@ -64,19 +64,13 @@ func Command(ctx context.Context) *cobra.Command {
 
 			log.Printf("Generated %d actions. Starting Execution...", len(actions))
 
-			taskQueue := make(chan core.Task, 1024)
-			for i := 0; i < 64; i++ {
-				core.WaitGroup().Add(1)
-				go core.Worker(ctx, taskQueue)
-			}
-			defer core.WaitGroup().Wait()
-			defer close(taskQueue)
+			for _, a := range actions {
+				log.Printf("Executing action: %s", a)
+				rootCmd := cmd.Root()
+				rootCmd.SetArgs(strings.Fields(a))
 
-			// Submit tasks to taskQueue
-			for i, a := range actions {
-				taskQueue <- &controller.ExecCommandTask{
-					Action: a,
-					TaskID: fmt.Sprintf("task%d", i),
+				if err := rootCmd.Execute(); err != nil {
+					log.Printf("Error executing action: %s", err)
 				}
 			}
 		},
--- a/cmd/registry/cmd/resolve/resolve_test.go
+++ b/cmd/registry/cmd/resolve/resolve_test.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/apigee/registry/cmd/registry/cmd/compute"
 	"github.com/apigee/registry/cmd/registry/cmd/upload"
 	"github.com/apigee/registry/cmd/registry/core"
 	"github.com/apigee/registry/connection"
@@ -29,6 +30,7 @@ import (
 	"github.com/apigee/registry/server/names"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/spf13/cobra"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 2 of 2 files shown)


cmd/registry/cmd/resolve/resolve.go
```diff
@@ -64,19 +64,15 @@ func Command(ctx context.Context) *cobra.Command {
...
 			// Submit tasks to taskQueue
-			for i, a := range actions {
-				taskQueue <- &controller.ExecCommandTask{
-					Action: a,
-					TaskID: fmt.Sprintf("task%d", i),
+			for _, a := range actions {
+				log.Printf("Executing action: %s", a)
+				rootCmd := cmd.Root()
+				rootCmd.SetArgs(strings.Split(a, " "))
+				log.Printf("%s", rootCmd.CalledAs())
```
seaneganx, line 72: Does this ever print anything? Since the command hasn't been called yet, the [documentation](https://pkg.go.dev/github.com/spf13/cobra#Command.CalledAs) says it should be an empty string. Please remove this or add some explanation to the message if it's going to log empty lines.

cmd/registry/cmd/resolve/resolve.go
```diff
@@ -64,19 +64,15 @@ func Command(ctx context.Context) *cobra.Command {
...
+				err = rootCmd.Execute()
+				if err != nil {
+					log.Printf("Error executing action: %s, %s", a, err.Error())
```
seaneganx, line 75: No action needed, but FYI errors are automatically unwrapped when we use `%s`. This has the same behavior:
```
if err := rootCmd.Execute(); err != nil {
    log.Printf("Error executing action: %s, %s", a, err)
}
```

We also log the action right before this error check, so the action itself doesn't need to be included. As long as it doesn't make the error hard to understand it's also fine to keep it there.

cmd/registry/cmd/resolve/resolve_test.go
```diff
@@ -193,11 +195,19 @@ func TestResolve(t *testing.T) {
 		t.Fatalf("Failed to upload the manifest: %s", err)
 	}
 
-	// Call the controller update command
-	resolveCmd := Command(ctx)
-	args = []string{"projects/" + testProject + "/artifacts/test-manifest"}
-	resolveCmd.SetArgs(args)
-	if err = resolveCmd.Execute(); err != nil {
+	// Set up the root command for resolve actions
+	// This is required because the resolve command expects itself to have a parent
```
seaneganx, line 199: From what I see in this PR, I think the expectation is that the resolve command's root (not necessarily the parent) is able to execute the resolved actions. This might be worth explaining in more detail because simply having a parent isn't enough.

If I'm misunderstanding please correct me.

seaneganx (approved):
LGTM. I would like to see some small changes or a response to clarify about the `CalledAs()` log message and the resolve command's expectations for executing actions.

cmd/registry/cmd/resolve/resolve.go
shrutiparabgoogle: I added it for debugging my test. Missed removing it, there is no need to log `CalledAs()`. Thanks for pointing it out!

cmd/registry/cmd/resolve/resolve.go
shrutiparabgoogle: Done.

cmd/registry/cmd/resolve/resolve_test.go
shrutiparabgoogle: I updated the comment with some more explanation, didn't realize that the context was not enough. 
Let me know if the new comment makes sense.

cmd/registry/cmd/resolve/resolve.go
```diff
@@ -64,19 +64,14 @@ func Command(ctx context.Context) *cobra.Command {
...
 
-			taskQueue := make(chan core.Task, 1024)
-			for i := 0; i < 64; i++ {
-				core.WaitGroup().Add(1)
-				go core.Worker(ctx, taskQueue)
-			}
-			defer core.WaitGroup().Wait()
-			defer close(taskQueue)
-
 			// Submit tasks to taskQueue
```
timburks, line 67: Is this comment still relevant? It seems that this change is removing the task queue. Is this intentionally now serializing task execution? Is that temporary?

shrutiparabgoogle: Yeah the comment was not relevant. Serial execution is a temporary change. We currently can't use task queues because of this issue https://github.com/apigee/registry/issues/239. Once that is resolved, the plan is to switch back to task queue.

cmd/registry/cmd/resolve/resolve_test.go
seaneganx: Looks good.

cmd/registry/cmd/resolve/resolve.go
```diff
@@ -64,19 +64,13 @@ func Command(ctx context.Context) *cobra.Command {
...
-			for i := 0; i < 64; i++ {
-				core.WaitGroup().Add(1)
-				go core.Worker(ctx, taskQueue)
-			}
-			defer core.WaitGroup().Wait()
-			defer close(taskQueue)
+			for _, a := range actions {
+				log.Printf("Executing action: %s", a)
+				rootCmd := cmd.Root()
+				rootCmd.SetArgs(strings.Split(a, " "))
```
seaneganx, line 70: I just learned the `strings` package already has a [function](https://pkg.go.dev/strings#Fields) for this. Consider using `strings.Fields(a)` instead.

shrutiparabgoogle: Thanks, done.

cmd/registry/cmd/resolve/resolve.go
timburks: OK. SGTM!

PR #244, opened 2021-08-02
restore processing of --output and --output_id flags

timburks:
A few flags were missed during the structural refactoring - I think this is all of them,.


cmd/registry/cmd/vocabulary/unique.go
```diff
@@ -60,6 +61,6 @@ func uniqueCommand(ctx context.Context) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&outputID, "output_id", "vocabulary-unique", "Artifact ID to use when saving each result vocabulary")
+	cmd.Flags().String("output_id", "vocabulary-unique", "Artifact ID to use when saving each result vocabulary")
```
seaneganx, line 64: Was there an issue with this one? It's working for me.

cmd/registry/cmd/vocabulary/difference.go
```diff
@@ -35,7 +34,10 @@ func differenceCommand(ctx context.Context) *cobra.Command {
 			if err != nil {
 				log.Fatalf("Failed to get filter from flags: %s", err)
 			}
-
+			output, err := cmd.Flags().GetString("output")
```
seaneganx, line 37: The only change that should be needed to make this (and each of the other cases) work is to switch from `cmd.Flags().String(...)` to `cmd.Flags().StringVar(&output, ...)`. That's how all the other commands are handling non-persistent flags, since an error check isn't necessary in that scenario.

seaneganx (approved):

cmd/registry/cmd/vocabulary/difference.go
timburks: I'd like to consistently use one mechanism - in this case, I saw other uses of `cmd.Flags().String()` and changed these to match. The benefit that I see in using `String()` instead of `StringVar()` is that with `String()`, we don't have to look for the variable referenced by `StringVar()` to be sure it's in a closure and not a global... using `String()` adds an error check, but we already have plenty of those, so I don't think eliminating the GetString() error check will significantly improve readability (vs the improvement that I mentioned)

cmd/registry/cmd/vocabulary/unique.go
timburks: I accidentally changed this when I was checking all of the "output" related flags. I'll revert this one, but want to revisit the consistency question (below).

PR #245, opened 2021-08-02
Accept duplicate API creation during bulk uploads.

timburks:
Because entries for API specs are created in parallel, often an API resource is created more than once (due to lack of synchonization of creation threads). This is a benign situation, so we avoid printing errors and instead report that the API was "found".


seaneganx (approved):
Should we update version creation to use the same check?

timburks:
Let's come back to the "version" question - parallel version creation doesn't happen in our current uploaders because there's always exactly one spec uploaded for each version - but if that changes, we might start seeing similar errors for versions (I think like these, they would also be benign).

PR #247, opened 2021-08-03
Metrics

Jamaalbwells:
Add Metrics Module To Registry. 
- Add Stats and Metrics Protos 
- Add Functionality for Calculation Stats and Metrics Protos From Classified Diffs
- Add Tests For Metrics Module


cmd/registry/metrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+// ComputeMetrics will compute the metrics proto for a list of Classified Diffs.
+func ComputeMetrics(calculatedDiffs []*rpc.ClassifiedChanges) *rpc.Metrics {
+	stats := calculateStats(calculatedDiffs)
+	return caluclateMetrics(stats)
+}
+
+func calculateStats(calculatedDiffs []*rpc.ClassifiedChanges) *rpc.Stats {
+	totalBreakingChanges := 0
+	totalNonbreakingChanges := 0
+	for _, caluclatedDiff := range calculatedDiffs {
```
seaneganx, line 16: Nit: `caluclatedDiff` -> `calculatedDiff`.

As an aside, I think these names are fairly verbose and can be clear with smaller names. For example, `calculatedDiffs` -> `diffs`, `totalBreakingChanges` -> `breaking`, `totalNonbreakingChanges` -> `nonbreaking`, `calculatedDiff` -> `diff`.

cmd/registry/metrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+
+		totalBreakingChanges = (totalBreakingChanges +
+			len(caluclatedDiff.BreakingChanges.Additions) +
+			len(caluclatedDiff.BreakingChanges.Deletions) +
+			len(caluclatedDiff.BreakingChanges.Modifications))
```
seaneganx, line 21: Nit: Consider adding incrementally instead of computing and assigning the whole sum.

```
totalBreakingChanges += len(...Additions)
totalBreakingChanges += len(...Deletions)
totalBreakingChanges += len(...Modifications)
```

cmd/registry/metrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+
+// ComputeMetrics will compute the metrics proto for a list of Classified Diffs.
+func ComputeMetrics(calculatedDiffs []*rpc.ClassifiedChanges) *rpc.Metrics {
+	stats := calculateStats(calculatedDiffs)
+	return caluclateMetrics(stats)
+}
+
+func calculateStats(calculatedDiffs []*rpc.ClassifiedChanges) *rpc.Stats {
+	totalBreakingChanges := 0
+	totalNonbreakingChanges := 0
```
seaneganx, line 15: Nit: If you declare these as int64 you won't need to cast them later. It doesn't matter much in this case, but if these sums ever exceeded 32 bits then declaring them as int64 would be really important.

```
var (
	totalBreakingChanges int64 = 0
	totalNonbreakingChanges int64 = 0
)
```

cmd/registry/metrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+		totalNonbreakingChanges = (totalNonbreakingChanges +
+			len(caluclatedDiff.NonBreakingChanges.Additions) +
+			len(caluclatedDiff.NonBreakingChanges.Deletions) +
+			len(caluclatedDiff.NonBreakingChanges.Modifications))
+
+		// Default Unknown Changes to Nonbreaking.
+		totalNonbreakingChanges = (totalNonbreakingChanges +
+			len(caluclatedDiff.UnknownChanges.Additions) +
+			len(caluclatedDiff.UnknownChanges.Deletions) +
+			len(caluclatedDiff.UnknownChanges.Modifications))
```
seaneganx, line 32: Consider adding these to a separate counter, then combining the two counts when you create the stats message. That might make it clearer that we're treating those as the same in output, even though they're different as input.

go.sum
```diff
@@ -1454,15 +1454,11 @@ google.golang.org/genproto v0.0.0-20210719143636-1d5a45f8e492 h1:7yQQsvnwjfEahbN
 google.golang.org/genproto v0.0.0-20210719143636-1d5a45f8e492/go.mod h1:ob2IJxKrgPT52GcgX759i1sleT07tiKowYBGbczaW48=
 google.golang.org/genproto v0.0.0-20210721163202-f1cecdd8b78a h1:17YmRUuEF4d+t2ygJZaDPhqNL2Hf17832xWKcMU7r2I=
 google.golang.org/genproto v0.0.0-20210721163202-f1cecdd8b78a/go.mod h1:ob2IJxKrgPT52GcgX759i1sleT07tiKowYBGbczaW48=
-google.golang.org/genproto v0.0.0-20210729151513-df9385d47c1b h1:4xoALQmXxqVdDdLimpPyPeDdsJzo+nFTJw9euAMpqgM=
```
seaneganx, line 1457: Did you mean to include this file?

cmd/registry/metrics/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"github.com/apigee/registry/rpc"
+)
+
+// ComputeMetrics will compute the metrics proto for a list of Classified Diffs.
+func ComputeMetrics(calculatedDiffs []*rpc.ClassifiedChanges) *rpc.Metrics {
+	stats := calculateStats(calculatedDiffs)
+	return caluclateMetrics(stats)
+}
```
seaneganx, line 11: Consider splitting this into two functions, `ComputeStats(...*rpc.ClassifiedChanges) *rpc.Stats` and `ComputeMetrics(*rpc.Stats) *rpc.Metrics`. This would also simplify the API by removing the need to include `Stats` as a submessage of `Metrics`.

Note: The `...` in `ComputeStats` makes it a [variadic function](https://gobyexample.com/variadic-functions) that can be passed any number of individual args, or a slice using the spread operator. It's just a convenience for users of the package.

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
```diff
@@ -43,3 +43,15 @@ message ClassifiedChanges {
   Diff unknownChanges = 3;
 }
 
+message Stats {
+  int64 totalChanges = 1;
```
seaneganx, line 47: These field names should be in snake_case instead of camelCase. I didn't notice this on the ClassifiedChanges message, but that can be corrected in a different PR.

seaneganx (commented):
Please remove the `go.sum` file from the PR if `go.mod` isn't being updated as well.

The only important comment here is the one about splitting ComputeMetrics into two steps. Aside from that, just some nits.

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
```diff
@@ -43,3 +43,15 @@ message ClassifiedChanges {
   Diff unknownChanges = 3;
 }
 
+message Stats {
```
timburks, line 46: Since this is still in the "applications" package, let's call this something more specific... maybe "ChangeStats" (and "ChangeMetrics" below)?

How did you decide which of these is the statistic and which is the metric?

Is one derived from the other? Comments on the message and field declarations could make this clearer.

timburks: I might call "ClassifiedChanges" "ChangeDetails" to make "Change" a consistent prefix.

timburks: Also, is "ValueChange" scoped enough that it could just be "Change"?

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
```diff
@@ -43,3 +43,15 @@ message ClassifiedChanges {
...
+  int64 totalChanges = 1;
+  int64 totalBreakingChanges = 2;
+  int64 totalNonBreakingChanges = 3;
+  int64 numberOfDiffs = 4;
```
jelds, line 50: nit: "numberOfDiffs" is more verbose than I'm used to, in particular, I haven't seen "of" included in any APIs. Generally something like "num_diffs" would suffice, but if there's precedent for this in our external code, please feel free to ignore.

jelds (changes requested):

cmd/registry/metrics/metrics.go
Jamaalbwells: Done

cmd/registry/metrics/metrics.go
Jamaalbwells: Done

cmd/registry/metrics/metrics.go
Jamaalbwells: Done

cmd/registry/metrics/metrics.go
Jamaalbwells: Done

go.sum
Jamaalbwells: Yes, I reran Make All for this next commit to resolve any weird issues, such as this.

cmd/registry/metrics/metrics.go
Jamaalbwells: Done

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
Jamaalbwells: Value Change Refers to the value of a field being changed Ex: 
components.schemas.Pet.properties.tag.type change field: from:"integer" to:"double"
that "from: Integer to: Double" in that line is a value change. 
I will address the ChangeDetails Change in a new PR where I'm addressing Sean's request as well.

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
Jamaalbwells: Done

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
```diff
@@ -43,3 +43,27 @@ message ClassifiedChanges {
...
+  // total_non_breaking_changes repersents the total number of non-breaking changes.
+  int64 total_non_breaking_changes = 3;
+  // num_diffs repersents the number of diffs used in this stats
+  int64 num_diffs = 4;
+  }
+
+/* ChangeMetrics holds metrics about a list of diffs. Each metric is computed from
+two or more stats. */
+message ChangeMetrics {
+  ChangeStats stats = 1;
```
seaneganx, line 61: I don't think we need to attach the input message as part of the output message now that we're computing metrics from the stats.

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
seaneganx: [AIP-141](https://google.aip.dev/141) actually addresses this. Maybe this should be `change_count`, `breaking_change_count`, `nonbreaking_change_count`, and `diff_count`?

google/cloud/apigee/registry/applications/v1alpha1/diff_analytics.proto
Jamaalbwells: Done

seaneganx (approved):

PR #249, opened 2021-08-04
Add tests to validate command coverage in controller/local-tasks.go

shrutiparabgoogle:
Add tests to secure the [switch case block](https://github.com/apigee/registry/blob/68e0507f798395bab574ed4b5a90db4e0cdc8cb0/cmd/registry/controller/local-tasks.go#L41) in `controller/local-tasks.go`

### DIFF
--- a/cmd/registry/controller/local-tasks.go
+++ b/cmd/registry/controller/local-tasks.go
@@ -40,7 +40,7 @@ type ExecCommandTask struct {
 
 func getCommand(ctx context.Context, action []string) (*cobra.Command, error) {
 	if len(action) == 0 {
-		return nil, fmt.Errorf("Empty action string")
+		return nil, fmt.Errorf("empty action string")
 	}
 
 	switch action[0] {
@@ -66,8 +66,10 @@ func getCommand(ctx context.Context, action []string) (*cobra.Command, error) {
 		return upload.Command(ctx), nil
 	case "vocabulary":
 		return vocabulary.Command(ctx), nil
+	case "resolve":
+		return nil, fmt.Errorf("cannot use command 'resolve' in action")
 	default:
-		return nil, fmt.Errorf("Action '%s' is not supported.", strings.Join(action, " "))
+		return nil, fmt.Errorf("action '%s' is not supported", strings.Join(action, " "))
 	}
 }
 
@@ -86,7 +88,7 @@ func (task *ExecCommandTask) Run(ctx context.Context) error {
 		cmd.SetArgs(args[1:])
 	}
 	if err := cmd.Execute(); err != nil {
-		return fmt.Errorf("Error executing action %s: %s", task.Action, err)
+		return fmt.Errorf("error executing action %s: %s", task.Action, err)
 	}
 
 	log.Printf("Finished executing taskId %s with action: %s", task.TaskID, task.Action)
--- a/cmd/capabilities/testdata/openapi.yaml
+++ b/cmd/capabilities/testdata/openapi.yaml
@@ -1,3 +1,17 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 openapi: "3.0.0"
 info:
   version: 1.0.0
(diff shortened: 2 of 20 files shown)


cmd/registry/tests/task-validation_test.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			ctx := context.Background()
+			_, err := controller.GetCommand(ctx, strings.Fields(test.action))
+			if err == nil {
+				t.Errorf("Expected GetCommand() to return error.")
+			}
+			if err.Error() != test.want {
+				t.Errorf("Generated unexpected error message. Want: %s, got: %s", test.want, err)
+			}
```
seaneganx, line 128: I don't think we should be testing the error message specifically because changing the error message isn't something we want to prevent. Here are two suggestions:

1. Remove this check so we only fail if `err == nil`. I think this is good enough, but it's possible a test case could incorrectly pass when a different error than we expect occurs. 
2. Export specific errors as variables so we can check for specific errors without relying on their message. For example, see the [os package](https://pkg.go.dev/os#pkg-variables). Packages usually only do this if the callers will handle the errors differently, e.g. ignoring not found errors and returning others.

cmd/registry/tests/task-validation_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"context"
+	"fmt"
+	"github.com/apigee/registry/cmd/registry/cmd"
+	"github.com/apigee/registry/cmd/registry/controller"
+	"strings"
+	"testing"
+)
+
+// Test that GetCommand() is returning the correct instance for each command string
+func TestValidateCommands(t *testing.T) {
```
seaneganx, line 13: Nit: Conventionally, this should be `TestGetCommand` if it's going to test GetCommand directly.

cmd/registry/tests/task-validation_test.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			if err.Error() != test.want {
+				t.Errorf("Generated unexpected error message. Want: %s, got: %s", test.want, err)
+			}
+		})
+	}
+}
+
+// Test that currently supported commands in the registry tool are all covered in GetCommand()
+func TestCommandCoverage(t *testing.T) {
```
seaneganx, line 134: Does this test do anything different than the other two groups of tests? This seems to be the only reason why these tests aren't in the same package as the GetCommand function being tested. Please correct me if I'm wrong.

I think the tests that manually cover each case are better because of their simplicity and because they don't depend on things like the root command that the package being tested doesn't depend on.

shrutiparabgoogle: This is to make sure that all the commands supported through rootCmd should be included in the switch case block. 

If rootCommand supports a new command and is not included in the switch block, the other test `TestValidateCommands` can't catch this case. Because the switch case block is not dynamic and hardcoded, we have to make sure that any addition of a new command gets included in the `resolve` command, and this test is a way for enforcing that.

seaneganx: Okay, that being said I think this is a test of the `Command` function in `cmd/registry/cmd` and should be in a test new file at `cmd/registry/cmd/root_test.go`. Similar to the other tests, I think it would be better if it made sure each subcommand can successfully run in a `ExecCommandTask` rather than making sure `GetCommand` returns without error.

cmd/registry/tests/task-validation_test.go
```diff
@@ -0,0 +1,147 @@
+package main
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done. Running addlicense added headers to a bunch of other files.

cmd/registry/controller/local-tasks.go
```diff
@@ -38,9 +38,9 @@ type ExecCommandTask struct {
 	TaskID string
 }
 
-func getCommand(ctx context.Context, action []string) (*cobra.Command, error) {
+func GetCommand(ctx context.Context, action []string) (*cobra.Command, error) {
 	if len(action) == 0 {
-		return nil, fmt.Errorf("Empty action string")
+		return nil, fmt.Errorf("Empty action string.")
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

I have installed gofmt plugin in my sublime editor. It usually auto-formats code. I will check the setting if I am missing some.

cmd/registry/controller/local-tasks.go
```diff
@@ -38,9 +38,9 @@ type ExecCommandTask struct {
 	TaskID string
 }
 
-func getCommand(ctx context.Context, action []string) (*cobra.Command, error) {
+func GetCommand(ctx context.Context, action []string) (*cobra.Command, error) {
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

cmd/registry/tests/task-validation_test.go
shrutiparabgoogle: Makes sense. I am not exporting the errors because the client is not treating any of the messages differently. The error message is just for debug purposes.

cmd/registry/tests/task-validation_test.go
shrutiparabgoogle: Done

cmd/registry/tests/task-validation_test.go
shrutiparabgoogle: Done

seaneganx (approved):

PR #253, opened 2021-08-06
Switch resolve command back to multi-process

shrutiparabgoogle:
Move ExecCommandTask back to using exec.Command()

### DIFF
--- a/cmd/registry/controller/local-tasks.go
+++ b/cmd/registry/controller/local-tasks.go
@@ -17,19 +17,8 @@ package controller
 import (
 	"context"
 	"fmt"
-	"github.com/apigee/registry/cmd/registry/cmd/annotate"
-	"github.com/apigee/registry/cmd/registry/cmd/compute"
-	"github.com/apigee/registry/cmd/registry/cmd/delete"
-	"github.com/apigee/registry/cmd/registry/cmd/export"
-	"github.com/apigee/registry/cmd/registry/cmd/get"
-	"github.com/apigee/registry/cmd/registry/cmd/index"
-	"github.com/apigee/registry/cmd/registry/cmd/label"
-	"github.com/apigee/registry/cmd/registry/cmd/list"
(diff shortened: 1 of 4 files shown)


cmd/registry/controller/local-tasks.go
```diff
@@ -38,59 +27,21 @@ type ExecCommandTask struct {
...
 		return err
 	}
-	// Set args if any
-	if len(args) > 1 {
-		cmd.SetArgs(args[1:])
-	}
-	if err := cmd.Execute(); err != nil {
-		return fmt.Errorf("error executing action %s: %s", task.Action, err)
-	}
+	log.Printf("Finished executing taskId %s with action: %s\n Output: %s", task.TaskID, task.Action, output)
```
seaneganx, line 44: Here we can use `cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr` to hook the process directly into the correct outputs. The logs for each action will be interleaved without any way to identify them separately right now though.

No action necessary if you prefer the current output until we have unique identifiers attached to command logs.

seaneganx (commented):
The workaround in `cmd/registry/cmd/resolve/resolve_test.go` can also be rolled back to test `Command(ctx)` directly instead of nesting it under a fake root command.

seaneganx (approved):
Approving since I'm on vacation for the first few days of next week. Please address the changes I suggested to `resolve_test.go` before submitting.

cmd/registry/controller/local-tasks.go
shrutiparabgoogle: Noted. I am planning to take care of the logging change in a separate PR.

shrutiparabgoogle:
> The workaround in `cmd/registry/cmd/resolve/resolve_test.go` can also be rolled back to test `Command(ctx)` directly instead of nesting it under a fake root command.

Done.

PR #256, opened 2021-08-11
Compute hashes and sizes of ungzipped contents of specs and artifacts

timburks:


server/models/blob.go
```diff
@@ -61,8 +61,8 @@ func NewBlobForArtifact(artifact *Artifact, contents []byte) *Blob {
 		VersionID:   artifact.VersionID,
 		SpecID:      artifact.SpecID,
 		ArtifactID:  artifact.ArtifactID,
-		Hash:        hashForBytes(contents),
-		SizeInBytes: int32(len(contents)),
+		Hash:        artifact.Hash,
+		SizeInBytes: artifact.SizeInBytes,
```
seaneganx, line 65: This looks pretty unsafe to me because it relies on the artifact model having these fields computed correctly using the contents that are being provided here. Why are we storing these fields in the Blob _and_ Artifact models anyway?

If Blobs are meant to store contents without knowing the MIME type, then Blobs should not contain the `Hash` and `SizeInBytes` fields that depend on the MIME type. If we are going to store these fields, then we should store the MIME type with it, and these fields should be computed based on the contents and the MIME type.

seaneganx (changes requested):
Unless I'm misunderstanding the code, it looks like updating specs with gzipped contents will cause incorrect size and hash.

Also, I think the API documentation is missing some relevant information about compression. GetApiSpecContents and GetArtifactContents don't mention that they decompress contents for some MIME types, and the `mime_type` field doesn't mention that the `+gzip` suffix has special meaning.

server/models/artifact.go
```diff
@@ -53,11 +54,19 @@ func NewArtifact(name names.Artifact, body *rpc.Artifact) *Artifact {
 	}
 
 	if body.GetContents() != nil {
-		artifact.SizeInBytes = int32(len(body.GetContents()))
-		artifact.Hash = hashForBytes(body.GetContents())
+		contents := body.GetContents()
+		// if contents are gzipped, uncompress before computing size and hash.
+		if strings.Contains(artifact.MimeType, "+gzip") && len(contents) > 0 {
```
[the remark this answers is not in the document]
timburks: This code is for artifacts, but I carried the length check forward from specs in order for [this test](https://github.com/apigee/registry/blob/4b21d12ae943fbacc8e435b999a0cb3eaa4512c9/server/actions_specs_test.go#L454) to work - its seed is a gzipped spec with zero length. Our tests don't notice if I remove the length check for artifacts but the above test breaks if I remove it for specs.

server/models/blob.go
timburks: This seems like a good topic for a refactor -- worth considering, but I'd like for this PR to add the capability and pass the tests (and refactor separately)

server/actions_artifacts.go
```diff
@@ -96,7 +96,10 @@ func (s *RegistryServer) CreateArtifact(ctx context.Context, req *rpc.CreateArti
 		}
 	}
 
-	artifact := models.NewArtifact(name, req.GetArtifact())
+	artifact, err := models.NewArtifact(name, req.GetArtifact())
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
timburks: This is a good pattern - since most of the other calls in this file are returning already-wrapped errors, I put the error-wrapping inside models.NewArtifact (where the gzipping might fail). I think we should do the same for models.NewSpec

server/internal/storage/models/gzip.go
```diff
@@ -0,0 +1,31 @@
...
+package models
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io/ioutil"
+)
+
+// GUnzippedBytes uncompresses a slice of bytes.
+func GUnzippedBytes(input []byte) ([]byte, error) {
```
seaneganx, line 24: Why is this exported? It seems like it's only used in the models package.

seaneganx (approved):
LGTM with a few outstanding comments.

server/internal/storage/models/gzip.go
timburks: Good point - I'll make it private

timburks: Double-checking, I found this is used in actions_specs.go (by a change in this PR) so I'll leave this as-is for now.

PR #259, opened 2021-08-13
Include completion information in controller's tasks

shrutiparabgoogle:
This PR includes the following changes:
1) Setting the output of child commands to StdOut and StdErr so that the logs generated by the child commands are all visible.
2) `local-tasks.go`: Each individual go routine will log the completion of a command execution by logging the status, action and taskID (will be tied to child command's uid in the future). This information should be enough for now to generate basic summary dashboards like: total num actions, num failures, % success, etc

### DIFF
--- a/cmd/registry/controller/local-tasks.go
+++ b/cmd/registry/controller/local-tasks.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"os/exec"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


cmd/registry/controller/local-tasks.go
```diff
@@ -36,12 +37,14 @@ func (task *ExecCommandTask) Run(ctx context.Context) error {
 		return fmt.Errorf("'resolve' not allowed in action, cannot invoke %q", task.Action)
 	}
 	cmd := exec.Command("registry", strings.Fields(task.Action)...)
-	output, err := cmd.CombinedOutput()
+	cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr
+	err := cmd.Run()
+	details := fmt.Sprintf("Finished execution %%s action={%s} taskID={%s}", task.Action, task.TaskID)
```
shrutiparabgoogle, line 42: taskID can later be passed as a flag to the child process as it's uid.

shrutiparabgoogle:
Planning to build a basic dashboard based on the uid prefixes and this information. Will then look for enhancements, once we have a connection between the parent generated taskID and the child's uid.

cmd/registry/controller/local-tasks.go
```diff
@@ -36,12 +37,14 @@ func (task *ExecCommandTask) Run(ctx context.Context) error {
...
 
 	if err != nil {
+		log.Printf(details, "FAILED")
 		return err
 	}
-	log.Printf("Finished executing taskId %s with action: %s\n Output: %s", task.TaskID, task.Action, output)
-
+	log.Printf(details, "SUCCESS")
```
seaneganx, line 48: Using the escaped formatter from the details variable is a bit confusing here. Could we capture less information in the details variable? These log messages can be `log.Printf("Failed execution of %s: %s", details, err)` and `log.Printf("Successful execution of %s", details)` if `details` only include the action and task ID.

Alternatively I actually think it's clearer to simply have two fully separate log messages that just repeat the common details, but there's no right answer for that.

seaneganx (approved):

cmd/registry/controller/local-tasks.go
shrutiparabgoogle: Right, thanks for calling it out. I ended up over optimizing the code. Made changes, hope it's more clear now

PR #262, opened 2021-08-19
Populate revision_tags field for specs

seaneganx:
Fixes #262 but doesn't include tests yet.


server/actions_specs.go
```diff
@@ -314,19 +331,55 @@ func (s *RegistryServer) UpdateApiSpec(ctx context.Context, req *rpc.UpdateApiSp
 	}
 
 	// If the spec contents were updated, save a new blob.
-	implicitUpdate := req.GetUpdateMask() == nil && len(req.ApiSpec.GetContents()) > 0
-	explicitUpdate := len(fieldmaskpb.Intersect(req.GetUpdateMask(), &fieldmaskpb.FieldMask{Paths: []string{"contents"}}).GetPaths()) > 0
-	if implicitUpdate || explicitUpdate {
+	if len(fieldmaskpb.Intersect(maskExpansion, &fieldmaskpb.FieldMask{Paths: []string{"contents"}}).GetPaths()) > 0 {
```
seaneganx, line 334: This is unrelated to the current PR. I noticed this is an easier way to check if contents are being updated now that masks like `nil` and `*` are expanded into a list of relevant fields.

timburks (approved):

Issue #257, opened 2021-08-13
Registry API: DeleteApiSpecRevision can delete the only remaining revision of a spec


seaneganx:
[AIP-162](https://google.aip.dev/162#deleting-revisions) states the following. We don't currently support this expected behavior.

> The method must not cause the resource to be deleted. Because a resource should not exist with zero revisions, the method must fail with FAILED_PRECONDITION if the user attempts to delete the only revision.

We have a [test](https://github.com/apigee/registry/blob/9c0e693d871ddb18e31137a59949bbcebf3fb1f5/server/actions_spec_revisions_test.go#L237) for this case that is currently skipped.

Issue #250, opened 2021-08-04
TagApiSpecRevision - mixed case tags


Krpate:
Request:
```
curl --request POST 'http://x.x.x.x/v1/projects/x/apis/my-api/versions/v1/specs/proto@2d63042a:tagRevision' \
--header 'Content-Type: application/json' \
--data-raw '{
    "tag": "tagME"
}'
```

Response:
```
{
    "name": "projects/x/apis/my-api/versions/v1/specs/proto@tagME",
    "filename": "",
    "description": "",
    "revisionId": "2d63042a",
    "mimeType": "",
    "sizeBytes": 0,
    "hash": "",
    "sourceUri": "",
    "contents": "",
    "revisionTags": [],
    "labels": {},
    "annotations": {},
    "createTime": "2021-08-03T23:58:38.966919Z",
    "revisionCreateTime": "2021-08-03T23:58:38.966919Z",
    "revisionUpdateTime": "2021-08-03T23:58:38.966919Z"
}
```

From the response it seemed that the revision was successfully tagged but when I tried to retrieve the spec with tag, it failed with `400 Bad Request`. 
```
cur --request GET 'http://x.x.x.x/v1/projects/x/apis/my-api/versions/v1/specs/proto@tagME'
```

The above case works as expected if all the letters in `tagme` are lower cased. Seems like we don't allow mixed case for tag names. Can we reflect the same (throw an error) when calling `TagApiSpecRevision`?

timburks:
[This expression](https://github.com/apigee/registry/blob/fa370d3032231142375fc67d124f06dfa610309d/server/registry/names/common.go#L32) is used to validate revision tags in resource names. It seems reasonable to use it to validate tags passed to `TagApiSpecRevision`

Issue #198, opened 2021-06-21
Registry v1 API: Add resource representing "deployments" 


timburks:
We would like to have a way to directly represent deployments of an API.

Proposed: a new `deployments` collection under `apis` that is implemented with a repeated `ApiDeployment` message. 
- `ApiDeployment` would have a display name that would contain values like "Prod", "Staging", etc. but (as with other display names), we would not enforce uniqueness of this display name.  
- `ApiDeployment` resource IDs would have to be unique (as with all other resources).
- `ApiDeployment` would include a "spec" field that points to a spec (including revision) for the API being served by the deployment. 
- `ApiDeployment` would support labels, annotations, and artifacts.
- `ApiDeployment` would have revisions.

timburks:
Addressed in #378 and in the [v0.3.9](https://github.com/apigee/registry/releases/tag/v0.3.9) tag

Issue #190, opened 2021-06-16
Create library for declarative server state initialization


seaneganx:
As more tests are written, it's becoming clear that much of the code written for testing is to set up the initial server state. This makes the behavior under test more difficult to determine. Having to make manual requests to a client for setup can make it more challenging to implement common test patterns like table-driven testing. Each set of tests being written seems to have slightly different methods of setting up the initial server state, with some using helper methods with fixed test data, and others using in-line requests with case-specific test data.

Having a common method of seeding the initial server state may be beneficial. Our tests can particularly benefit from a library that makes it easy to construct a desired server state in a declarative manner.

timburks:
Could "initializing server state" also be expressed as "constructing a project"? 

Let's consider using YAML as an external representation. See the output of "registry export yaml" for one possible YAML format - if we settle on something we can write an importer for it. Spec contents could be either included inline (base64, ugly, possibly very large) or as references to side files. We might also consider mapping the registry contents to a directory structure that matches resource names (this might require/assume-that resource lists are always ordered alphabetically by resource id).

seaneganx:
I think it's more than just constructing a project. Most tests create other resources like APIs, versions, specs, and artifacts.

If we want to use an external representation I think YAML would be reasonable. Being able to reference other files for spec contents or creating a directory structure that can represent the resource hierarchy seems like it would trend into the territory of DSLs with the associated pros/cons. Particularly I think it would increase the difficulty of writing tests because developers would have to learn how the storage structure works and put in the effort to set it up for their needs.

What I had in mind was something closer to what is used in the [server tests](https://github.com/apigee/registry/blob/21587ed876e8bd4073334eb8c24eba0fa2a9a871/server/actions_specs_test.go#L475). Each test case can declaratively set the resource hierarchy by providing the resources they depend on. In those tests the helpers actually create parent resources that are implicitly required by the seed. You can see in those tests how many different states we might want the server to have across a suite.

Test isolation is fairly important for this to be useful, so that problem might be best to tackle first. If every test in the repo has to avoid collisions with others it becomes a burden.

timburks:
To clarify my comment: "project" meant a full project with all of its child resources.

PR #263, opened 2021-08-20
Add declarative seeding library to server

### LINKED ISSUE #112: Unit tests do not cover some revisioning behaviors
On PR #111 the demo tests caught a bug that should reasonably be caught by unit tests. Specifically ListSpecs was returning all revisions of all specs, but the unit test suite passed. Given the difficulty of seeding multiple revisions, there are likely many other cases that are not tested.

The unit test suite could benefit from having easier methods of seeding multiple spec revisions for a test case. Currently tests have to manually create and update a spec with new contents to create multiple revisions, which is cumbersome and discourages developers from thoroughly testing behaviors with multiple revisions. In the case of ListSpecs, the table-driven tests are not capable of seeding multiple revisions right now, so a separate test would need to be developed without better seeding helpers.

seaneganx:
* Add parent resource name accessor to name types.
* Add declarative seeding library.
* Update server tests to use seeding library.
* Add test for listing specs with multiple revisions.

Fixes #112, #190. With some minor changes to the server we can more easily test creating resource revisions and tagging revisions, which is helpful for #198, #250, and #257.


timburks (approved):
Thanks - if you're available, please merge this today.

PR #264, opened 2021-08-21
Refactor server storage layer

### LINKED ISSUE #147: Remove storage layer GORM abstractions

seaneganx:
This change refactors the storage layer of the registry server. Some primary achievements:
- RPC handlers in the server package interact only with the new storage package. The gorm client is abstracted away.
- The front end of the storage layer (formerly known as "dao") no longer interacts with the gorm package through limited interfaces.
- All storage-specific packages are now internal-only and located in one parent directory.

Open questions:
- Should the RPC handlers be responsible for converting RPC messages into storage models?
- Does the gorm package need that package-scoped mutex?

Fixes #147.  Fixes missed cases for #78.


server/internal/storage/gorm/client.go
```diff
@@ -163,36 +137,36 @@ func (c *Client) IsNotFound(err error) bool {
...
+	lock()
+	defer unlock()
+	return c.db.Where("key = ?", k.Name).First(v).Error
 }
 
 // Put puts an entity using the storage client.
-func (c *Client) Put(ctx context.Context, k storage.Key, v interface{}) (storage.Key, error) {
-	mylock()
-	defer myunlock()
+func (c *Client) Put(ctx context.Context, k *Key, v interface{}) (*Key, error) {
```
timburks, line 147: Not for this PR: I expect that we'll eventually flatten out this artifact of using Cloud Datastore

server/internal/storage/gorm/client.go
```diff
@@ -37,124 +36,99 @@ type Client struct {
 var mutex sync.Mutex
 var disableMutex bool
 
-func mylock() {
+func lock() {
```
timburks, line 39: I added this lock when I noticed problems due to concurrent access to the sqlite backend. I don't think our integration tests check this (yet), but when the demo scripts run bulk uploads (and other concurrent tasks), the server gets parallel writes - so this lock serializes them. As I understand it (and discussed elsewhere in existing comments), Postgres doesn't need this, so we only enable locks for SQLite backends.

... so I wouldn't want to remove this lock without tests showing that concurrent writes to SQLite are safe without it (and we should probably also validate the assumption that concurrent writes are fine with Postgres).

timburks:
This looks great for readability and non-disruptive to functionality. I didn't get what you meant by this question "Should the RPC handlers be responsible for converting RPC messages into storage models?" but think it's probably not critical - let's discuss it at a convenient time.

timburks (approved):

PR #265, opened 2021-08-21
Add "/locations/global" to the names of all resources below projects.

timburks:
This set of changes adds a location identifier to all resource paths below projects. While this doesn't provide direct value to the registry API, it makes it compatible with Google Cloud guidance to explicitly locate all resources under projects. That allows resources to be explicitly stored in regions of the user's choice (including countries). We expect that a registry is a global resource for an organization that will be stored in a single location. In this initial set of changes, we set locations to the "global" identifier, but in the future, this may be made configurable to allow a specific cloud region to be set in the registry instance. A Google-hosted version of the registry server would configure the server to use the desired location identifier (TODO if needed).

With this initial set of changes, all CI tests should pass. Subsequent commits will address issues found when running scripts in the demos directory.

### DIFF
--- a/server/names/names_test.go
+++ b/server/names/names_test.go
@@ -55,8 +55,8 @@ func TestResourceNames(t *testing.T) {
 			name:   "apis",
 			regexp: ApisRegexp(),
 			pass: []string{
-				"projects/googleLOCATION/apis",
-				"projects/-LOCATION/apis",
+				"projects/google/locations/LOCATION/apis",
+				"projects/-/locations/LOCATION/apis",
 			},
 			fail: []string{
 				"-",
@@ -66,26 +66,26 @@ func TestResourceNames(t *testing.T) {
 			name:   "api",
 			regexp: ApiRegexp(),
 			pass: []string{
-				"projects/googleLOCATION/apis/sample",
-				"projects/-LOCATION/apis/-",
-				"projects/123LOCATION/apis/abc",
-				"projects/1-2-3LOCATION/apis/abc",
+				"projects/google/locations/LOCATION/apis/sample",
+				"projects/-/locations/LOCATION/apis/-",
+				"projects/123/locations/LOCATION/apis/abc",
+				"projects/1-2-3/locations/LOCATION/apis/abc",
 			},
 			fail: []string{
 				"-",
 				"invalid",
-				"projects/LOCATION/apis/123",
-				"projects/123LOCATION/apis/",
-				"projects/123LOCATION/invalid/123",
-				"projects/123LOCATION/apis/ 123",
+				"projects//locations/LOCATION/apis/123",
+				"projects/123/locations/LOCATION/apis/",
+				"projects/123/locations/LOCATION/invalid/123",
+				"projects/123/locations/LOCATION/apis/ 123",
 			},
 		},
 		{
 			name:   "versions",
 			regexp: VersionsRegexp(),
 			pass: []string{
-				"projects/googleLOCATION/apis/sample/versions",
-				"projects/-LOCATION/apis/-/versions",
+				"projects/google/locations/LOCATION/apis/sample/versions",
+				"projects/-/locations/LOCATION/apis/-/versions",
 			},
 			fail: []string{
 				"-",
@@ -95,26 +95,26 @@ func TestResourceNames(t *testing.T) {
 			name:   "version",
 			regexp: VersionRegexp(),
 			pass: []string{
-				"projects/googleLOCATION/apis/sample/versions/v1",
-				"projects/-LOCATION/apis/-/versions/-",
-				"projects/123LOCATION/apis/abc/versions/123",
-				"projects/1-2-3LOCATION/apis/abc/versions/123",
+				"projects/google/locations/LOCATION/apis/sample/versions/v1",
+				"projects/-/locations/LOCATION/apis/-/versions/-",
+				"projects/123/locations/LOCATION/apis/abc/versions/123",
+				"projects/1-2-3/locations/LOCATION/apis/abc/versions/123",
 			},
 			fail: []string{
 				"-",
 				"invalid",
-				"projects/LOCATION/apis/123",
-				"projects/123LOCATION/apis/",
-				"projects/123LOCATION/invalid/123",
-				"projects/123LOCATION/apis/ 123",
+				"projects//locations/LOCATION/apis/123",
+				"projects/123/locations/LOCATION/apis/",
+				"projects/123/locations/LOCATION/invalid/123",
+				"projects/123/locations/LOCATION/apis/ 123",
 			},
 		},
 		{
 			name:   "specs",
 			regexp: SpecsRegexp(),
 			pass: []string{
-				"projects/googleLOCATION/apis/sample/versions/v1/specs",
-				"projects/-LOCATION/apis/-/versions/-/specs",
+				"projects/google/locations/LOCATION/apis/sample/versions/v1/specs",
+				"projects/-/locations/LOCATION/apis/-/versions/-/specs",
 			},
 			fail: []string{
 				"-",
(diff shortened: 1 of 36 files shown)


cmd/registry/core/get.go
```diff
@@ -109,13 +109,13 @@ func GetArtifact(ctx context.Context,
...
+		request.Name = "projects/" + segments[1] + "/locations/global"
 	} else if segments[5] == "" {
-		request.Name = "projects/" + segments[1] + "/apis/" + segments[3]
+		request.Name = "projects/" + segments[1] + "/locations/global/apis/" + segments[3]
 	} else if segments[7] == "" {
-		request.Name = "projects/" + segments[1] + "/apis/" + segments[3] + "/versions/" + segments[5]
+		request.Name = "projects/" + segments[1] + "/locations/global/apis/" + segments[3] + "/versions/" + segments[5]
 	} else {
-		request.Name = "projects/" + segments[1] + "/apis/" + segments[3] + "/versions/" + segments[5] + "/specs/" + segments[7]
+		request.Name = "projects/" + segments[1] + "/locations/global/apis/" + segments[3] + "/versions/" + segments[5] + "/specs/" + segments[7]
```
Krpate, line 118: A general observation unrelated to this change: go/go-style/best-practices#string-concat-fmt

demos/nytimes.sh
```diff
@@ -0,0 +1,135 @@
...
+	--project.description "APIs collected from the APIs.guru OpenAPI Directory"
+
+# Now compute summary details of all of the APIs in the project. 
+# This will log errors if any of the API specs can't be parsed,
+# but for every spec that is parsed, this will set the display name
+# and description of the corresponding API from the values in the specs.
+registry compute details projects/nytimes/apis/-
+
+# Verify this for one of the APIs.
+registry get projects/nytimes/apis/archive
```
Krpate, line 55: `locations/global` is missing in resource names in this entire file.

demos/walmart.sh
```diff
@@ -0,0 +1,137 @@
...
+
+# Verify this for one of the APIs.
+registry get projects/walmart/apis/inventory
+
+# You can also get this API with the lower-level apg command.
+# Add the --json option to get JSON-formatted output.
+apg registry get-api --name projects/walmart/apis/inventory --json
+
+# List all of the APIs in the project.
+registry list projects/walmart/apis
```
Krpate, line 62: `locations/global` is missing in the resource name in this file.

server/names/names_test.go
```diff
@@ -54,8 +55,8 @@ func TestResourceNames(t *testing.T) {
 			name:   "apis",
 			regexp: ApisRegexp(),
 			pass: []string{
-				"projects/google/apis",
-				"projects/-/apis",
+				"projects/google/locations/LOCATION/apis",
```
Krpate, line 58: Should this be `global`?
Here and everywhere below.

cmd/registry/core/get.go
timburks: Thanks! My old-school aversion to excessive function calls is showing... happy to change to fit the style guide, though

demos/nytimes.sh
timburks: Thanks for catching this - I'm going to remove this and the other organization-specific demo scripts from this PR since they weren't already in the repo

google/cloud/apigee/registry/v1/registry_models.proto
```diff
@@ -59,7 +59,7 @@ message Project {
 message Api {
   option (google.api.resource) = {
     type: "registry.googleapis.com/Api"
-    pattern: "projects/{project}/apis/{api}"
+    pattern: "projects/{project}/locations/{location}/apis/{api}"
```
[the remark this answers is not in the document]
timburks: I think it might be a breaking change (for client libraries) to switch from "global" to "{location}" (and I think that eventually we'll want to support non-global locations)

demos/nytimes.sh
timburks: Removed

demos/walmart.sh
timburks: Removed this file

server/names/names_test.go
timburks: This package is already generalized to support arbitrary locations (and confirmed with this test) -- and since I expect that we'll want to support non-global locations, I'd like to keep this.

cmd/registry/core/get.go
timburks: This involves several lines of change in this file and probably others, so I'd prefer to address this in a future PR

Krpate: Sounds good :)

google/cloud/apigee/registry/v1/registry_models.proto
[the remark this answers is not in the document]
Krpate: Makes sense. Just so that I understand it correctly, the core registry server will be able to serve requests like CreateApi under parent `projects/xyz/locations/us-central1`. But these requests should be blocked at the One Platform API level if the location is anything other than `global`(until we support the configuration), correct?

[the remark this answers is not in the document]
timburks: I'm thinking that on startup, the registry server would read a configuration that sets the "location" value (defaulting to "global"). This would be the location where the server is running and would be the only location that the server would recognize in requests or return in responses. Thus if the location was "us-central1", then "global" wouldn't work, and vice-versa.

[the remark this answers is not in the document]
Krpate: Thanks for clarifying!

Krpate (approved):

shrutiparabgoogle:
I have one clarifying question on this:

The path of resources with this change will look like this: `projects/{project}/locations/{location}/apis/{api}/versions/...`, this suggests that location is a property of the project and not the registry itself. The way I understand it is that `a particular project in registry can have multiple locations` as opposed to `a registry location hosting multiple projects` (`locations/{location}/projects/{project}/apis/{api}/versions/...`). 

My understanding of the locations property was the latter one (since a project can be created in a particular registry instance hosted at a particular location). 

It is totally possible that my understanding is different than what was agreed upon. This won't matter for `global` locations but might affect when we start supporting multiple locations.

timburks:
@shrutiparabgoogle We'll want to clarify that this resource structure exists to allow us to share an open source implementation that has exactly the same interface as its corresponding hosted service. In our hosting hierarchy, "project" is a top-level billing entity that can own resources in multiple locations.

A registry instance will be running in one of these locations, so for a hosted instance, both "project" and "location" will be fixed to constant values.

If we supported a configurable location, the location would probably be best specified in the registry-server config file. We could also modify the core to replace the project entity with a configurable "project" string, but at this point, our multi-project support is useful for testing and leaves room for us to potentially support multi-tenant (i.e. multi-project) instances.

shrutiparabgoogle:
> @shrutiparabgoogle We'll want to clarify that this resource structure exists to allow us to share an open source implementation that has exactly the same interface as its corresponding hosted service. In our hosting hierarchy, "project" is a top-level billing entity that can own resources in multiple locations.
> 
> A registry instance will be running in one of these locations, so for a hosted instance, both "project" and "location" will be fixed to constant values.
> 
> If we supported a configurable location, a registry instance would be running in a single location, so the location could just be specified in the registry-server config file.
> 
> Because we are also fixing hosted instances to single projects, we might also modify the core to replace the project entity with a configurable "project" string, but at this point, our multi-project support is useful for testing and leaves room for us to potentially support multi-tenant (i.e. multi-project) instances.

Got it makes sense! I guess I was missing context on the "project" hierarchy and wanting to be as close to the hosted APIs. Thanks for the clarification.

shrutiparabgoogle (approved):

PR #267, opened 2021-08-25
Add support for placeholder artifacts in the controller

shrutiparabgoogle:
Summary of changes:

Support for placeholder flag:
* Add a placeHolder flag in the manifest definition to mark the placeholder target artifacts. This helps the user understand that no. artifact information is stored in the registry, the artifact is just representing the action taken.
* From the implementation POV, we can make use of that flag to decide whether to execute a `touch` operation on the artifact.


shrutiparabgoogle:
> Mostly unimportant details, but there are some things to address. It's also not clear to me from this PR why we want placeholder artifacts. Is there additional context that would help me understand?

See the linked issue where I have explained the requirement for a placeholder flag.

cmd/registry/controller/parser.go
```diff
@@ -69,8 +70,37 @@ func ExtendSourcePattern(
 
 }
 
+func ResourceNameFromDependency(
```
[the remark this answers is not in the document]
seaneganx: That file is in the same package as this one. If that's the only place this is called I think we can unexport this for now. As a rule, you can know if you're calling a function from a different package because the package's name will need to be included, e.g. `regexp.MustCompile` is in the `regexp` package.

cmd/registry/controller/controller.go
```diff
@@ -168,24 +175,36 @@ func generateActions(
...
 			takeAction := true
 			var args []Resource
+			var resourceName string
 			if _, ok := visited[key]; !ok {
 				for _, dMap := range dependencyMaps {
 					collection, ok := dMap[key]
 					if ok {
+						// Since the GeneratedResource is non-existent here,
+						// we will have to derive the exact name of the target resource
+						resourceName, _ = ResourceNameFromDependency(
```
seaneganx, line 206: Now that this function has validation and can return an error, this should check that error and handle it. If ignoring it is the intention, the comment should explain why it's safe to ignore.

seaneganx (commented):
LGTM other than one outstanding comment. Deferring to @timburks for approval.

cmd/registry/controller/controller.go
shrutiparabgoogle: Done

timburks (approved):
The overall structure and plan LGTM. 

I don't want to block on this concern, but I think we should reconsider our use of the word "placeholder" for these artifacts. To me it suggests that we'll be replacing the placeholder with some other artifact later, and that's not the case. 

Searching for alternatives, I started by looking at [synonyms for "certificate"](https://www.powerthesaurus.org/certificate)  (which has a crypto-based definition which would surely confuse things if we used it). The word that most stood out to me was "receipt". WDYT?

shrutiparabgoogle:
> The overall structure and plan LGTM.
> 
> I don't want to block on this concern, but I think we should reconsider our use of the word "placeholder" for these artifacts. To me it suggests that we'll be replacing the placeholder with some other artifact later, and that's not the case.
> 
> Searching for alternatives, I started by looking at [synonyms for "certificate"](https://www.powerthesaurus.org/certificate) (which has a crypto-based definition which would surely confuse things if we used it). The word that most stood out to me was "receipt". WDYT?

I agree, we are not storing any temporary artifacts and placeholder might create confusion. "receipt" makes sense to me. since we. are actually storing the receipt of the executed action. I will follow-up on this rename effort in a separate PR: https://github.com/apigee/registry/issues/276

Issue #271, opened 2021-08-26
Viper-based environment variables clash with automatically-created Kubernetes environments


timburks:
While testing a Kubernetes deployment of the `registry-server`, I noticed that my server was being started on random ports instead of the port specified in my config file. Digging into this, I found that a call to `viper.GetInt("port")` was returning zero (thus the random port selection), even though the port in my config file was nonzero.

Looking at the environment in my Kubernetes container, I found several environment variables matching `REGISTRY_PORT`:
```
/ # printenv | grep REGISTRY_PORT
REGISTRY_PORT_80_TCP=tcp://10.152.183.62:80
REGISTRY_PORT=tcp://10.152.183.62:80
REGISTRY_PORT_80_TCP_ADDR=10.152.183.62
REGISTRY_PORT_80_TCP_PORT=80
REGISTRY_PORT_80_TCP_PROTO=tcp
```

Looking at these and the other environment variables in the container, I believe they were created because my cluster has a [service](https://kubernetes.io/docs/concepts/services-networking/service/)  that I named "registry". This caused the REGISTRY_PORT variable to be set to something other than an int, which I suspect caused viper.GetInt() to return zero, and because of Viper's precedence rules, this overruled the port that was explicitly set in my config file.

We might try to fix this by telling Kubernetes users "don't have a service named registry", but most users would probably miss that warning. 

We might also change the name of our environment variable or prefix to avoid collisions, possibly by changing our call to `viper.SetEnvPrefix("registry")` to use "registry_server" or some other prefix, but someone still might collide with anything that we choose, particularly if we try to pick something that is easy to remember or infer.

Or, instead of the above, we might drop environment-based configuration for `registry-server` altogether and only support file-based configuration, which seems to me to have the least possibility of ambiguity. It appears that we explicitly enable it, so removing it is (hopefully) probably not difficult.

seaneganx:
Again, let's stop talking about implementation details until we can agree what we want to implement. From what I've gathered in our discussion, these seem to be the most important features of configuration.

- Load from a YAML config file provided via startup flag.
- Load from the environment.
- Avoid collisions with k8s reserved environment variables.

Is there anything else we want for configuration as a feature?

timburks:
#273 is to put things back into a known-working state that removes complexity introduced by Viper, which we at no point agreed was appropriate to add to registry-server (for example home directory based configuration). I think that discussion of desired features should come after we revert, not before.

PR #273, opened 2021-08-26
Simplify registry-server configuration.

timburks:
This commit simplifies server configuration by replacing Viper-based configuration with a YAML reader that accepts environment variables in configuration source files. This allows users to directly see and control the environment variables that affect configuration and clarifies that config files have highest priority (making it easier to debug configuration problems). It also removes the home-directory-based configuration that seems useful for command-line tools but adds complexity to server deployments. The README also clarifies the default configuration (when no file is specified).

This does not preclude the use of Viper for configuration of command-line tools like the registry tool, but we should carefully review pros/cons before committing to that.

This was tested locally with existing CI tests, in Kubernetes with a ConfigMap-based configuration file, and in Docker with environment variable configuration.

Addresses concerns discussed in #271


timburks:
[Merged as admin, recognizing the lack of consensus in #271]

PR #278, opened 2021-09-01
Fixing the annotations in the proto file for List and Create Artifacts.

Krpate:


timburks (approved):
manually verified with a locally-running envoy (using configuration in [deployments/envoy](https://github.com/apigee/registry/tree/main/deployments/envoy)).

PR #280, opened 2021-09-01
Parser tests

shrutiparabgoogle:
Added tests for parser.go

### DIFF
--- a/cmd/registry/controller/parser.go
+++ b/cmd/registry/controller/parser.go
@@ -88,17 +88,24 @@ func resourceNameFromDependency(
 	//    returns projects/demo/locations/global/apis/petstore/versions/1.0.0/specs/openapi.yaml/artifacts/custom-artifact
 
 	// Replace apis/- pattern with the corresponding name and so on.
-	apiPattern := regexp.MustCompile(`.*(/apis/)-`)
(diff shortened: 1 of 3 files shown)


cmd/registry/controller/parser.go
```diff
@@ -84,17 +88,24 @@ func resourceNameFromDependency(
...
 
 	// Replace apis/- pattern with the corresponding name and so on.
-	apiPattern := regexp.MustCompile(`.*(/apis/)-`)
-	resourceName := apiPattern.ReplaceAllString(resourcePattern, fmt.Sprintf("%s", dependency.GetApi()))
+	// dependency.GetApi() returns the full api name projects/demo/locations/global/apis/petstore
+	// We use stringsSplit()[-1] to extract only the API name
+	// apiPattern := regexp.MustCompile(`/apis/-`)
+	apiName := strings.Split(dependency.GetApi(), "/")
+	resourceName := strings.Replace(resourcePattern, "/apis/-",
+		fmt.Sprintf("/apis/%s", apiName[len(apiName)-1]), -1)
```
seaneganx, line 96: Nit: `strings.ReplaceAll` seems like a better function to use here so you don't have to pass the `-1` each time.

seaneganx (approved):

cmd/registry/controller/parser.go

Issue #276, opened 2021-08-31
Registry controller: renaming placeholder flag


shrutiparabgoogle:
I think we should reconsider our use of the word "placeholder" for these artifacts. To me it suggests that we'll be replacing the placeholder with some other artifact later, and that's not the case.

Searching for alternatives, I started by looking at [synonyms for "certificate"](https://www.powerthesaurus.org/certificate) (which has a crypto-based definition which would surely confuse things if we used it). The word that most stood out to me was "receipt". WDYT?

Issue #274, opened 2021-08-27
Define a proto for storing receipt artifacts


shrutiparabgoogle:
In the current implementation of the placeholder artifacts, the only data stored in the artifact is an encoded string of the action which is executed by the controller. Define a simplistic proto which can store this information in the registry and we don't have to store the information as a string.

PR #281, opened 2021-09-01
Rename flag 'placeholder' to 'receipt' and define a proto

shrutiparabgoogle:
Summary of changes:
* Renamed flag `placeholder` to `receipt` https://github.com/apigee/registry/issues/276
* Define a proto for ReceiptArtifact https://github.com/apigee/registry/issues/274

### DIFF
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -26,7 +26,7 @@ import (
 type Action struct {
 	Command           string
 	GeneratedResource string
-	Placeholder       bool
+	Receipt           bool
 }
 
 type ResourceCollection struct {
@@ -182,7 +182,7 @@ func generateActions(
 			action := &Action{
 				Command:           cmd,
 				GeneratedResource: resource.GetName(),
-				Placeholder:       generatedResource.Placeholder,
+				Receipt:           generatedResource.Receipt,
 			}
 			actions = append(actions, action)
 		}
@@ -231,7 +231,7 @@ func generateActions(
 				action := &Action{
 					Command:           cmd,
 					GeneratedResource: resourceName,
-					Placeholder:       generatedResource.Placeholder,
+					Receipt:           generatedResource.Receipt,
 				}
 				actions = append(actions, action)
 			}
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -26,7 +26,7 @@ import (
 type Action struct {
 	Command           string
 	GeneratedResource string
-	Receipt           bool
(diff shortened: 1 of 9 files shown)


google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
```diff
@@ -46,7 +46,7 @@ message GeneratedResource {
 	];
 	string filter = 2;
 	//set to true if the action is not going to store any information in the GeneratedResource
-	bool placeholder = 3;
+	bool receipt = 3;
```
timburks, line 49: Could we make this "is_receipt" or something similar? (cc @seaneganx for recommendations) 

I think "receipt" by itself suggests that the receipt is true or false (a boolean) and what we mean here is that the generated resource is a receipt.

cmd/registry/controller/controller.go
```diff
@@ -26,7 +26,7 @@ import (
 type Action struct {
 	Command           string
 	GeneratedResource string
-	Placeholder       bool
+	Receipt           bool
```
timburks, line 29: Does this mean that this action requires a receipt? (if so it might be clearer to name this "RequiresReceipt")

timburks (approved):
Approved with some light thoughts about a couple of names, final decision is up to you.

google/cloud/apigee/registry/applications/v1alpha1/registry_manifest.proto
seaneganx: Using an "is_" prefix is recommended against in [AIP-140](https://google.aip.dev/140#booleans). I think adding documentation to the field will be sufficient because it's an obvious location to seek clarification if someone doesn't have enough context.

Note: Some people (myself included) are already going to read this automatically as "`GeneratedResource` is a `receipt`". If someone wanted to suggest that the receipt is a bool itself, I would expect them to have a `Receipt` message with a field `value`.

timburks: OK - receipt SGTM

shrutiparabgoogle: I updated the comment to make it more clear  (the purpose of the receipt field)

cmd/registry/controller/controller.go
shrutiparabgoogle: Updated this to `RequiresReceipt`, makes the code more readable. Thanks!

shrutiparabgoogle:
Thanks for the feedback!

PR #283, opened 2021-09-03
Implement API Style Guide Proto definition, and ability to upload API Style Guides in YAML Format

muhammadharis:
## Overview:
API Style Guides are a declarative solution to define the static analysis of API specs within a project. An API Style Guide is described by the proto that is defined in this pull request. It contains a list of guidelines along with linters that enforce those guidelines.

This pull request creates the API Style Guide proto, and also implements a CLI to upload YAML files that describe an API Style Guide to the registry, where it is attached as an artifact to a provided project ID.

## Test:

1. Create a file called `styleguide.yaml` in your home directory, with the following contents:
```yaml
name: openapistyleguide
mime_types:
guidelines:
  - name: refproperties
    display_name: Govern Ref Properties
    rules:
      - name: norefsiblings
        description: An object exposing a $ref property cannot be further extended with additional properties.
        linter: spectral
        linter_rulename: no-$ref-siblings
        severity: 1
    status: 1
linters:
  - name: spectral
    uri: https://github.com/stoplightio/spectral
```

2. Ensure that the API Registry Server is running on localhost:8080
3. Run the following commands on the registry:
```bash
apg registry delete-project --name projects/protos

apg registry create-project --project_id protos --insecure --address localhost:8080

registry upload styleguide ~/styleguide.yaml --project_id protos

registry get projects/protos/locations/global/artifacts/openapistyleguide --contents
```

### DIFF
--- a/cmd/registry/cmd/upload/styleguide.go
+++ b/cmd/registry/cmd/upload/styleguide.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
 	"io/ioutil"
@@ -31,14 +32,16 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-func styleGuideRelation(styleGuideName string) string {
-	return "styleguide-" +
-		strings.ReplaceAll(
-			strings.ToLower(styleGuideName), " ", "-",
-		)
+// normalizeArtifactId makes a provided Artifact Id a lower-cased alphanumeric string
+func normalizeArtifactId(resourceName string) string {
+	reg, err := regexp.Compile("[^a-zA-Z0-9]+")
+	if err != nil {
+		log.Fatal(err)
+	}
+	return strings.ToLower(reg.ReplaceAllString(resourceName, ""))
 }
 
-func readStyleGuideProto(filename string) (*rpc.APIStyleGuide, error) {
+func readStyleGuideProto(filename string) (*rpc.StyleGuide, error) {
 
 	yamlBytes, err := ioutil.ReadFile(filename)
 	if err != nil {
@@ -46,7 +49,7 @@ func readStyleGuideProto(filename string) (*rpc.APIStyleGuide, error) {
 	}
 
 	jsonBytes, err := yaml.YAMLToJSON(yamlBytes)
-	m := &rpc.APIStyleGuide{}
+	m := &rpc.StyleGuide{}
 	err = protojson.Unmarshal(jsonBytes, m)
 
 	if err != nil {
@@ -84,7 +87,7 @@ func styleGuideCommand(ctx context.Context) *cobra.Command {
 				Name: "projects/" +
 					projectID +
 					"/locations/global/artifacts/" +
-					styleGuideRelation(styleGuide.GetName()),
+					normalizeArtifactId(styleGuide.GetName()),
 				MimeType: core.MimeTypeForMessageType(
 					"google.cloud.apigee.registry.applications.v1alpha1.styleguide",
 				),
--- a/cmd/registry/cmd/upload/styleguide.go
+++ b/cmd/registry/cmd/upload/styleguide.go
@@ -1,0 +1,104 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package upload
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+
+	"io/ioutil"
+
+	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"github.com/ghodss/yaml"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+func styleGuideRelation(styleGuideName string) string {
+	return "styleguide-" +
+		strings.ReplaceAll(
+			strings.ToLower(styleGuideName), " ", "-",
+		)
+}
+
+func readStyleGuideProto(filename string) (*rpc.APIStyleGuide, error) {
+
+	yamlBytes, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	jsonBytes, err := yaml.YAMLToJSON(yamlBytes)
+	m := &rpc.APIStyleGuide{}
+	err = protojson.Unmarshal(jsonBytes, m)
+
+	if err != nil {
+		return nil, fmt.Errorf("in file %q: %v", filename, err)
+	}
+
+	return m, nil
+}
+
+func styleGuideCommand(ctx context.Context) *cobra.Command {
+	var projectID string
+	cmd := &cobra.Command{
+		Use:   "styleguide FILE_PATH --project_id=value",
+		Short: "Upload an API style guide",
+		Args:  cobra.MinimumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			styleGuidePath := args[0]
+			if styleGuidePath == "" {
+				log.Fatal("Please provide style guide path")
+			}
+
+			styleGuide, err := readStyleGuideProto(styleGuidePath)
+			if err != nil {
+				log.Fatal(err.Error())
+			}
+			styleGuideData, _ := proto.Marshal(styleGuide)
+
+			ctx := context.Background()
+			client, err := connection.NewClient(ctx)
+			if err != nil {
+				log.Fatalf("%s", err.Error())
+			}
+
+			artifact := &rpc.Artifact{
+				Name: "projects/" +
+					projectID +
+					"/locations/global/artifacts/" +
+					styleGuideRelation(styleGuide.GetName()),
+				MimeType: core.MimeTypeForMessageType(
+					"google.cloud.apigee.registry.applications.v1alpha1.styleguide",
+				),
+				Contents: styleGuideData,
+			}
+			log.Printf("uploading %s", artifact.Name)
+			err = core.SetArtifact(ctx, client, artifact)
+			if err != nil {
+				log.Fatal(err.Error())
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&projectID, "project_id", "", "Project ID to use when storing the styleguide artifact")
+	cmd.MarkFlagRequired("project_id")
+	return cmd
+}
--- a/cmd/registry/cmd/upload/styleguide.go
+++ b/cmd/registry/cmd/upload/styleguide.go
@@ -18,8 +18,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"regexp"
-	"strings"
 
 	"io/ioutil"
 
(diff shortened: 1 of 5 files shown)


google/cloud/apigee/registry/applications/v1alpha1/registry_api_style_guide.proto
```diff
@@ -0,0 +1,107 @@
...
+    // A list of guidelines that are specified by this style guide.
+    repeated Guideline guidelines = 3;
+
+    // A list of linters that this style guide uses.
+    repeated Linter linters = 4;
+}
+
+message Guideline {
+    // Resource name of the guideline.
+    string name = 1 [
```
timburks, line 54: As we've discussed them, these fields are identifiers and not resource names. 

@seaneganx WDYT about calling these fields "id" instead of "name"? I'm uncomfortable with "name" because the AIPs say that name should be a full resource name, and I think in the context of this YAML format, it would be redundant and awkward to spell out the full name.

We also might want to restrict the character set to valid identifier characters (which don't include $, which I see in the examples, which might mean that we need an additional field for linter_rulename).

cmd/registry/cmd/upload/styleguide.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"github.com/ghodss/yaml"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+// normalizeArtifactId makes a provided Artifact Id a lower-cased alphanumeric string
+func normalizeArtifactId(resourceName string) string {
```
shrutiparabgoogle, line 36: This function is doing more than just doing lower-case conversion. IIUC it is also taking out all the non-alphanumeric chars from the `Name`.  There is already a name validation in place for artifactIDs. 
Example: Trying to upload a manifest with name: `Test-Manifest`
```
$ registry upload manifest cmd/registry/cmd/resolve/testdata/manifest.yaml --project_id=demo
[ 63550255 ] 2021/09/08 13:43:21 uploading projects/demo/locations/global/artifacts/Test-Manifest
[ 63550255 ] 2021/09/08 13:43:21 rpc error: code = InvalidArgument desc = invalid identifier "Test-Manifest": must match "^[a-z0-9-.]+$"
```

Is there a specific reason why we are removing other characters? I think we, currently allow hyphens in the artifact names. A comment explaining why might help.

If we do this, then I suggest we should log this operation happening. The artifact name that we will end up storing in the registry will be different than what the user defined in the configuration, the user should be notified of this change through the logs.

muhammadharis: Makes sense to me. I was attempting to sanitize a user-provided name (which is possibly invalid) into a valid name. However, I think the method you're describing is more conventional in the registry, which is to simply throw an error if an invalid name is encountered. Fixed in the recent update 👍

timburks (approved):

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
```diff
@@ -0,0 +1,121 @@
...
+    // A list containing style (format) descriptors for this style guide that 
+    // are specified as a Media Type (https://en.wikipedia.org/wiki/Media_type).
+    // Possible values include "application/vnd.apigee.proto", 
+    // "application/vnd.apigee.openapi", and "application/vnd.apigee.graphql",
+    // with possible suffixes representing compression types. These hypothetical
+    // names are defined in the vendor tree defined in RFC6838 
+    // (https://tools.ietf.org/html/rfc6838) and are not final.
+    repeated string mime_types = 2 [
+		(google.api.field_behavior) = REQUIRED
+	];
```
seaneganx, line 43: Why multiple MIME types? It's unclear to me from this description how I would use multiple types for one style guide. Do you have an example or something that can be added to the description to clarify?

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
```diff
@@ -0,0 +1,121 @@
...
+
+    // A list of guidelines that are specified by this style guide.
+    repeated Guideline guidelines = 3;
+
+    // A list of linters that this style guide uses.
+    repeated Linter linters = 4;
+}
+
+message Guideline {
```
seaneganx, line 52: Nit: This message should have documentation. So should the `Rule` and `Linter` messages.

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
```diff
@@ -0,0 +1,121 @@
...
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // A detailed description of the guideline.
+    string description = 3;
+
+    // A list of rules that this guideline specifies.
+    repeated Rule rules = 4;
+
+    // Indicates the status of the guideline.
```
seaneganx, line 69: Nit: This seems like it should be moved to document the `status` field instead of the `Status` enum definition. The enum might be better described as the possible statuses of a guideline.

This feedback applies similarly to the `Rule` severity type and field definitions.

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
```diff
@@ -0,0 +1,121 @@
...
+    enum Status {
+        PROPOSED = 0;
```
seaneganx, line 71: Is this a reasonable default value for the status? [AIP-126](https://google.aip.dev/126#guidance) recommends having an "unspecified" default value if there isn't any clear "default" status.

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
```diff
@@ -0,0 +1,121 @@
...
+	];
+
+    // Name of the rule on the linter that enforces it.
+    string linter_rulename = 5 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // Severity of violating this rule.
+    enum Severity {
+        ERROR = 0;
```
seaneganx, line 103: Is this a reasonable default value for the severity? [AIP-126](https://google.aip.dev/126#guidance) recommends having an "unspecified" default value if there isn't any clear "default" severity.

seaneganx (commented):
Here's some questions I have and some feedback to consider in future changes to this code.

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
timburks: In the short term this allows us to match variations in mime type strings related to compression or OpenAPI version without someone having to write a general mime type matching service.

In the long term this allows for a style guide to support specs of different formats - we expect but wouldn't require all style guides to apply to exactly one format - this would really depend on what the linters being used would accept. We also considered that an official mime type for OpenAPI (outside of the vendor space) might appear and that a registry might include a mixture of these types (messy but not the messiest thing that has ever happened)

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
timburks: +1 to adding `STATUS_UNSPECIFIED = 0` as the first value

google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
timburks: As above, I agree that `SEVERITY_UNSPECIFIED = 0` would be a good first value.

timburks:
@seaneganx I noticed that although the CI tests passed when we ran them on this PR, after it was merged, CI on main now fails due to formatting errors. When I check out main and run `gofmt -s -l .` I see no problems (with `go version go1.16.2 darwin/amd64`). Did formatting change with `1.17.0`?

timburks:
Here's what changed when I upgraded to Go 1.17.1:
```
 git diff server/internal/storage/gorm/cgotrue.go 
diff --git a/server/internal/storage/gorm/cgotrue.go b/server/internal/storage/gorm/cgotrue.go
index e0fb71f..0b0b8f9 100644
--- a/server/internal/storage/gorm/cgotrue.go
+++ b/server/internal/storage/gorm/cgotrue.go
@@ -1,3 +1,4 @@
+//go:build cgo
 // +build cgo
 ```
(same for cgofalse.go)

**Comment by timburks** [comment]
https://golang.org/doc/go1.17#gofmt

**Comment by timburks** [comment]
I put this change in #286

File: google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto (hunk 3 above)
Comment (seaneganx) [reply google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto:43]: You've lost me from the first sentence because I'm failing to understand how the field is used. I'll have to revisit your points after you help me get even a basic grasp on what this field does. An example would be really helpful. What does it mean if I create a style guide with the following MIME types?

```
mime_types = [
    "application/vnd.apigee.openapi",
    "application/vnd.apigee.graphql",
]
```

I currently have no idea what that represents, and I've tried my best to figure it out given the documentation. That said, there's almost certainly room for improvement in this field's comment. If I can't even figure this out, who is the intended audience of this documentation? What are we assuming they understand a priori?

>A list containing style (format) descriptors for this style guide that are specified as a Media Type

I'll call out this sentence specifically. What does it mean? How can a style guide have multiple formats simultaneously? Unless the formats are overlapping (e.g. JSON and YAML) I'm not seeing how this makes any sense.

Any corrections on my misunderstandings or answers to my questions would be appreciated.

Comment (muhammadharis) [reply google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto:43]: @seaneganx I agree that the documentation can and will be improved in a subsequent PR. Here's an example of what Tim was talking about:

```yaml
mime_types:
   - application/x.openapi+gzip;version=3.0.0
```

As in, "the rules in this style guide apply to OpenAPI Version 2 and Version 3 specs". In the future, the registry may support mime type parsing (where we can parse out the message, compression, version and this can be refactored, and so the user would simply write:

```yaml
mime_types:
   - application/x.openapi
```

Until then, we're choosing to allow multiple mime types. I'll make this explicit and clear in a subsequent PR that I'll open up, because the documentation does not reflect this currently.

seaneganx: Thanks, that clears everything up. This field lists the MIME types of the specs that the style guide applies to.

Without that in the documentation I misunderstood it as the MIME types of the style guide itself since it's a field on the style guide resource. Our other APIs have MIME type fields that indicate that resource's type, so we will want to make this difference clear.

Now for a subtle detail. Are rules generic or do they only apply to specific MIME types?

muhammadharis: The short answer is yes. Each style guide applies to a list of given MIME types. The style guide has a list of guidelines, which contain rules. For now, it is the user's job to make sure that the rules that they provide in the guidelines of a style guide actually apply to the mime type that is specified in the style guide. When this is actually being implemented (and I get to the point where I'm implementing error-handling), I will implement a way to make sure that the rules that the user is applying to the MIME type are actually valid for that specific MIME type.

timburks: We should keep in mind that the style guide YAML will serve as a user interface for people who are writing and configuring style guides, so we probably want the format to be as clear and unsurprising as possible.

Subjectively, I don't expect that we would see many style guides that contain mixtures of rules where some rules apply to some mime types and other rules apply to different mime types. So putting a mime type qualifier on individual rules seems tedious and surprising to users. If we did allow it (maybe we should for exceptional cases), I think it would only be used in those exceptional cases.

I (again subjectively) don't feel obligated to have resources "all the way down", but even if we did, the user benefit (that I perceive) of easier authoring suggests to me that it's better to keep the mime type list at the style guide level.

This might get clearer when we write some sample style guide YAMLs, like an externalization of the api-linter guidelines, for example.

(I wouldn't think of the style guide as an ancestor to the rules as much as an owner of or context for the rules - so this seems more like a delegation pattern than inheritance)

seaneganx: I agree this is a good way to represent the YAML for authoring by users, but I don't think it's a great representation for the application to handle. What's best for users is often not the best internal representation for an application.

>Subjectively, I don't expect that we would see many style guides that contain mixtures of rules where some rules apply to some mime types and other rules apply to different mime types. So putting a mime type qualifier on individual rules seems tedious and surprising to users. If we did allow it (maybe we should for exceptional cases), I think it would only be used in those exceptional cases.

For design flexibility, I think we should consider using rule-level MIME types on the protos as I suggested, but decouple it from the YAML representation which can be kept as-is for ease of authoring. For example, the [server config](https://github.com/apigee/registry/blob/main/cmd/registry-server/main.go) does this, converting from an ideal user representation in YAML to a different internal format. 

This approach doesn't require us to support the "exceptional" cases, but makes it easy to add that functionality if we want to in the future because it's a more flexible (and IMO application-friendly) design. The change would only affect the YAML parsing code instead of everything that touches the protos.

timburks: OK, I think that's worth considering in the future. These are alpha protos, so we're not committing to them permanently, just as a good starting point.

PR #285, opened 2021-09-07
Add support for third party commands in registry-tool

shrutiparabgoogle:
Summary of contributions:
* Added support for executing third-party commands through the registry-tool.
* Added test case for receipt artifact, to demonstrate the use of third-party commands.


cmd/registry/cmd/root.go
```diff
@@ -43,6 +44,7 @@ func Command(ctx context.Context) *cobra.Command {
 	cmd.AddCommand(resolve.Command(ctx))
 	cmd.AddCommand(delete.Command(ctx))
 	cmd.AddCommand(export.Command(ctx))
+	cmd.AddCommand(exec.Command(ctx))
```
timburks, line 47: super-nit: this is out of alphabetic order

cmd/registry/cmd/exec/exec.go
```diff
@@ -0,0 +1,54 @@
...
+			commandStr := args[0]
+			if commandStr == "" {
+				log.Fatal("please provide a command to execute")
+			}
+
+			commandFields := strings.Fields(commandStr)
+			command := commandFields[0]
+			commandArgs := []string{}
+			if len(commandFields) > 1 {
+				commandArgs = commandFields[1:]
```
timburks, line 42: This appears to pass all the remaining arguments to the executed command (which I think is good) but I noticed that the test (below) only passes a single argument to "echo" - it might be worth adding a couple of extra arguments to that test example to protect this behavior.

timburks (approved):

cmd/registry/cmd/exec/exec.go
shrutiparabgoogle: Done.

cmd/registry/cmd/root.go
shrutiparabgoogle: Done.

seaneganx (commented):
Why are we creating a Registry tool command to execute 3p bash commands instead of simply asking for 3p commands directly in the manifest?

Or rephrased as an example, why specify `exec echo ...` and `compute lint ...` in the manifest when we can use `echo ...` and `registry compute lint ...` instead?

shrutiparabgoogle:
> Why are we creating a Registry tool command to execute 3p bash commands instead of simply asking for 3p commands directly in the manifest?
> 
> Or rephrased as an example, why specify `exec echo ...` and `compute lint ...` in the manifest when we can use `echo ...` and `registry compute lint ...` instead?

There is not much difference in the two approaches, since both the approaches are executing commands in a subprocess. One advantage that we get by doing this is, if we go through the `registry tool`, we get to append the `taskIDs` to the output of the 3p subCommands which are executing. If we directly execute the 3p command, we don't have any control over the output generated and no way to associate the logs generated to a particular task.

Makes sense? Open to evaluate advantages of the other approach if I missed any.

seaneganx:
>One advantage that we get by doing this is, if we go through the registry tool, we get to append the taskIDs to the output of the 3p subCommands which are executing. If we directly execute the 3p command, we don't have any control over the output generated and no way to associate the logs generated to a particular task.

Please correct me if I'm misunderstanding, but I think we can do this in both approaches. As you mentioned, in both we use the `os/exec` package to run the command in a new process. That gives us control over the stdout and stderr streams. We can also identify whether the command is 1p or 3p before executing it by inspecting the command string in either approach. 

I think it will be easier for users if they can provide 1p and 3p commands exactly as they would call them in bash instead of requiring a special syntax (the `exec` subcommand) for 3p commands. This approach will also have less code to maintain without the `exec` subcommand.

shrutiparabgoogle:
> > One advantage that we get by doing this is, if we go through the registry tool, we get to append the taskIDs to the output of the 3p subCommands which are executing. If we directly execute the 3p command, we don't have any control over the output generated and no way to associate the logs generated to a particular task.
> 
> Please correct me if I'm misunderstanding, but I think we can do this in both approaches. As you mentioned, in both we use the `os/exec` package to run the command in a new process. That gives us control over the stdout and stderr streams. We can also identify whether the command is 1p or 3p before executing it by inspecting the command string in either approach.
> 
> I think it will be easier for users if they can provide 1p and 3p commands exactly as they would call them in bash instead of requiring a special syntax (the `exec` subcommand) for 3p commands. This approach will also have less code to maintain without the `exec` subcommand.

We can always read the err and out stream of the exec commands, but we decided in our previous discussions that letting the subcommands log their own output/err is a better approach than reading the streams, aggregating and modifying them. The controller should only know whether the subcommand was a success or failure and should not care about the output that it generated.

The way the TaskID implementation currently works is as follows:
* the registry rootCommand adds a taskID prefix in the log (https://github.com/apigee/registry/blob/main/cmd/registry/main.go#L30)
* TODO: Pass a flag to the registry command which the controller can set to a particular taskID, and use that value as a log prefix. This way we can connect the tasks generated by the controller and the output generated by each of them. 

I agree with your point that letting the users specify the bash command as is in the config adds simplicity to the manifest. 

One thing I would like to add here is, the mention of the word `exec` is also a way to indicate to the user that the action is a 3p command, it will help users identify whether they are running 1p or 3p commands and if 3p, then the generated resources will usually be a `receipt` artifact. The way I look at it is a way to mark the 3p vs 1p commands with a prefix `exec`.

seaneganx:
>We can always read the err and out stream of the exec commands, but we decided in our previous discussions that letting the subcommands log their own output/err is a better approach than reading the streams, aggregating and modifying them. The controller should only know whether the subcommand was a success or failure and should not care about the output that it generated.

Right. My point is that we have the same level of control and information in either approach, and therefore neither one has functional advantages. I didn't mean to imply we should read/modify the outputs - just that we can make that choice in either approach.

>TODO: Pass a flag to the registry command which the controller can set to a particular taskID, and use that value as a log prefix. This way we can connect the tasks generated by the controller and the output generated by each of them.

I think this is the best approach since we don't have any easy way to propagate `context.Context` between processes in the multi-process solution. Internally we need to identify whether the command is 1p or 3p to pass this flag to 1p commands, and we can do that in either approach.

>I agree with your point that letting the users specify the bash command as is in the config adds simplicity to the manifest.

Okay. Providing bash commands as-is means that users don't make a distinction between 1p and 3p commands. Since we agree on this, what specifically is the benefit to asking users to provide 3p commands with the `exec` prefix?

>One thing I would like to add here is, the mention of the word exec is also a way to indicate to the user that the action is a 3p command, it will help users identify whether they are running 1p or 3p commands and if 3p, then the generated resources will usually be a receipt artifact. The way I look at it is a way to mark the 3p vs 1p commands with a prefix exec.

Right. The `exec` subcommand identifies the difference if we want users to think about 1p and 3p commands as different concepts. Similarly without the `exec` subcommand though, users will know if a command is 1p or 3p because 1p commands will use the `registry` program. It's clear that `registry compute lint ...` is 1p and `echo "hello world"` is 3p.

Circling back to my earlier question. What specifically do we gain from asking users to call 3p commands using `exec` instead of providing normal bash commands both 1p and 3p?

timburks:
> Circling back to my earlier question. What specifically do we gain from asking users to call 3p commands using `exec` instead of providing normal bash commands both 1p and 3p?

Chiming in here, I think we're talking about two alternatives for how we specify actions in the manifest:
(1) prefix all the external commands with `exec`
(2) prefix all the registry tool commands with `registry`

I had been thinking in terms of (1) but agree that (2) is at least as good and seems more future-oriented, especially if we expect more capabilities to be defined outside the registry tool than in it.

If we use the "registry" prefix, that will somewhat commit us to keeping "registry" as the tool name since we'll be putting it all through our manifest files (although not totally necessarily since we could accept multiple initial words as synonyms for "registry")

seaneganx:
>Chiming in here, I think we're talking about two alternatives for how we specify actions in the manifest:
(1) prefix all the external commands with exec (with no prefix for registry tool commands)
(2) prefix all the registry tool commands with registry (with no prefix for the external commands)

Correct, although the phrasing for (2) isn't really how I think about it. In that case the `registry` commands _are_ external since the "prefix" is actually just the name of the external program being executed, same as any other "external" command.

>If we use the "registry" prefix, that will somewhat commit us to keeping "registry" as the tool name since we'll be putting it all through our manifest files (although not totally necessarily since we could accept multiple initial words as synonyms for "registry")

Is there a reason why we can't simply update the commands in the manifests when we update the program name? I don't see why using the name `registry` in manifests would commit us to anything that our scripts and documentation using the name don't already lock us into.

shrutiparabgoogle:
```
Correct, although the phrasing for (2) isn't really how I think about it. In that case the registry commands are external since the "prefix" is actually just the name of the external program being executed, same as any other "external" command.
```
One more point to add here. I feel that there should always be a distinguish between internal(1p) and external(3p) commands **from the user's perspective**. 
* From the user's perspective, internal commands take care of everything from fetching resource from the registry, running analysis and submitting results back to the registry.
* Whereas, external commands will have to take care of all these things. (The users should be aware that they need to fetch the resource (if needed), execute an action and include a `receipt` flag for generating the receipt artifacts).

I agree that from the implementation POV, we can treat both as external commands (since we are simply `exec`ing). However, the users should always be aware of what actions they are running 1p or 3p. Having said that, I agree that both option 1) and 2) offer this distinction with either a `registry` or `exec` prefix.

One open question which remains (as mentioned earlier) is, connecting the 3p subcommand logs with taskID prefixes. Let's have a discussion on this @seaneganx

timburks:
This naming question is worth discussing but I think the suggestion that we change the way we call built-in registry commands goes outside the scope of this PR. For the sake of velocity, I think we should proceed with this as it's been built and address remaining questions in issues and future PRs.

seaneganx (approved):

PR #287, opened 2021-09-10
Pin the Go version used for CI to 1.15.*

timburks:
Temporary change to keep CI working while we make changes needed for 1.16 and 1.17

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
-        go-version: ^1.15
+        go-version: 1.15.15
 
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
-        go-version: ^1.15
+        go-version: 1.15.15
```
seaneganx, line 51: Let's pin to `1.15.*`, `1.15.x`, or `~1.15` (these should all be equivalent) to keep up-to-date with patches to the 1.15 minor version.

seaneganx (changes requested):

seaneganx (approved):

PR #288, opened 2021-09-10
Improve documentation on Style Guide, and do code cleanup for uploading functionality

muhammadharis:
This pull request addresses comments left post-merge on #283. It addresses all but one, which is to break out mime types into a constant field. Such a change is broad and can be part of a refactor where it is also done in other places in the registry.

### DIFF
--- a/google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
+++ b/google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
@@ -69,11 +69,23 @@ message Guideline {
     // A list of rules that this guideline specifies.
     repeated Rule rules = 4;
 
+    // Possible severities for this guideline.
     enum Status {
-        UNSPECIFIED = 0;
+        // The default value, unused.
+        STATUS_UNSPECIFIED = 0;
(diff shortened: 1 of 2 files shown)


google/cloud/apigee/registry/applications/v1alpha1/registry_styleguide.proto
```diff
@@ -98,16 +104,19 @@ message Rule {
...
     enum Severity {
-        ERROR = 0;
-        WARNING = 1;
-        INFO = 2;
-        HINT = 3;
+        UNSPECIFIED = 0;
+        ERROR = 1;
+        WARNING = 2;
+        INFO = 3;
+        HINT = 4;
```
seaneganx, line 112: As above, the unspecified field should be `SEVERITY_UNSPECIFIED` and it would be nice to have comments on the type and these values explaining the different severities.

seaneganx (commented):
Looks good but there are some small details to address. I'll defer to Tim for final approval.

timburks (approved):
Looks great! Thanks @seaneganx for your reviews/guidance!

PR #290, opened 2021-09-11
Add tests for controller error paths

shrutiparabgoogle:
Summary of changes:
* Added error path tests to controller
* Refactored `controller_test.go` to use the table driven approach
* Added demo scripts for controller
* Improved error logging in the controller package


Makefile
```diff
@@ -60,7 +60,6 @@ ifndef REGISTRY_MANIFEST_ID
 	@echo "Error! REGISTRY_MANIFEST_ID must be set."; exit 1
 endif
 	gcloud container clusters get-credentials registry-backend --zone us-central1-a
-	export DEFAULT_SA_KEY=$(kubectl get secrets -o jsonpath="{.items[0].metadata.name}" | grep default-token)
```
timburks, line 63: Why was this removed? It doesn't seem related to testing.

timburks (approved):
LGTM - nit: the commit contains an empty README.md file (tests/controller/README.md).

Makefile
shrutiparabgoogle: The default SA key is not required in the job since the registry server we are deploying in GKE is without an auth server. The key was present previously, but was not made use of.

I made changes to `deployments/controller/gke-job/cron-job.yaml` as well to take out the `APG_REGISTRY_TOKEN` env variable.

shrutiparabgoogle:
> LGTM - nit: the commit contains an empty README.md file (tests/controller/README.md).

Thanks for catching it. There was a one lint description which got missed out from the initial commit. Updated the patch to include the ReadMe changes

Makefile
timburks: SGTM - thanks!

cmd/registry/controller/parser.go
```diff
@@ -42,13 +42,15 @@ func extendDependencyPattern(
...
 
-	entityRegex := regexp.MustCompile(fmt.Sprintf(`\%s\.(api|version|spec|artifact)`, resourceKW))
+	entityRegex := regexp.MustCompile(fmt.Sprintf(`(\%s\.(api|version|spec|artifact))(/|$)`, resourceKW))
 	matches := entityRegex.FindStringSubmatch(dependencyPattern)
-	if len(matches) <= 1 {
-		return "", errors.New(fmt.Sprintf("Invalid source pattern: %s", dependencyPattern))
+	// dependencyPattern: "$resource.api/artifacts/score"
+	// matches: ["$resource.api/", "$resource.api", "api"]
+	if len(matches) <= 2 {
+		return "", errors.New(fmt.Sprintf("Invalid dependency pattern: %s", dependencyPattern))
```
seaneganx, line 50: Nit: `errors.New(fmt.Sprintf(...))` is equivalent to `fmt.Errorf(...)`. Also, error messages (only in the `error` type) should not be capitalized since they're often nested within other errors.

seaneganx (approved):

cmd/registry/controller/parser.go

PR #295, opened 2021-09-14
Use structured logger for registry tool

### LINKED ISSUE #231: Use configurable logging library
Currently we use the standard library's `log` package for the registry server and registry tool logging. The registry server supports five configurable logging levels, but only three different levels of verbosity. The registry tool does not support custom log levels. Logs are not configurable to save to disk.

Using [google/logger](https://pkg.go.dev/github.com/google/logger) we can print logs using `Info`, `Error`, `Warning`, or `Fatal` according to the situation where the logger is called. This allows more control over log visibility than the interceptors currently used by the server. We can also control the formatting of these logs.

seaneganx:
Currently this will have little effect on log contents and behavior. We do not have any method of configuring logs for the registry tool, so all existing logs will be displayed in a user friendly format. After configuration is added, we can set up different handlers for local vs. production development. For example, printing logs as JSON in production environments for integration with Cloud Logging.

All logs that used `log.Print` or `log.Printf` now use `log.Debug` or `log.Debugf`. Logs that used `log.Fatal` or `log.Fatalf` remain as fatal logs. In the future we may establish guidance on what kinds of information to log at other levels like `warn`, `info`, and `error`.

The apex/log library was chosen because it's lightweight, provides an interface to easily log at multiple levels, and gives full control of output format using custom handlers.

Fixes #231.


shrutiparabgoogle:
Looks great! 

There were some updates made to [controller.go](https://github.com/apigee/registry/blob/main/cmd/registry/controller/controller.go) in the latest PR, adding some log debug messages. As this PR is updating the logging library for the whole repo, we can update controller.go as well in this PR.

seaneganx:
>There were some updates made to controller.go in the latest PR, adding some log debug messages. As this PR is updating the logging library for the whole repo, we can update controller.go as well in this PR.

This PR only updated logs in the `cmd/registry/cmd/` directory. It should probably cover everything in the `cmd/registry` directory though. I'll make those changes now.

timburks (approved):

shrutiparabgoogle (approved):

PR #296, opened 2021-09-14
Conformance command to compute Lint Results based on API Style Guides

muhammadharis:
## Summary

This pull request aims to add the `conformance` subcommand under the `compute` command in the registry CLI. This command, when passed the path of an API Spec, will compute and attach Lint objects as artifacts to the spec according to the API Style Guides defined on the project that contains the spec.

To go into detail, consider the following scenario: there are three API style guides present on the project named `example` that govern OpenAPI, AsyncAPI, and GraphQL specs. Then, the user calls 
```bash
registry compute conformance projects/example/locations/global/apis/-/versions/-/specs/-
```
After this, the three style guides under the project are looked up. Each linter is configured to use the rules defined for it across all three style guides. For each spec, the mime type is looked up. If the mime type matches the mime type a linter supports (OpenAPI, AsyncAPI, or GraphQL), the linter is called on the spec, and the resulting Lint object is stored as an artifact on the spec.

## Note
Currently, the Lint objects being attached are blank objects. In a subsequent pull request, the logic of the Spectral and API Linter plugins will be implemented, which will do the actual linting and populate the lint objects.

## Test

- Have registry running on `localhost:8080`.
- Ensure that https://github.com/APIs-guru/openapi-directory is extracted to your Desktop.
- Ensure that the following style guide is located under `~/styleguide.yaml`:
```yaml
name: openapistyleguide
mime_types:
guidelines:
  - name: refproperties
    display_name: Govern Ref Properties
    rules:
      - name: norefsiblings
        description: An object exposing a $ref property cannot be further extended with additional properties.
        linter: spectral
        linter_rulename: no-$ref-siblings
        severity: 1
    status: 1
linters:
  - name: spectral
    uri: https://github.com/stoplightio/spectral
```

Now, run the following series of bash commands:
```bash
apg registry delete-project --name projects/example

apg registry create-project --project_id example --insecure --address localhost:8080

registry upload bulk openapi --project_id example ~/Desktop/openapi-directory/APIs

registry upload styleguide ~/styleguide.yaml --project_id=example

registry compute conformance projects/example/locations/global/apis/zappiti.com/versions/4.15.174/specs/swagger.yaml

registry list projects/example/locations/global/apis/zappiti.com/versions/4.15.174/specs/-/artifacts/-
```

**Expected output**:
```bash
projects/example/locations/global/apis/zappiti.com/versions/4.15.174/specs/swagger.yaml/artifacts/lint-spectral
```


cmd/registry/cmd/compute/compute.go
```diff
@@ -26,6 +26,7 @@ func Command(ctx context.Context) *cobra.Command {
 		Short: "Compute properties of resources in the API Registry",
 	}
 
+	cmd.AddCommand(conformantsCommand(ctx))
```
timburks, line 29: Let's spell this "conformance"

cmd/registry/cmd/compute/conformance.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/apigee/registry/cmd/registry/cmd/compute/conformance"
+	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/names"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/proto"
+)
+
+func conformantsCommand(ctx context.Context) *cobra.Command {
```
timburks, line 35: "conformance" (noted above)

cmd/registry/cmd/compute/conformance.go
```diff
@@ -0,0 +1,208 @@
...
+				log.Fatalf("%s", err.Error())
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&linter, "linter", "", "The linter to use (aip|spectral|gnostic)")
+	return cmd
+}
+
+type computeConformantTask struct {
```
timburks, line 146: nit: "Conformance"

cmd/registry/cmd/compute/conformance/lint_factory.go
```diff
@@ -0,0 +1,17 @@
...
+	"errors"
+	"fmt"
+)
+
+// CreateLinter returns a Linter object when provided the name of a linter
+func CreateLinter(linter_name string) (Linter, error) {
+	if linter_name == "spectral" {
+		return SpectralLinter{Rules: make(map[string][]string)}, nil
+	} else if linter_name == "api-linter" {
+		return SpectralLinter{Rules: make(map[string][]string)}, nil
```
timburks, line 13: Should this be returning a different linter?

timburks (approved):
This is intermediate work and I expect that some names and their meanings will change, so I'll just point out to observers that "plugin" as used here isn't the kind of plugin that we're aiming for in the final product - there we expect plugins to be separate binaries that are invoked by the registry tool which are either standalone linters or adapters that call third-party linters like Spectral and the Google API Linter.

(approved with some minor name change requests)

cmd/registry/cmd/compute/conformance.go
```diff
@@ -0,0 +1,208 @@
...
+				log.Fatalf("%s", err.Error())
+			}
+
+			projectSegments := []string{"projects", spec.ProjectID}
+
+			linterNameToLinter := make(map[string]conformance.Linter)
+			err = core.ListArtifacts(ctx, client, projectSegments, filter, true, func(artifact *rpc.Artifact) {
+				// Only consider artifacts which have the styleguide mimetype
+				messageType, err := core.MessageTypeForMimeType(artifact.GetMimeType())
+				if err != nil {
```
landrito, line 75: These errors should be handled.

cmd/registry/cmd/compute/conformance.go
```diff
@@ -0,0 +1,208 @@
...
+			}
+			spec, err := names.ParseSpec(name)
+			if err != nil {
+				log.Fatalf("%s", err.Error())
+			}
+
+			projectSegments := []string{"projects", spec.ProjectID}
+
+			linterNameToLinter := make(map[string]conformance.Linter)
+			err = core.ListArtifacts(ctx, client, projectSegments, filter, true, func(artifact *rpc.Artifact) {
```
landrito, line 72: Remind me the behavior of this method. Does this block on the network request and the callback?

cmd/registry/cmd/compute/conformance.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+func conformantsCommand(ctx context.Context) *cobra.Command {
+	var linter string
+	cmd := &cobra.Command{
+		Use:   "conformance",
+		Short: "Compute lint results for API specs",
+		Args:  cobra.MinimumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			filter, err := cmd.Flags().GetString("filter")
```
landrito, line 42: I think this flag should be registered to the command with `cmd.Flags()`?

cmd/registry/cmd/compute/conformance.go
```diff
@@ -0,0 +1,208 @@
...
+
+				}
+			})
+
+			if err != nil {
+				log.Fatalf("%s", err.Error())
+			}
+
+			// Initialize task queue.
+			taskQueue, wait := core.WorkerPool(ctx, 16)
```
landrito, line 120: If a task returns an error, how does the error get handled?

landrito (commented):
What is the overall testing strategy for this CLI code?

timburks:
@landrito regarding testing strategy, this PR is following an emerging pattern that's also used by tests in https://github.com/apigee/registry/tree/main/cmd/registry/cmd/upload and https://github.com/apigee/registry/tree/main/cmd/registry/cmd/label (for examples). Tests for individual commands are going alongside the code for the command (in the corresponding subdirectory). See #299 for some tests of the initial style guide uploader.

landrito:
Sweet I'll get caught up on that strategy! Thanks

cmd/registry/cmd/compute/conformance.go
muhammadharis: Yes it does. It makes a single call to get the artifacts via the client, and then loops over them an applies the provided callback function to each.

cmd/registry/cmd/compute/conformance.go
muhammadharis: The worker that executes this task selects on either a successfully `Done` context or an error. In the case of an error, the task logs the message to standard error, but does not drop/penalize other tasks that are already running.

For info:
https://github.com/apigee/registry/blob/05a9904c6d039605c5bd993968f2796a9d079e46/cmd/registry/core/tasks.go#L59-L63

cmd/registry/cmd/compute/conformance.go
muhammadharis: Great catch!

cmd/registry/cmd/compute/conformance.go
muhammadharis: Yes it does. It makes a single call to get the artifacts via the client, and then loops over them an applies the provided callback function to each.

cmd/registry/cmd/compute/conformance.go
muhammadharis: Agreed. I've logged the errors where they occur, but I don't think that it would make sense to `log.Fatal` them. The scenario where I see this error not being `nil` is if a style guide is somehow malformed or the mime type has been tampered with. In this scenario, it makes sense to not penalize the other style guides which are correct, but rather log info regarding the bad style guide and continue.

PR #301, opened 2021-09-16
* Simplify the references, support only $resource.entity patterns

shrutiparabgoogle:
This will help limit the invalid patterns and prevent the controller from generating incorrect actions
  We can always add $0, $1 (dependencies) support later
* Add validation check for matching patterns and references in action strings


cmd/registry/cmd/upload/manifest.go
```diff
@@ -63,8 +64,15 @@ func manifestCommand(ctx context.Context) *cobra.Command {
 			if err != nil {
 				log.WithError(err).Fatal("Failed to read manifest")
 			}
-			manifestData, _ := proto.Marshal(manifest)
 
+			// validate the manifest
+			for _, resource := range manifest.GeneratedResources {
+				if err := controller.ValidateResourceEntry(resource); err != nil {
+					log.WithError(err).Fatal("Invalid manifest")
```
timburks, line 71: With this being "Fatal", I think we're only going to report the first error found in a manifest. Let's at least add "TODO: report all errors before returning" so that we can later add more complete error reporting.

timburks (approved):

cmd/registry/cmd/upload/manifest.go
shrutiparabgoogle: Thanks, good catch! Made a change instead of deferring it in a TODO.

Now, it will log all the errors found in the manifest and exit with a FATAL log "Manifest contains errors".

Issue #293, opened 2021-09-13
Registry Controller: Create a connection from the controller to subcommands using TaskIDs


shrutiparabgoogle:
The controller is responsible for generating multiple actions and executing each one of them. Currently, the controller does this by opening a new subprocess for each action and executing it in that subprocess. This leads to interleaved logs from different actions and makes it difficult to debug. 

We implemented a partial solution to solve this issue:  The registry tool adds a unique process ID to its's logs as a prefix. While this helps in grouping the logs into corresponding processes, we still can't tie a connection from the controller's action to the subprocess output.

This can be solved by letting the controller generate the `uuid`s and passing them as a flag to the `registry` tool. The. registry tool can then prepend the flag value to it's logs.

shrutiparabgoogle:
cc @seaneganx

PR #303, opened 2021-09-17
Support for taskID flag in registry tool

shrutiparabgoogle:
* Add taskid flag to the rootCmd
* Make the controller pass a taskID to the subCommands
   This will create a connection from the action computed by the controller to it's execution logs
*. Remove dependency on `registry exec` for executing 3p commands.
* Add a logger to handle logs generated by the 3p commands.


shrutiparabgoogle:
Fixes https://github.com/apigee/registry/issues/293

cmd/registry/cmd/root.go
```diff
@@ -30,20 +32,32 @@ import (
...
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			// Initialize global default logger with unique process identifier.
+			if len(taskID) == 0 {
+				taskID = fmt.Sprintf("[ %.8s ] ", uuid.New())
+			}
+			logger := &log.Logger{
+				Level:   log.DebugLevel,
+				Handler: text.Default,
+			}
+			log.Log = logger.WithField("uid", taskID)
```
seaneganx, line 53: Can we set this in the program's main function, or is there some issue with that approach I'm not noticing?

That seems like a more appropriate spot to set the global logger variable if all else is equal. Although we don't have any reason to call this `Command` function multiple times right now, if we do in the future then setting global state here will cause strange behavior.

cmd/registry/cmd/root.go
```diff
@@ -53,5 +67,6 @@ func Command(ctx context.Context) *cobra.Command {
 	cmd.AddCommand(upload.Command(ctx))
 	cmd.AddCommand(vocabulary.Command(ctx))
 
+	cmd.PersistentFlags().StringVar(&taskID, "task_id", "", "Assign an ID to this execution")
```
seaneganx, line 70: Can we name this something more generic like `log_id` and mention in the description that it gets attached to the program's logs? Right now it seems unclear what assigning an ID actually does and may confuse users.

cmd/registry/controller/local-tasks.go
```diff
@@ -47,16 +65,38 @@ func (task *ExecCommandTask) Run(ctx context.Context) error {
...
 	}
 
-	cmd := exec.Command("registry", strings.Fields(task.Action.Command)...)
-	cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr
-	if err := cmd.Run(); err != nil {
-		logger.WithError(err).Debug("Failed Execution: failed running command")
-		return errors.New("failed running command")
+	// first party registry commands
+	if strings.HasPrefix(task.Action.Command, "registry") {
+		fullCmd := strings.Fields(fmt.Sprintf("%s --task_id=%s", task.Action.Command, task.TaskID))
```
seaneganx, line 75: Optional: `append(strings.Fields(task.Action.Command), fmt.Sprintf("--task_id=%s", task.TaskID))` feels a bit clearer to me.

cmd/registry/controller/local-tasks.go
```diff
@@ -29,6 +29,24 @@ import (
...
 
+// Implement io.Writer interface https://pkg.go.dev/io#Writer
+type LogWriter struct {
+	logger *log.Entry
+}
+
+func (w LogWriter) Write(p []byte) (n int, err error) {
+	// This is required to attach logger info to each line in multiline logs
+	for _, s := range strings.Split(strings.TrimRight(string(p), "\n"), "\n") {
+		w.logger.Debug(s)
```
seaneganx, line 40: Do we know that this method is only called with complete lines? I'm concerned we might end up writing partial/incomplete lines with this approach.

cmd/registry/controller/local-tasks.go
```diff
@@ -47,16 +65,38 @@ func (task *ExecCommandTask) Run(ctx context.Context) error {
...
+
+		if err := cmd.Run(); err != nil {
+			logger.WithError(err).Debug("Failed Execution: failed running command")
+			return errors.New("failed running command")
+		}
+	} else { //third party commands
+		fullCmd := strings.Fields(task.Action.Command)
+		cmdLogger := &LogWriter{
+			logger: log.WithFields(log.Fields{
+				"uid": fmt.Sprintf("{%s}", task.TaskID),
```
seaneganx, line 88: Are we wrapping this for backward compatibility with the monitoring dashboard log matchers? If so we should explain that in a comment. If not, is there any other reason why we can't just use the unwrapped ID here (and on line 65)?

cmd/registry/cmd/root.go
shrutiparabgoogle: This is done here because we are setting the `uid` through a rootCmd flag `log_id`. The flags are parsed at runtime ([in the execute func](https://github.com/spf13/cobra/blob/de187e874d1ca382320088f8f6d76333408e5c2e/command.go#L769)). They are not available in `main.go` when we create just the command instance `cmd := cmd.Command(ctx)`

Hence, I moved the logic to `PersistentPreRun` when we have the `log_id` value available.

cmd/registry/cmd/root.go
shrutiparabgoogle: Done

cmd/registry/controller/local-tasks.go
shrutiparabgoogle: Done

cmd/registry/controller/local-tasks.go
seaneganx: Using the logger on each call to `Write` modifies the output by splitting it into multiple log entries, so the same problem still exists without the newline splitting. I think we can either 1) write 3p command output to a buffer and print the entire buffer as one log entry, or 2) modify the LogWriter to buffer output until it sees a newline/EOF, writing the log only when a full line of output is received in the buffer.

If you've tested your existing solution and it doesn't have any issues, then I don't think we need to address this issue in this PR. It will be easy enough to follow up with a fix at a later time.

seaneganx (approved):

cmd/registry/controller/local-tasks.go
```diff
@@ -29,6 +29,21 @@ import (
...
+
+func (w LogWriter) Write(p []byte) (n int, err error) {
+	w.logger.Debug(string(p))
+	return len(p), nil
+}
+
+func NewCmdLogWriter(l *log.Entry) *LogWriter {
+	lw := &LogWriter{logger: l}
+	return lw
+}
```
seaneganx, line 45: This seems to be unused.

cmd/registry/controller/local-tasks.go
```diff
@@ -29,6 +29,21 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// Implement io.Writer interface https://pkg.go.dev/io#Writer
+type LogWriter struct {
```
seaneganx, line 33: This seems like it can be unexported.

cmd/registry/controller/local-tasks.go
shrutiparabgoogle: I didn't find any issues with my current testing, but I agree that what you are pointing out is possible when we execute long running actions. I am leaning towards option 2) from your suggestion. Currently, we are waiting for the command to finish executing, however, eventually we would want to move away from that when we start supporting long running actions. I will create a bug to revisit this later. (https://github.com/apigee/registry/issues/310)

cmd/registry/controller/local-tasks.go
shrutiparabgoogle: Done

cmd/registry/controller/local-tasks.go
shrutiparabgoogle: Done

PR #307, opened 2021-09-22
Create Conformance Report Proto

muhammadharis:
## Objective
Recently, there has been a new addition that to the API registry called API Style Guides. They allow users to declaratively set guidelines, rules, and linters to govern the static analysis on their API specs. Once a style guide is created for an API spec, there should be some method of determining how well that spec _conforms_ to that style guide. For this reason, this pull request creates the conformance report proto. It reports how well a spec conforms to any given API Style Guide.

## Future work
This proto is the first step in the ultimate goal, which is to allow users to see their specs' conformance reports after creating and API style guide for them. In subsequent pieces of work, there will be an implementation of a command that computes conformance reports for a given spec.


google/cloud/apigee/registry/applications/v1alpha1/registry_conformance_report.proto
```diff
@@ -0,0 +1,96 @@
...
+    string rule_name = 1 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // Resource name of the spec that the rule was breached on.
+    string spec_name = 2 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // URI of the Rule object on the style guide.
```
timburks, line 56: As we discussed in-person, let's remove this field.

muhammadharis: Done!

google/cloud/apigee/registry/applications/v1alpha1/registry_conformance_report.proto
```diff
@@ -0,0 +1,91 @@
...
+    ];
+
+    // A list of guideline reports.
+    repeated GuidelineReport guideline_reports = 2 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+}
+
+// RuleReportGroup is an abstraction that maps severity
+// (ERROR WARNING, INFO, HINT) to a list of rule reports for
```
landrito, line 79: Question probably orthogonal to this PR: does it make sense to make severity an enumeration? What if the user wants more granularity than that?

timburks: We've discussed adding a weight field (a float?) that could be used to compute scores based on report results. I think what we have here is good because it is simple and that the weight field (when added) might address what you have in mind.

timburks (approved):

google/cloud/apigee/registry/applications/v1alpha1/registry_conformance_report.proto
muhammadharis: This is an interesting point. Let's consider the alternative approach, which is to make severity a string that the user specifies. I agree, doing something like this would definitely allow more flexibility in specifying granularity of severities. This is good because it can allow us to differentiate between, say, a syntax error and semantic error. However, here are some things to consider if we opt for such an approach:

- The conventions of typical linters usually define a set of severity levels. Spectral defines[ 4 severity levels](https://meta.stoplight.io/docs/spectral/ZG9jOjI1MTg5-custom-rulesets#severity) which were the inspiration for this proto definition. AIPs, which is what inspired the style guide definition makes use of "must", "should", and "may" (observed in this [guideline](https://google.aip.dev/140)), which correspond to "error", "warning", and "info/suggestion". If we want to follow the convention, then it makes sense for us to also define some set levels here.

- There would need to be some method of users ranking the importance of the rules they create. In our case, it is clear that ERROR > WARNING > INFO > HINT. However, how is this ranking system going to be preserved if we allow users to create arbitrary rules? There needs to be a mechanism that we'd have to think of for this.

- Can lead to inconsistency and disorganization in style guides. If I create two different style guides, and in one I refer to errors as "error" and in another style guide someone refers to it as "fail", it may create inconsistency and confusion to someone who is unaware that they're actually the same thing when we fomulate our conformance reports.

- Without specific severities to categorize lint results in, new users who are adding rules to an existing style guide may find it difficult to know the exact category to add errors in. This would perhaps require additional documentation for users to know the exact severity, whereas our categories in the enum are well documented and clear.

All in all, in light of us following the conventions of AIPs, I think it makes sense to leave this field as an enumeration of types for now, but this is a really good idea for an extension. Perhaps an alternative solution could be that we can create a way in the future to further break down errors into various error types, or warnings into various warning types, etc. (which the user provides). I will take note of this as an extension and consider this problem more for sure.

landrito: These are great points. Thanks for thinking it through so thoroughly.

PR #309, opened 2021-09-22
Add support for dry_run flag in resolve command

shrutiparabgoogle:


cmd/registry/cmd/resolve/resolve.go
```diff
@@ -101,5 +111,7 @@ func Command(ctx context.Context) *cobra.Command {
 			}
 		},
 	}
+
+	cmd.Flags().BoolVar(&dryRun, "dry_run", false, "if set, actions will only be printed and not executed")
```
timburks, line 115: I think the flag should be "--dry-run" (dash separator and not underscore) since both git and kubectl use this name.

timburks: I notice we have a few existing flags that use underscores instead of dashes, but of the tools that I've checked (docker, kubectl, git, clang, javac), dashes are consistently used as separators (so I think we should update preexisting flags to follow this)

timburks (approved):

cmd/registry/cmd/resolve/resolve.go
shrutiparabgoogle: Done

PR #315, opened 2021-09-28
Refactor 'name' field to ID on style guide and manifest proto definition

muhammadharis:
## Summary
In the registry, the convention that we use is that `name` fields are typically given to resource names. A resource name is one that matches: `projects/{project_id}/locations/{location}/apis/{api}/versions/{version}/specs/{spec}/artifacts/{artifact}`.

However, in the case of manifests and style guides, the `name` field is actually being used as an identifier rather than an actual resource name. Thus, we should correct our terminology and call them what they really are, `id`s.

## Test
Existing unit tests should pass. There is no extra manual test that needs to be done, since all of the tests pertaining to this are automated.

### DIFF
--- a/cmd/registry/cmd/upload/manifest.go
+++ b/cmd/registry/cmd/upload/manifest.go
@@ -85,7 +85,7 @@ func manifestCommand(ctx context.Context) *cobra.Command {
 			}
 
 			artifact := &rpc.Artifact{
-				Name:     "projects/" + projectID + "/locations/global/artifacts/" + manifest.Id,
+				Name:     "projects/" + projectID + "/locations/global/artifacts/" + manifest.GetId(),
 				MimeType: core.MimeTypeForMessageType("google.cloud.apigee.registry.applications.v1alpha1.Manifest"),
 				Contents: manifestData,
 			}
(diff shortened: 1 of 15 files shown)


shrutiparabgoogle (commented):
Look good.

Can you update the manifest defined here as well? https://github.com/apigee/registry/blob/main/tests/controller/testdata/manifest.yaml

This doesn't show up in the tests, but is used as part of an end-to-end walkthrough

muhammadharis:
> Look good.
> 
> Can you update the manifest defined here as well? https://github.com/apigee/registry/blob/main/tests/controller/testdata/manifest.yaml
> 
> This doesn't show up in the tests, but is used as part of an end-to-end walkthrough

Done!

cmd/registry/cmd/upload/manifest.go
```diff
@@ -85,7 +85,7 @@ func manifestCommand(ctx context.Context) *cobra.Command {
 			}
 
 			artifact := &rpc.Artifact{
-				Name:     "projects/" + projectID + "/locations/global/artifacts/" + manifest.Name,
+				Name:     "projects/" + projectID + "/locations/global/artifacts/" + manifest.Id,
```
shrutiparabgoogle, line 88: Since we are making a change here, can we use `manifest.GetId()`? That seems like the right approach to get the field, the way it is done in `style_guide.go`

shrutiparabgoogle (approved):

landrito (approved):

timburks:
@muhammadharis thank you!

PR #316, opened 2021-09-28
Include required codegen files in Go module

### LINKED ISSUE #314: Use Go modules for dependency installation
Some of our code dependencies are generated and installed using scripts that circumvent the Go ecosystem. This issue proposes we include all of our dependencies in the module so it is fully functional after installing it using `go install`.

Notably the `rpc` and `gapic` packages in the root of this repository are not installed when the module declared by this repo is installed. The `make protos` rule runs two scripts: one installs the latest version of the [api-common-protos](https://github.com/googleapis/api-common-protos) that our local protos depend on, and the next uses [protoc and gapic](https://github.com/apigee/registry/blob/2f1c7797bd3a8a6bf1126801b79a09a5da256303/tools/COMPILE-PROTOS.sh#L24) to generate the `rpc` and `gapic` Go packages from our local protos. The tools used by these scripts are not version pinned, making them nondeterministic for any snapshot of our repo and protos.

We have encountered issues using tools like [golangci-lint](https://github.com/golangci/golangci-lint) which expect Go modules to be functionally complete when installed. In this case we try to set up the entire repo before running the linter, but it results in noisy logs when the linter tries to install everything once more.

Including our generated code in the Go module will make installation of the server and related tools easier by eliminating the additional build steps. This should prevent some integration issues with other tools in the Go ecosystem. It will also remove the nondeterminism introduced by the generation scripts, stabilizing snapshots of our repository.

seaneganx:
Fixes #314.


tools/GENERATE-OPENAPI.sh
```diff
@@ -15,23 +15,19 @@
...
-ANNOTATIONS="third_party/api-common-protos"
+if [ ! -d "third_party/api-common-protos" ]
+then
+  git clone https://github.com/googleapis/api-common-protos third_party/api-common-protos
+fi
 
-PROTOS=( \
-	google/cloud/apigee/registry/v1/registry_models.proto \
-	google/cloud/apigee/registry/v1/registry_service.proto \
+SERVICE_PROTOS=(
```
timburks, line 25: I agree with splitting these generation steps into separate scripts, but notice that the declarations (SERVICE_PROTOS, ALL_PROTOS) are repeated across multiple files. Since we seem to require that these scripts be run from the repo root, I think we could move these declarations into a common file ("tools/PROTOS.sh"?) and include that from the scripts - could you make this change to reduce duplication and the possibility of accidental inconsistency?

timburks (approved):

Issue #319, opened 2021-09-28
Use structured logging in the registry server


seaneganx:
The registry server currently uses the standard logger to print unstructured log messages. Using structured logging we can provide logs in a format that's easier to search. Output format should be configurable so different environments can be configured to print human readable text or machine readable JSON.

Issue #318, opened 2021-09-28
Expand logging in the registry server


seaneganx:
The registry server currently logs some basic information about which methods are called and whether an error occurs in response. This issue proposes additional information to include in existing logs, and additional logs to generate.

Request scoped information:
- Unique identifiers for the life of a request.
- Identifiers provided from clients to link the logs in a series of requests.
- Names of resources being accessed.

Entry scoped information:
- Log severity, e.g. error, warning, info, debug.

Additional log generation:
- Database interactions with query and connection info.
- Pub/Sub interactions with message and connection info.

PR #321, opened 2021-09-30
Structured logging for the registry server

seaneganx:
This PR addresses most features of #318 and #319. 

Notably this does not make the log output format configurable which means Cloud Logging may not even recognize that the logs are structured. It also doesn't include any logging for database interactions in the server. These features are important to add before closing those issues, but this PR stands on its own without them.

### DIFF
--- a/server/actions_apis.go
+++ b/server/actions_apis.go
@@ -73,7 +73,7 @@ func (s *RegistryServer) CreateApi(ctx context.Context, req *rpc.CreateApiReques
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	_ = s.notify(ctx, rpc.Notification_CREATED, name.String())
+	s.notify(ctx, rpc.Notification_CREATED, name.String())
 	return message, nil
 }
 
@@ -99,7 +99,7 @@ func (s *RegistryServer) DeleteApi(ctx context.Context, req *rpc.DeleteApiReques
 		return nil, err
 	}
 
-	_ = s.notify(ctx, rpc.Notification_DELETED, name.String())
+	s.notify(ctx, rpc.Notification_DELETED, name.String())
 	return &emptypb.Empty{}, nil
 }
 
@@ -211,6 +211,6 @@ func (s *RegistryServer) UpdateApi(ctx context.Context, req *rpc.UpdateApiReques
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	_ = s.notify(ctx, rpc.Notification_UPDATED, name.String())
+	s.notify(ctx, rpc.Notification_UPDATED, name.String())
 	return message, nil
 }
--- a/server/internal/storage/gorm/client.go
+++ b/server/internal/storage/gorm/client.go
@@ -17,10 +17,10 @@ package gorm
 import (
 	"context"
 	"fmt"
-	"log"
 	"sync"
 
 	_ "github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/postgres"
+	"github.com/apex/log"
 	"github.com/apigee/registry/server/internal/storage/models"
 	"gorm.io/driver/postgres"
 	"gorm.io/driver/sqlite"
@@ -169,7 +169,7 @@ func (c *Client) Put(ctx context.Context, k *Key, v interface{}) (*Key, error) {
 		if rowsAffected == 0 {
 			err := tx.Create(v).Error
 			if err != nil {
-				log.Printf("CREATE ERROR %s", err.Error())
+				log.WithError(err).Infof("CREATE ERROR")
 			}
 		}
 		return nil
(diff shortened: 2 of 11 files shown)


seaneganx:
Some example log output. You can see how this makes it easier to reason with the logs when requests are processed concurrently.
```
2021/09/29 18:15:50  info Handling request.         method=GetArtifact request_id=8a0334a4 resource=projects/test/locations/global/artifacts/sample
2021/09/29 18:15:50  info Success.                  duration=1.105716ms method=GetArtifact request_id=8a0334a4 resource=projects/test/locations/global/artifacts/sample status_code=OK
2021/09/29 18:15:50  info Handling request.         method=GetArtifactContents request_id=5346b51a resource=projects/test/locations/global/artifacts/sample
2021/09/29 18:15:50  info Success.                  duration=1.273446ms method=GetArtifactContents request_id=5346b51a resource=projects/test/locations/global/artifacts/sample status_code=OK
2021/09/29 18:15:50  info Handling request.         collection=projects/test/locations/global/apis/sample method=CreateArtifact request_id=35792608
2021/09/29 18:15:50  info Handling request.         method=DeleteProject request_id=103d99df resource=projects/filters
2021/09/29 18:15:50  info Success.                  collection=projects/test/locations/global/apis/sample duration=4.24436ms method=CreateArtifact request_id=35792608 resource=projects/test/locations/global/apis/sample/artifacts/sample status_code=OK
2021/09/29 18:15:50  info User error.               duration=3.337854ms error=project "projects/filters" not found in database method=DeleteProject request_id=103d99df resource=projects/filters status_code=NotFound
```

server/notifications.go
```diff
@@ -16,63 +16,65 @@ package server
...
+	logger := log.FromContext(ctx)
+	if !s.notifyEnabled {
+		return
+	}
 
-func (s *RegistryServer) notify(ctx context.Context, change rpc.Notification_Change, resource string) error {
-	if !s.notifyEnabled || s.projectID == "" {
-		return nil
+	if s.projectID == "" {
+		logger.Warn("Notifications are enabled but project ID is not set. Skipping notification.")
```
timburks, line 38: Does this warn only once? If not, I think this could be very noisy.

server/internal/storage/gorm/client.go
```diff
@@ -167,10 +166,7 @@ func (c *Client) Put(ctx context.Context, k *Key, v interface{}) (*Key, error) {
 		// Update all fields from model: https://gorm.io/docs/update.html#Update-Selected-Fields
 		rowsAffected := tx.Model(v).Select("*").Where("key = ?", k.Name).Updates(v).RowsAffected
 		if rowsAffected == 0 {
-			err := tx.Create(v).Error
-			if err != nil {
-				log.Printf("CREATE ERROR %s", err.Error())
-			}
+			tx.Create(v)
```
timburks, line 169: Is ignoring errors ok here because this will just fail further on?

server/actions_apis.go
```diff
@@ -73,7 +73,7 @@ func (s *RegistryServer) CreateApi(ctx context.Context, req *rpc.CreateApiReques
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	_ = s.notify(ctx, rpc.Notification_CREATED, name.String())
+	s.notify(ctx, rpc.Notification_CREATED, name.String())
```
timburks, line 76: Is this linter-related churn? I thought I saw some assignments to "_" show up when the linter CI was added.

timburks (commented):
Looks great - I just have some maintenance-related questions

server/internal/storage/gorm/client.go
seaneganx: Functionally speaking we always have ignored these errors. You can see that this method has an error return value, but that error is always `nil` whether or not the operation succeeds. This PR simply removes the log for consistency because we don't log any other database errors.

Addressing how we _handle_ database errors seems like it's outside the scope of this PR. I would like to see that happen before we add logging to this package because it may affect whether we need to log errors in this package or let them get caught by the logger at a higher level (e.g. the calling function or the server log interceptor). I created #322 to handle this.

server/notifications.go
seaneganx: It will be noisy if the issue arises, but I think that's appropriate given the circumstance. Each warning log would correspond to one instance of unexpected behavior in the server, where the server is configured to send a notification but chooses not to try because it's missing the required project ID.

Currently though, our config validation should prevent the server from starting if PubSub is enabled without a project ID. This should only execute if we accidentally sneak past that validation in the future.

server/actions_apis.go
seaneganx: The linter enforces that we explicitly discard errors if they're returned, so you're right that I added those recently. I updated the method to not return an error at all, which is why you're seeing the explicit discard removed from every call. It's not really related to logging so maybe it belongs in a different PR, but it was a trivial and non-functional change.

My assumption is that we don't want users to receive something like an internal error when something goes wrong publishing a notification, so we can just log the errors without returning them. I added those logs in this PR.

timburks: Agreed - thanks for the explanation!

server/internal/storage/gorm/client.go
timburks: SGTM

server/notifications.go
timburks: Great, that initial config validation is what I would have suggested (but we already have it).

timburks (approved):

PR #324, opened 2021-10-01
Add README for controller

shrutiparabgoogle:


tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+    dependencies:
+      - pattern: "$resource.spec"
+    action: "registry compute complexity $resource.spec"
+```
+
+#### Run the controller
+There are three ways to run the controller:
+
+##### Dry-run mode:
+In this mode, the controller will generate the actions, which it calculates, are needed to bring registry up-to-date. However, it will not execute any of this actions. This is a way to try and see what the controller is doing without making any modifications to the registry. The controller caan be invoked by using the `registry resolve` command:
```
seaneganx, line 83: >The controller caan be invoked

Typo.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+Deploy the controller job which will use the `registry-linters` image and run the `registry resolve` command, every 5 mins
+```shell
+export REGISTRY_MANIFEST_ID=projects/demo/locations/global/artifacts/test-manifest
+
+make deploy-controller-job
+```
+
+###### Step 3: Verification
+* Once the job is deployed, you should see that in every iteration it is reading the manifest and calculating the actions. 
+* Since, we already have the artifacts generated fom our previous commands, we need to clean up the previously generated artifacts.
```
seaneganx, line 148: >artifacts generated fom our previous commands

Typo.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+
+make deploy-controller-job
+```
+
+###### Step 3: Verification
+* Once the job is deployed, you should see that in every iteration it is reading the manifest and calculating the actions. 
+* Since, we already have the artifacts generated fom our previous commands, we need to clean up the previously generated artifacts. 
+`registry delete projects/demo/locations/global/apis/-/versions/-/specs/-/artifacts/-`.
+Once the artifacts are cleaned up, we can see the controller in action.
+* Use list command to check what artifacts are generated by the controller, the behavior should be same as the one. described in the standalone case.
```
seaneganx, line 151: >the behavior should be same as the one. described

I think this should say "be the same" and the period after "one" seems incorrect.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+  source auth/GKE.sh
+
+  # Create demo project
+  apg registry create-project --project_id demo --json
+
+  # Create 10 versions of the petstore API
+  ./tests/controller/create_apis.sh 10
+  ``` 
+
+#### Upload the "desired state" definition
```
seaneganx, line 51: IMO the header here should be "Upload the manifest" with any additional explanation in the section's body. This seems like an attempt to simplify the documentation but I think it has the opposite effect here because readers don't necessarily know that "desired state definition" and "manifest" are the same thing.

If we really do want to explain again that the manifest represents a "desired state" we can link back to the first explanation. Maybe the document should have a "definitions" section explaining controller specific terminology.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+        - pattern: $resource.spec
+          filter: "mime_type.contains('openapi')"
+      action: "registry compute lint $resource.spec --linter spectral"
+  ```
+  The above manifest defines that for every spec (`pattern: $resource.spec`) of type openapi (`filter: "mime_type.contains('openapi')"`) in the registry, there should exist a corresponding lint-spectral artifact (`pattern: apis/-/versions/-/specs/-/artifacts/lint-spectral`). 
+  If the target artifact is not present or out-of-date, then it can be re-generated using the specified action (`action: "registry compute lint $resource.spec --linter spectral"`).
+
+* The controller will compare the current and the supplied desired state of the registry. For resources which are non-existing or outdated, the controller will execute the action supplied in the manifest.
+
+### How to use it?
```
seaneganx, line 23: Consider a header that isn't a question. For example "Using the Controller" states the section's purpose. Most headers except the first few seem to follow that style and I think it makes it easier to search the doc for specific topics.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+
+With this basic definition in mind, you can supply different configurations to the controller to generate various artifacts in the registry.
+
+#### Setup the registry server:
```
seaneganx, line 27: Consider removing the `:` from the headers. The formatting of headers makes it unnecessary and it doesn't seem to be consistently included.

seaneganx (commented):
Looks good. I added a bunch of questions and suggestions that I thought of while reading the document and considering various situations such as being a new user, or looking for specific info as a current user.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
-This folder contains steps to deploy the controller framework with an example manifest.yaml
\ No newline at end of file
+##What is a controller?
+
+In simple terms, controller is a framework to keep the registry up-to-date. The idea for this framework is inspired from the [Kubernetes controller](https://kubernetes.io/docs/concepts/architecture/controller/) pattern. Controller can be used to enrich the API data stored in the registry, with additional analytical artifacts. 
+
+### How does it work?
+* A Controller will take in configuration (aka manifest) which describes the desired state of the registry. The desired state in this context is essentially dependency relations between different entities in the database.
```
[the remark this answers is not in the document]
timburks: We've discussed having different controller instances watching different manifests (one system manifest, one or more user-specified ones)

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+  id: "example-manifest"
+  generated_resources:
+    - pattern: apis/-/versions/-/specs/-/artifacts/lint-spectral
+      dependencies:
+        - pattern: $resource.spec
+          filter: "mime_type.contains('openapi')"
+      action: "registry compute lint $resource.spec --linter spectral"
+  ```
+  The above manifest defines that for every spec (`pattern: $resource.spec`) of type openapi (`filter: "mime_type.contains('openapi')"`) in the registry, there should exist a corresponding lint-spectral artifact (`pattern: apis/-/versions/-/specs/-/artifacts/lint-spectral`). 
+  If the target artifact is not present or out-of-date, then it can be re-generated using the specified action (`action: "registry compute lint $resource.spec --linter spectral"`).
```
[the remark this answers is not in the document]
timburks: I think this is point-of-view -- from the point of view of the controller, the artifact can be regenerated with the specified command. From the user's point of view, this "would" be generated... or maybe "should" is a safer word to use here since we expect it to happen if there are no errors.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+* The controller will compare the current and the supplied desired state of the registry. For resources which are non-existing or outdated, the controller will execute the action supplied in the manifest.
+
+### How to use it?
+
+With this basic definition in mind, you can supply different configurations to the controller to generate various artifacts in the registry.
+
+#### Setup the registry server:
+Make sure you have a registry server running. We will deploy the server to GKE for the purpose of this walkthrough. 
+
+- Make sure you have the right config set in [config/registry-server.yaml](https://github.com/apigee/registry/blob/main/config/registry-server.yaml).
```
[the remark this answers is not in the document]
timburks: +1

timburks (approved):

tests/controller/README.md
```diff
@@ -1 +1,209 @@
-This folder contains steps to deploy the controller framework with an example manifest.yaml
\ No newline at end of file
+##What is a controller?
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

tests/controller/README.md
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

tests/controller/README.md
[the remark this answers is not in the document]
shrutiparabgoogle: Yes, Controller is just a framework, it is possible to deploy multiple instances which can each watch a different `manifest`.

tests/controller/README.md
shrutiparabgoogle: Done.

tests/controller/README.md
shrutiparabgoogle: Done.

tests/controller/README.md
```diff
@@ -1 +1,209 @@
...
+
+In simple terms, controller is a framework to keep the registry up-to-date. The idea for this framework is inspired from the [Kubernetes controller](https://kubernetes.io/docs/concepts/architecture/controller/) pattern. Controller can be used to enrich the API data stored in the registry, with additional analytical artifacts. 
+
+### How does it work?
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

tests/controller/README.md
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

tests/controller/README.md
shrutiparabgoogle: Done.

tests/controller/README.md
shrutiparabgoogle: Done.

tests/controller/README.md
shrutiparabgoogle: Done.

seaneganx (approved):

PR #340, opened 2021-10-09
Structural changes to allow multiple gRPC service handlers to be run on the same gRPC server.

timburks:
These changes allow additional gRPC service handlers to be installed alongside the RegistryServer instance. That makes it possible for the registry-server executable to support multiple gRPC services, each with its own service handler concurrently serving on the same gRPC interface. Additional gRPC services might implement capabilities including API search and database management.


timburks:
@seaneganx I picked "server" and later "servers" for this directory because it contained the implementation of the [RegistryServer](https://github.com/apigee/registry/blob/c424b49cdf4a05966b60d7a954c9abd3d79cf989/rpc/registry_service_grpc.pb.go#L420) interface that was generated from the protos - [service Registry](https://github.com/apigee/registry/blob/c424b49cdf4a05966b60d7a954c9abd3d79cf989/google/cloud/apigeeregistry/v1/registry_service.proto#L34) and any other future services ("Search", "Admin", etc) would each have a generated interface and a backing implementation.

Looking at this following your feedback and our conversation, it seems that naming the directory "services" instead of "servers" would more clearly indicate that these are service implementations. I'm less sure about making the directory name singular because (I think) we want to include multiple services here and wouldn't want to suggest that there is just one service (also, I notice that our project already includes directories named "containers", "examples", "tests", "tools", and ".github/workflows".

WDYT?

seaneganx:
>Looking at this following your feedback and our conversation, it seems that naming the directory "services" instead of "servers" would more clearly indicate that these are service implementations.

Changing from server(s) to service(s) sounds good to me. Whether it's plural is a different discussion. 

>I'm less sure about making the directory name singular because (I think) we want to include multiple services here and wouldn't want to suggest that there is just one service

When I'm organizing my documents and non-code files on my personal workstation I use plural because it reads better when I'm browsing in a file explorer. In that case I want to see my _Documents_ because I'm using directories like I would use physical folders in a filing cabinet. This is how I think you're making a naming decision.

In code I'm always importing a singular thing, like a Registry _server_, a base64 _encoding_, or a heap _container_. I don't think importing from `encoding/base64` or `container/heap` suggests that `encoding/hex` and `container/list` wouldn't exist. The [stdlib](https://pkg.go.dev/std) uses singular form whenever possible, with the only exceptions being packages that would otherwise use reserved keywords, e.g. `bytes`, `errors`, `strings`. 

>(also, I notice that our project already includes directories named "containers", "examples", "tests", "tools", and ".github/workflows".

Right, our repository is mixed. Other directories exist in singular form like `cmd` even though we have multiple commands, and `config` even though it contained several configs. I don't think we want to change these names because they aren't matching the quantity of things in the directory.

>WDYT?

We can go either way, and I don't think having mixed style across the repo is something to avoid. We can keep Go package names optimized for their usage in code by following the singular naming convention, and also use plurals when appropriate in other cases like "examples" or "containers."

seaneganx:
>Do you think these might be a result of our Go version change? I generated these files with go mod tidy using Go 1.17.2. 

I just tested `go mod tidy` against our main branch using Go 1.17.1 and found nothing needed to be updated. Just in case something changed on the minor version bump, I tried again after upgrading to 1.17.2 and again found no change.

Checking out your PR and running `go mod tidy` I find it's actually removing all the dependencies that were added. Below you can see my steps to reproduce (go.sum output omitted, but it seems to remove all the added entries.

```
$ gh pr checkout 340
From github.com:apigee/registry
 * [new ref]         refs/pull/340/head -> multiple-servers
Switched to branch 'multiple-servers'
$ go mod tidy
$ git --no-pager diff
diff --git a/go.mod b/go.mod
index 695f394..be3b47f 100644
--- a/go.mod
+++ b/go.mod
@@ -53,7 +53,7 @@ require (
 	github.com/go-openapi/jsonpointer v0.19.5 // indirect
 	github.com/go-openapi/swag v0.19.15 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
-	github.com/golang/protobuf v1.5.2
+	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/jackc/chunkreader/v2 v2.0.1 // indirect
@@ -73,7 +73,6 @@ require (
 	github.com/mschoch/smat v0.2.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/spf13/viper v1.8.1
 	github.com/steveyen/gtreap v0.1.0 // indirect
 	github.com/stoewer/go-strcase v1.2.0 // indirect
 	github.com/stretchr/objx v0.2.0 // indirect
@@ -92,17 +91,4 @@ require (
 	google.golang.org/appengine v1.6.7 // indirect
 )

-require (
-	github.com/fsnotify/fsnotify v1.4.9 // indirect
-	github.com/hashicorp/hcl v1.0.0 // indirect
-	github.com/magiconair/properties v1.8.5 // indirect
-	github.com/mitchellh/mapstructure v1.4.1 // indirect
-	github.com/pelletier/go-toml v1.9.3 // indirect
-	github.com/spf13/afero v1.6.0 // indirect
-	github.com/spf13/cast v1.3.1 // indirect
-	github.com/spf13/jwalterweatherman v1.1.0 // indirect
-	github.com/subosito/gotenv v1.2.0 // indirect
-	gopkg.in/ini.v1 v1.62.0 // indirect
-)
-
```

Running `git diff main go.mod go.sum` afterward also verifies that my run of `go mod tidy` is undoing everything the previous commit(s) on the branch added because it returns no difference. Can you run `go mod tidy` again on this branch and confirm the results? I'll be surprised and interested to find out why if we're seeing different results.

PR #341, opened 2021-10-09
Possible structure for a search service.

timburks:
For discussion. This requires #340 and illustrates a way to include a search service into the registry server. This structure keeps the search operations in a distinct hander that could be built into a separate binary or independently enabled/disabled in a common binary. Currently both RegistryServer and SearchServer run on the same gRPC interface, each handling their own gRPC requests.

This can be tested with the `apg` command. `apg search` accesses the search-related operations: `apg search index` and `apg search query`. The autogenerated `apg` tool uses different environment variables to configure each service (registry and search), but the additional variables needed are set in `auth/LOCAL.sh`.


timburks:
Moved to https://github.com/apigee/registry-experimental/pull/15

PR #350, opened 2021-10-14
Ensure the existence of tables on server creation only.

timburks:


timburks:
@seaneganx updated - PTAL

server/registry/internal/storage/gorm/client.go
```diff
@@ -88,15 +86,18 @@ func NewClient(ctx context.Context, driver, dsn string) (*Client, error) {
...
 	default:
 		unlock()
 		return nil, fmt.Errorf("unsupported database %s", driver)
 	}
 }
 
+// EnsureTables ensures that all necessary tables exist in the database.
+func (db *Client) EnsureTables() error {
+	return db.ensureTables()
+}
```
seaneganx, line 99: Why not export `ensureTables` directly instead of wrapping it?

server/registry/server.go
```diff
@@ -71,7 +71,15 @@ func New(config Config) *RegistryServer {
...
 
-	return s
+	db, err := storage.NewClient(context.Background(), s.database, s.dbConfig)
+	if err != nil {
+		return nil, err
+	}
+	defer db.Close()
+	if err := db.EnsureTables(); err != nil {
+		return nil, err
+	}
```
seaneganx, line 81: I still think constructors should only construct their return type - setting fields and applying options/defaults. The fact that this constructor returns an error and mutates the database indicates it's doing much more than just constructing. It would be more accurately named `NewAndInitialize` with this addition.

If we're going to bump this into a future PR, let's make sure to actually do that soon. It seems like a really easy change to split this into a separate initialize method for this PR though.

server/registry/server_test.go
```diff
@@ -43,20 +43,28 @@ func defaultTestServer(t *testing.T) *RegistryServer {
 	t.Helper()
 
 	if !usePostgres {
-		return serverWithSQLite(t)
+		if server, err := serverWithSQLite(t); err != nil {
+			t.Fatalf("Setup: failed to get server with SQLite: %s", err)
```
seaneganx, line 47: Since we're not actually handling this error, can we fail the test inside the `serverWithSQLite` function? The Postgres function returns an error so we can handle it by falling back to SQLite - here it just complicates the control flow and duplicates our fail code.

seaneganx (approved):

server/registry/internal/storage/gorm/client.go
timburks: No reason, fixed.

server/registry/server.go
timburks: I appreciate the aesthetic, and I found [someone who shares it](https://www.reddit.com/r/golang/comments/9bkql2/stop_returning_errors_in_constructors_use/) but I think the followup discussion there suggests that it's not universal or always practical. There are cases (a few, but some) where constructors in the standard library return errors, and I don't (personally) think that creating a server needs to exclude some assurance that the server being returned is valid - that's what's happening [here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/http/request.go;l=863)  and [here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/archive/zip/reader.go;l=87).

server/registry/server_test.go
timburks: Yes, this code is more verbose - but OTOH I think it would be odd for serverWithSQLite() and serverWithPostgres() to treat internal errors differently only to optimize the way that these functions are used in this instance. I'd rather err on the side of consistency here.

PR #366, opened 2021-10-26
Update demos

timburks:
Updates demo scripts:
- demo projects can be now set with the PROJECT variable
- bulk upload commands are now not expected to create projects

Updates registry tool:
- bulk upload commands no longer create projects (and have no dependency on the Admin service)
- bulk upload discovery "sanitizes" API and version names to be valid identifiers
- logging levels are adjusted so that invalid specs are warnings (for complexity, details, and vocabulary computation)
- vocabulary operation results are stored compressed and uncompressed when exported to sheets


cmd/registry/cmd/compute/complexity.go
```diff
@@ -105,7 +106,8 @@ func (task *computeComplexityTask) Run(ctx context.Context) error {
 		}
 		document, err := oas3.ParseDocument(data)
 		if err != nil {
-			return fmt.Errorf("invalid OpenAPI: %s", spec.Name)
+			log.WithError(err).Warnf("Invalid OpenAPI: %s", spec.Name)
```
seaneganx, line 109: This log seems to point out an error - something claims to be OpenAPI v3 but can't be parsed as OpenAPI v3. I think of warnings as "successful but worthy of attention" events. Can you elaborate on your choice?

+1 toward writing down guidelines for each log level at some point.

seaneganx (approved):

cmd/registry/cmd/compute/complexity.go
timburks: Thanks, that's valid - I didn't think very hard about the "Warn" level; I mainly wanted to keep these problems from causing their parent operations to fail (so that a computation could fail for one invalid spec but continue and succeed for others). I've changed these from `Warnf` to `Errorf` and the demos still seem to run as expected, so I'll merge that and we can revisit again if needed.

PR #376, opened 2021-10-29
Support for deployment resources in the names package

timburks:
This also updates the names tests to use Parse functions instead of regular expressions (now fully hiding the regular expressions inside the names package)


server/registry/names/deployment.go
```diff
@@ -0,0 +1,133 @@
...
+func deploymentCollectionRegexp() *regexp.Regexp {
+	return regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/deployments$",
+		identifier, Location, identifier))
+}
+
+// deploymentRegexp returns a regular expression that matches a deployment resource name.
+func deploymentRegexp() *regexp.Regexp {
+	return regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/deployments/%s$",
+		identifier, Location, identifier, identifier))
+}
```
seaneganx, line 104: Since these are unexported, consider moving them into package level vars similar to deployment revisions. I don't think there's any benefit to reconstructing the pattern every time we use it.

seaneganx (approved):

server/registry/names/deployment.go
timburks: Agreed! I'll make that change after the current set of PRs settle.

PR #377, opened 2021-10-29
Rename docker image target to "registry-server".

timburks:
This updates the product of the docker-publish GitHub action to match the executable name and the container definition in the [containers](https://github.com/apigee/registry/tree/main/containers) directory.

I guess this build is using the Dockerfile at the root of the repo. That's a link to this [Dockerfile](https://github.com/apigee/registry/blob/main/containers/registry-server/Dockerfile) in [containers/registry-server](https://github.com/apigee/registry/tree/main/containers). If possible, it would be good to have a second build that builds the [registry-tools](https://github.com/apigee/registry/blob/main/containers/registry-tools/Dockerfile) container. @giteshk could that be done with this action or would we need to set up another one? 

@shrutiparabgoogle if we built the `registry-tools` image in a GitHub action (see above), we could rewrite the Dockerfile for `registry-linters` to use that one and we could build `registry-linters` in a GitHub action as well.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -33,7 +33,7 @@ env:
   # Use docker.io for Docker Hub if empty
   REGISTRY: ghcr.io
   # github.repository as <account>/<repo>
-  IMAGE_NAME: ${{github.repository_owner}}/registry-backend
+  IMAGE_NAME: ${{github.repository_owner}}/registry-server
 
 
 jobs:
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -14,5 +14,5 @@
 
 steps:
 - name: 'gcr.io/cloud-builders/docker'
-  args: [ 'build', '-t', 'gcr.io/${PROJECT_ID}/registry-backend', '.']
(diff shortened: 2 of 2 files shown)


shrutiparabgoogle:
> @shrutiparabgoogle if we built the `registry-tools` image in a GitHub action (see above), we could rewrite the Dockerfile for `registry-linters` to use that one and we could build `registry-linters` in a GitHub action as well.

That would be great! @giteshk  Can we do a multi-stage build in github-actions? `registry-linters` uses `registry-tools` as it's base image, we might need to build this in two stages. Or open to any better way of doing this that I am unaware of.

Also, just curious, when we build these images through github-actions where are we pushing them? Do we have a public repo now for registry images?

timburks:
@giteshk thanks for catching that - I've changed cloudbuild.yaml.

@shrutiparabgoogle the images are under [packages](https://github.com/orgs/apigee/packages) at the top of the [main Apigee page](https://github.com/apigee) and in the "Packages" sidebar of the [main registry page](https://github.com/apigee/registry).

giteshk (approved):

PR #379, opened 2021-11-02
Add support for allow_missing in updates of projects, APIs, and versions

timburks:
This makes our support for allow_missing consistent across all major (non-artifact) resources: projects, apis, versions, specs, and deployments (specs and deployments already supported allow_missing).

(This builds on the deployments-implementation branch and will be easier to review when that is merged)


timburks:
@seaneganx with the deployments support merged, I think this is clean enough to review.

server/registry/actions_apis.go
```diff
@@ -193,8 +199,21 @@ func (s *RegistryServer) UpdateApi(ctx context.Context, req *rpc.UpdateApiReques
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	if req.GetAllowMissing() {
```
timburks, line 202: I discovered this race condition (described below) when modifying spec bulk uploads to use updates with allow_missing. Because the uploads upload specs in parallel, and because often multiple specs exist for a single API, it's common for multiple calls to update an API (allowing missing) to occur at once.

This lock is a crude way to prevent that - I think a better solution would address this at the database level - and although I don't currently imagine situations where this would occur for other resources, it seemed prudent to add this locking for all resources with updates that accept allow_missing.

seaneganx: In practice this probably covers the race conditions that are likely to occur.

Does locking updates to the entire resource type affect performance noticeably? An optimization would be to make a simple "resource locker" type that accepts a resource name to lock - allowing different APIs to update simultaneously.

seaneganx (approved):

server/registry/actions_apis.go
timburks: I didn't notice a performance impact, but my testing included some rework of the uploader that would have made it faster. I agree that this is our most likely race condition, but since this locking is in-process, when we scale horizontally (multiple servers on the same DB) we could still have this problem. We might fix that with a database-level lock (on a table?) or better yet, by streamlining our handling of this to handle both allow_missing (true/false) operations with single Postgres operations.

timburks: Noting this in #382 - thanks @seaneganx !

PR #384, opened 2021-11-03
Build a new Registry Envoy dockerfile

giteshk:
Added a container to build the Registry Envoy image
Updated the build files to reference the registry-server dockerfile from the containers folder.


deployments/envoy/envoy.yaml.tmpl
```diff
@@ -0,0 +1,82 @@
...
+admin:
+  access_log_path: /tmp/admin_access.log
+  address:
+    socket_address: { address: 0.0.0.0, port_value: 9901 }
+
+static_resources:
+  listeners:
+    - name: listener1
+      address:
+        socket_address: { address: 0.0.0.0, port_value: 9999 }
```
timburks, line 23: This is telling Envoy to serve on port 9999. This is not a blocking question, but should that be parameterized too?

.github/workflows/docker-publish.yml
```diff
@@ -33,7 +33,9 @@ env:
   # Use docker.io for Docker Hub if empty
   REGISTRY: ghcr.io
   # github.repository as <account>/<repo>
-  IMAGE_NAME: ${{github.repository_owner}}/registry-server
+  REGISTRY_SERVER: ${{github.repository_owner}}/registry-server
+  # github.repository as <account>/<repo>
+  REGISTRY_ENVOY: ${{github.repository_owner}}/registry-envoy
```
timburks, line 38: I'm not clear on why we need to build a container for Envoy instead of directly using an existing Envoy container - in Kubernetes setups, I've been able to just replace envoy.yaml with a ConfigMap. Is this to allow Envoy to be configured with environment variables instead?

giteshk: Yes that is correct. 
Additionally the envoy container needs a proto.pb (which is build in the first stage of the Dockerfile) for the envoy proxy to work.

deployments/envoy/envoy.yaml.tmpl
giteshk: Makes sense.. we should use the PORT variable. I will add this change in.

.github/workflows/docker-publish.yml
timburks: Agreed about the compiled protos - I've also added that with a ConfigMap. Let's go ahead with this and add a README later that explains how/why to use this.

timburks (approved):

.github/workflows/docker-publish.yml
timburks: (A concern that I have about this is that it is tied to a specific release of Envoy, but using this is certainly optional - people who know Envoy well can configure their own containers)

PR #385, opened 2021-11-04
Update "registry update bulk" commands to use allow_missing and hashes to reduce traffic.

timburks:


cmd/registry/cmd/upload/bulk/discovery.go
```diff
@@ -246,6 +206,18 @@ func (task *uploadDiscoveryTask) fetchDiscoveryDoc() error {
...
+	// normalize the doc to produce consistent hashes
+	var m interface{}
+	err = json.Unmarshal(bytes, &m)
+	if err != nil {
+		return err
+	}
+	normalized, err := json.MarshalIndent(m, "", "  ")
+	if err != nil {
+		return err
+	}
```
seaneganx, line 219: Consider moving this into a helper - the comment doesn't seem necessary if we name it `normalizeJSON`. It's not clear to me how this normalizes the JSON though... Even though we indent it consistently, wouldn't the fields still be out of order?

cmd/registry/cmd/upload/bulk/discovery.go
```diff
@@ -155,23 +145,37 @@ func (task *uploadDiscoveryTask) createVersion(ctx context.Context) error {
...
-	}); !core.NotFound(err) {
-		return err // Returns nil when spec is found without error.
+	}); err == nil {
+		if int(spec.GetSizeBytes()) == len(task.contents) {
+			hash := hashForBytes(task.contents)
+			if spec.GetHash() == hash {
+				log.Debugf("Matched existing spec %s", task.specName())
+				return nil // this spec is already uploaded
+			}
+		}
```
seaneganx, line 159: This block seems to be more readable when flattened.

```
// Use the spec size and hash to avoid unnecessary uploads.
spec, err := task.client.GetApiSpec(ctx, &rpc.GetApiSpecRequest{
	Name: task.specName(),
})

if err == nil && int(spec.GetSizeBytes()) == len(task.contents) && spec.GetHash() == hashForBytes(task.contents) {
	log.Debugf("Matched existing spec %s", task.specName())
	return nil // this spec is already uploaded
}
```

cmd/registry/cmd/upload/bulk/discovery.go
```diff
@@ -155,23 +145,37 @@ func (task *uploadDiscoveryTask) createVersion(ctx context.Context) error {
 	return nil
 }
 
-func (task *uploadDiscoveryTask) createSpec(ctx context.Context) error {
-	if _, err := task.client.GetApiSpec(ctx, &rpc.GetApiSpecRequest{
+func (task *uploadDiscoveryTask) createOrUpdateSpec(ctx context.Context) error {
+	// Use the spec size and hash to avoid unnecessary uploads.
```
seaneganx, line 149: Are we optimizing for bandwidth or performance? Unless we have metrics that you used to make this decision, it might actually be faster to upload without making an additional round-trip to check if we need to reupload.

This approach probably does save some amount of bandwidth, but it's unclear to me why we're optimizing for that and how much of a difference it really makes in practice.

cmd/registry/cmd/upload/bulk/discovery.go
```diff
@@ -155,23 +145,37 @@ func (task *uploadDiscoveryTask) createVersion(ctx context.Context) error {
...
+	if spec, err := task.client.GetApiSpec(ctx, &rpc.GetApiSpecRequest{
 		Name: task.specName(),
-	}); !core.NotFound(err) {
-		return err // Returns nil when spec is found without error.
+	}); err == nil {
+		if int(spec.GetSizeBytes()) == len(task.contents) {
+			hash := hashForBytes(task.contents)
+			if spec.GetHash() == hash {
+				log.Debugf("Matched existing spec %s", task.specName())
+				return nil // this spec is already uploaded
```
seaneganx, line 157: I think the log message should include a mention that we're skipping this upload _because_ it matches an existing spec. In addition to being more informative, the comment will become unnecessary.

cmd/registry/cmd/upload/bulk/openapi.go
```diff
@@ -166,78 +165,82 @@ func (task *uploadOpenAPITask) populateFields() error {
...
-		return err // Returns nil when spec is found without error.
+	}); err == nil {
+		if int(spec.GetSizeBytes()) == len(contents) {
+			hash := hashForBytes(contents)
+			if spec.GetHash() == hash {
+				log.Debugf("Matched existing spec %s", task.specName())
+				return nil // this spec is already uploaded
+			}
+		}
+	}
```
seaneganx, line 223: Does it make sense to refactor this into a helper? It seems identical to the discovery doc and proto functionality.

cmd/registry/cmd/upload/bulk/discovery.go
timburks: "The json package always orders keys when marshalling." https://stackoverflow.com/a/18668885

cmd/registry/cmd/upload/bulk/openapi.go
timburks: Maybe, I noticed a lot of similarity in the three uploaders. I think that's worth watching.

cmd/registry/cmd/upload/bulk/discovery.go
seaneganx: Right, so the use of MarshalIndent instead of Marshal is just for formatting. Thanks.

cmd/registry/cmd/upload/bulk/discovery.go
timburks: The "it might actually be faster" observation might be true for very small specs, but consider that this is both a utility and sample code - it shows that when bandwidth is a consideration, hashes can be used to reduce API traffic.

cmd/registry/cmd/upload/bulk/discovery.go
timburks: OK - updated.

cmd/registry/cmd/upload/bulk/discovery.go
timburks: Yes, the docs come indented from the Discovery service, so that seemed best.

I've added the helper that you suggested.

cmd/registry/cmd/upload/bulk/discovery.go
timburks: ok, I'll revise this

cmd/registry/cmd/upload/bulk/discovery.go
timburks: Done.

cmd/registry/cmd/upload/bulk/discovery.go
timburks: Done.

timburks:
@seaneganx Updated - PTAL - thanks!

cmd/registry/cmd/upload/bulk/openapi.go
```diff
@@ -166,112 +165,82 @@ func (task *uploadOpenAPITask) populateFields() error {
...
 	if err == nil {
-		log.Debugf("Created %s", response.Name)
-	} else if core.AlreadyExists(err) {
-		log.Debugf("Found %s", task.apiName())
+		log.Debugf("Updated %s", response.Name)
 	} else {
 		log.WithError(err).Debugf("Failed to create API %s", task.apiName())
+		// Returning this error ends all tasks, which seems appropriate to
+		// handle situations where all might fail due to a common problem
+		// (a missing project or incorrect project-id).
```
seaneganx, line 182: Does this comment belong on every error returned by tasks? Does this type of task especially depend on this behavior? From my perspective it seems like it's documentation for the task pool.

cmd/registry/cmd/upload/bulk/protos.go
```diff
@@ -117,127 +116,111 @@ func (task *uploadProtoTask) Run(ctx context.Context) error {
...
-	if err != nil && !core.NotFound(err) {
-		return err
+	}); err == nil {
+		if int(spec.GetSizeBytes()) == len(contents) {
+			hash := hashForBytes(contents)
+			if spec.GetHash() == hash {
+				log.Debugf("Matched already uploaded spec %s", task.specName())
+				return nil
+			}
+		}
```
seaneganx, line 204: Shouldn't this be removed now that the flattened check is included above?

seaneganx (approved):
LGTM - Please remove the redundant hash check before merging or clarify if I'm misunderstanding.

cmd/registry/cmd/upload/bulk/protos.go
timburks: Yes, I missed this. Thanks for pointing this out! (removed now)

cmd/registry/cmd/upload/bulk/openapi.go
timburks: That's a good point - looking back at the code, I recall putting this here in an attempt to explain why createAPI returns errors but not createVersion or createSpec (at some point, we might decide that they should too). I'll leave these in place but am open to alternatives and future changes.

PR #386, opened 2021-11-06
Share logger attributes between registry tool and server

### LINKED ISSUE #317: Share logger attributes between registry tool and server
The registry tool currently logs a unique identifier that can be included as a filter to inspect logs that resulted from one command execution. When that command makes requests to the registry server, the logs produced by the server do not include an identifier.

Since the registry tool passes its context argument with all of its requests to the server, we can include relevant logger attributes in the context and attach them to the server logs for those requests.

seaneganx:
Registry tool commands have a unique ID attached to their logs. This change allows that ID (and potentially other information) to be passed to the server. For example, here's a command that makes an API call.

```
$ registry get projects/p
  DEBUG[0000] Failed to get resource    error=rpc error: code = NotFound desc = project "projects/p" not found in database uid=5434deea
```

The server logs associated with this command can be found by searching for the matching `uid`.

```
  INFO[0004] Handling request.         method=GetProject request_id=cdf11306 resource=projects/p uid=5434deea
  DEBUG[0004] Database operation.       duration=120.276µs method=GetProject query=SELECT * FROM `projects` WHERE key = "projects/p" ORDER BY `projects`.`key` LIMIT 1 request_id=cdf11306 resource=projects/p uid=5434deea
  INFO[0004] User error.               duration=4.387131ms error=project "projects/p" not found in database method=GetProject request_id=cdf11306 resource=projects/p status_code=NotFound uid=5434deea
```

Fixes #317.

### DIFF
--- a/cmd/registry-server/main.go
+++ b/cmd/registry-server/main.go
@@ -22,7 +22,8 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/apex/log"
+	"github.com/apigee/registry/log"
+	"github.com/apigee/registry/log/interceptor"
 	"github.com/apigee/registry/rpc"
 	"github.com/apigee/registry/server/registry"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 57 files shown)


cmd/registry-server/main.go
```diff
@@ -93,30 +94,37 @@ func main() {
 	pflag.StringVarP(&configPath, "configuration", "c", "", "The server configuration file to load.")
 	pflag.Parse()
 
+	bootLogger := log.NewLogger()
```
timburks, line 97: Guessing that "bootLogger" means that this is only used during startup of the server, and that this is a separate logger because it can't be configured before it is needed.

cmd/registry/cmd/root.go
```diff
@@ -66,6 +58,5 @@ func Command(ctx context.Context) *cobra.Command {
 	cmd.AddCommand(upload.Command(ctx))
 	cmd.AddCommand(vocabulary.Command(ctx))
 
-	cmd.PersistentFlags().StringVar(&logID, "log-id", "", "Assign an ID which gets attached to the log produced")
```
timburks, line 69: Was this option fully removed? If I run the registry tool from the command line, is there a way for me to know which ID in the server logs corresponds to the run that I just made?

cmd/registry/googleauth/client.go
```diff
@@ -42,8 +42,7 @@ func GetAuthenticatedClient(ctx context.Context) (*http.Client, error) {
 	}
 	// Create configuration for specified scopes.
 	// If you modify these scopes, delete your previously saved token.
-	config, err := google.ConfigFromJSON(b,
-		"https://www.googleapis.com/auth/spreadsheets")
+	config, err := google.ConfigFromJSON(b, "https://www.googleapis.com/auth/spreadsheets")
```
timburks, line 45: Did an automatic formatter make this change?

timburks (commented):
Could we get a brief note explaining how this works? Specifically I'd like to have an explanation of the following:
1. How request IDs are being added to the request metadata sent by the registry tool (is apex doing that?)
2. What that request metadata looks like
3. How the interceptor is being used in the server to get this metadata
4. What are the requirements for writing logging code in the registry tool and server (we could say "copy the existing pattern", but I'd like for us to have more explanation of how this works, especially since there's an external dependency involved)

I think this could go well in the registry/log package as comments in code or in a README.

Thanks!

timburks (approved):
(no blocking concerns, but please see my note about documentation)

cmd/registry/cmd/root.go
seaneganx: Yeah it's fully removed. When we run a command there should always be at least one log produced by the command that can identify the ID for users. If there are any exceptions I think we should add some logging to those so users know something is happening.

In general I think it would be helpful to add at least one general-purpose log "hook" like the inceptor we use in the server. It could simply be a post-run "Success" or "Fail" with some metrics like duration attached, similar to the server.

cmd/registry/googleauth/client.go
seaneganx: This shouldn't have been included in the PR.

seaneganx:
>Could we get a brief note explaining how this works? Specifically I'd like to have an explanation of the following:
>1. How request IDs are being added to the request metadata sent by the registry tool (is apex doing that?)
>2. What that request metadata looks like
>3. How the interceptor is being used in the server to get this metadata
>4. What are the requirements for writing logging code in the registry tool and server (we could say "copy the existing pattern", but I'd like for us to have more explanation of how this works, especially since there's an external dependency involved)

I'll address these comments and add some documentation on the logger.

PR #387, opened 2021-11-08
Implement a sample linter for OpenAPI that enforces rules on descriptions

muhammadharis:
## Summary
The following pull request creates a sample OpenAPI linter which serves as a proof-of-concept for custom linter creation in the API Registry. It lints two rules:
- `description-less-than-1000-chars`: Ensures that description fields in the OpenAPI spec are less than 1000 characters.
- `description-contains-no-tags`: Ensures that description fields in the OpenAPI spec contain no tags (such as HTML tags).

## Test
- Have registry running on `localhost:8080`.
- Place [the following OpenAPI Sample spec](https://gist.github.com/muhammadharis/13f1681ad65bf6a59366f3068748c0d8) into ~/petstore.yaml 
- Run `LOCAL.sh`
- Ensure that the following style guide is located under `~/styleguide.yaml`:
```yaml
id: openapistyleguide
mime_types:
  - application/x.openapi+gzip;version=3
guidelines:
  - id: Description
    display_name: Govern Description Properties
    rules:
      - id: DescriptionLessThan1000Chars
        description: Description fields should be less than 1000 chars
        linter: openapi-sample
        linter_rulename: description-less-than-1000-chars
        severity: ERROR

      - id: DescriptionContainsNoTags
        description: Description fields should not contain tags
        linter: openapi-sample
        linter_rulename: description-contains-no-tags
        severity: ERROR
    status: ACTIVE
linters:
  - name: openapi-sample
```

Run the following series of commands:

```bash
apg admin delete-project --name projects/example

apg admin create-project --project_id example --insecure --address localhost:8080

apg registry create-api --api_id petstore --parent=projects/example/locations/global

apg registry create-api-version --api_version_id v1 --parent=projects/example/locations/global/apis/petstore

registry upload spec --version=projects/example/locations/global/apis/petstore/versions/v1 --style=openapi ~/petstore.yaml

registry upload styleguide ~/styleguide.yaml --project-id=example

registry compute conformance projects/test/locations/global/apis/petstore/versions/v1/specs/petstore.yaml

registry get projects/example/locations/global/apis/petstore/versions/v1/specs/petstore.yaml/artifacts/conformance-openapistyleguide --contents
```

Ensure that the output matches [the following gist](https://gist.github.com/muhammadharis/7ed03bf240d6b34cbcf97a89f6c2718e).

## Running Unit Tests
`go test -timeout 30s github.com/apigee/registry/cmd/registry/plugins/registry-lint-openapi-sample`

### DIFF
--- a/cmd/registry/plugins/registry-lint-openapi-sample/main.go
+++ b/cmd/registry/plugins/registry-lint-openapi-sample/main.go
@@ -151,7 +151,7 @@ func getDescriptionsFromSpecHelper(node *yaml.Node, results *[]*DescriptionField
 func enforceDescriptionLessThan1000Chars(descriptions *[]*DescriptionField) []*rpc.LintProblem {
 	problems := make([]*rpc.LintProblem, 0)
 	for _, description := range *descriptions {
-		if len(description.Description) >= 100 {
+		if len(description.Description) >= 1000 {
 			problems = append(problems, &rpc.LintProblem{
 				Message: fmt.Sprintf(
 					"Description field should be less than 1000 chars, currently it is\n %s\n",
--- a/cmd/registry/plugins/registry-lint-openapi-sample/main.go
+++ b/cmd/registry/plugins/registry-lint-openapi-sample/main.go
@@ -1,0 +1,224 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/apex/log"
+	lint "github.com/apigee/registry/cmd/registry/plugins/linter"
+	"github.com/apigee/registry/rpc"
+	"gopkg.in/yaml.v3"
+)
+
+// spectralLintCommandExecuter is an interface through which the Spectral command executes.
+type spectralLintCommandExecuter interface {
+	// Runs the spectral linter with a provided spec and configuration path
+	Execute(specPath string, ruleIds []string) ([]*rpc.LintProblem, error)
+}
+
+type DescriptionField struct {
+	LineNumber   int
+	ColumnNumber int
+	Description  string
+}
+
+// sampleOpenApiLinterRunner implements the LinterRunner interface for the Spectral linter.
+type sampleOpenApiLinterRunner struct{}
+
+// concreteSampleOpenApiLintCommandExecuter implements the spectralLintCommandExecuter interface
+// for the Spectral linter.
+type concreteSampleOpenApiLintCommandExecuter struct{}
+
+const descriptionLessThan1000CharsRuleId = "description-less-than-1000-chars"
+const descriptionContainsNoTagsRuleId = "description-contains-no-tags"
+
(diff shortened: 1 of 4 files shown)


timburks (commented):
nit: In the test instructions, 
`apg registry delete-project --name projects/example`
should be
`apg admin delete-project --name projects/example`

cmd/registry/plugins/registry-lint-openapi-sample/main.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/apex/log"
+	lint "github.com/apigee/registry/cmd/registry/plugins/linter"
+	"github.com/apigee/registry/rpc"
+	"gopkg.in/yaml.v3"
+)
+
+// sampleOpenApiLintCommandExecuter is an interface through which the Sample OpenAPI linter executes.
+type sampleOpenApiLintCommandExecuter interface {
+	// Runs the sample OpenAPI linter with a provided spec and configuration path
+	Execute(specPath string, ruleIds []string) ([]*rpc.LintProblem, error)
```
timburks, line 34: This could just be a function type - the struct seems to only exist to define this function - when needed, we could pass a function pointer instead of a struct instance.

cmd/registry/plugins/registry-lint-openapi-sample/main.go
```diff
@@ -0,0 +1,215 @@
...
+const descriptionLessThan1000CharsRuleId = "description-less-than-1000-chars"
+const descriptionContainsNoTagsRuleId = "description-contains-no-tags"
+
+func (linter *sampleOpenApiLinterRunner) Run(req *rpc.LinterRequest) (*rpc.LinterResponse, error) {
+	return linter.RunImpl(req, &concreteSampleOpenApiLintCommandExecuter{})
+}
+
+func (linter *sampleOpenApiLinterRunner) RunImpl(
+	req *rpc.LinterRequest,
+	executer sampleOpenApiLintCommandExecuter,
```
timburks, line 59: This could be a function parameter (as I mentioned above)

cmd/registry/plugins/registry-lint-openapi-sample/main.go
```diff
@@ -0,0 +1,215 @@
...
+			return err
+		}
+
+		// Async API and Open API specs are YAML files
+		if !strings.HasSuffix(path, ".yaml") && !strings.HasSuffix(path, ".yml") {
+			return nil
+		}
+
+		// Execute the linter.
+		problems, err := executer.Execute(path, req.GetRuleIds())
```
timburks, line 75: Here we would call the function that was passed into this function.

cmd/registry/plugins/registry-lint-openapi-sample/main.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return &rpc.LinterResponse{
+		Lint: &rpc.Lint{
+			Name:  "registry-lint-openapi-sample",
+			Files: lintFiles,
+		},
+	}, nil
+}
+
+func (*concreteSampleOpenApiLintCommandExecuter) Execute(specPath string, ruleIds []string) ([]*rpc.LintProblem, error) {
```
timburks, line 101: From above, I think this should just be a function (not an instance method). I also think the name "Execute" isn't very informative. "lintFile" seems to describe what this does and would be one better name for this IMO.

timburks (changes requested):
I'd like to simplify the structure a bit, particularly since this will be sample code.

muhammadharis:
@timburks Thanks so much for the review. I've addressed your comments, and here is a summary of what I addressed:
> It's not clear to me why this is a separate function and not just part of Run()

You are right. I was trying to keep consistent with the Spectral and API Linter plugin in which we were stubbing out the calls to the linter in unit tests through the `executer` since they were path-dependent (would only work if the linter was installed on the path, which might not be the case in unit tests). I have removed additional Impl function.

>This could just be a function type - the struct seems to only exist to define this function - when needed, we could pass a function pointer instead of a struct instance.

Great point, I like this idea. Although since I've removed the executer altogether in this specific PR, I think this doesn't need to be done anymore. However with that being said this method can be used to simplify the logic for the Spectral and API Linter plugin. I will open up a subsequent PR implementing this in Spectral and API linter.

timburks (approved):

PR #391, opened 2021-11-13
Refactor regexs in controller package

shrutiparabgoogle:
Make use of server/registry/names package

Two new interfaces in controller package:
* ResourceName: This is used to create objects for resourcePatterns. Functions like `GetApi()`, `GetSpec()`, etc. will be used to extract entities from resourcePatterns. This interface will be used to create objects for the patterns found in the manifest definition.
* ResourceInstance: This is used to create objects for an instance of the resource. These objects will embed resourceName and also attach a timestamp. This interface will be used to store all the resources received from querying the registry.


shrutiparabgoogle:
Will wait for approval from both @timburks and @seaneganx

cmd/registry/controller/resource-name.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (ar ArtifactName) GetSpec() string {
+	specName := names.Spec{
+		ProjectID: ar.Artifact.ProjectID(),
+		ApiID:     ar.Artifact.ApiID(),
+		VersionID: ar.Artifact.VersionID(),
+		SpecID:    ar.Artifact.SpecID(),
+	}
+	// if err := specName.Validate(); err != nil {
```
shrutiparabgoogle, line 120: @timburks 

I currently can't validate the generated specName because `ValidateID()` functions currently doesn't support `-` hyphens, and thus `Validate()` not supporting such patterns `projects/demo/locations/global/apis/-/versions/-/specs/-`. Am I using this correctly, or am I missing anything here?

timburks: It might be better to call this a "spec pattern" than a "spec name", since it can match multiple specs. The `ValidateID()` function was written to validate actual ids (not patterns) and I think its current usage expects that.

Instead of calling specName.Validate(), you could validate this pattern by passing it to the [`names.ParseSpec`](https://github.com/apigee/registry/blob/cb70a1d794d589db434c0097af98ae471d9d01f4/server/registry/names/spec.go#L127) function, which (I believe) accepts wildcards.

timburks (approved):

cmd/registry/controller/resource-name.go
shrutiparabgoogle: Thanks that makes sense. I can use `ParseSpec` to validate the generated name.

While this issue is not blocking, I realized another thing in the names package behavior:
```
spec, err := names.ParseSpec("projects/demo/locations/global/apis/-/versions/-/specs/-")
	if err != nil {
		fmt.Printf("ERROR PARSE: %s", err)
	}
	err = spec.Validate()
	if err != nil {
		fmt.Printf("ERROR VALIDATE: %s", err)
	}
```

If I execute the above code snippet, `ParseSpec()` allows the presence of `-` hyphens whereas, `Validate()` doesn't support that. This behavior feels a little confusing to me. Either both or non of them should support the use of `-` hyphens.

cmd/registry/controller/parser.go
```diff
@@ -78,39 +110,51 @@ func extendDependencyPattern(
...
-	// apiPattern := regexp.MustCompile(`/apis/-`)
-	apiName := strings.Split(extractEntityName(groupKey, "api"), "/")
-	resourceName := strings.ReplaceAll(resourcePattern, "/apis/-",
-		fmt.Sprintf("/apis/%s", apiName[len(apiName)-1]))
-
-	versionName := strings.Split(extractEntityName(groupKey, "version"), "/")
-	resourceName = strings.ReplaceAll(resourceName, "/versions/-",
-		fmt.Sprintf("/versions/%s", versionName[len(versionName)-1]))
+	groupName, err := parsePattern(groupKey)
+	resourceName := resourcePattern
```
seaneganx, line 127: This copy was confusing to me as a reader because it gives two names for the same value. Eventually I realized its purpose is to maintain the original parameter value so it can be used in the error message at the end of the function. 

Can this function accept resource name types as parameters instead of raw strings? That would eliminate the need for validation entirely, and would also give the function more tools to manipulate the names instead of using string replacement.

seaneganx (approved):
The wrapper types provided in `resource-name.go` and `resource.go` seem problematic to me. They seem to augment our existing types to be more generic but it makes for code that's hard to understand. 

Maybe we can update the names package to be more friendly for generic resource handling.

cmd/registry/controller/list.go
```diff
@@ -50,3 +55,47 @@ func ListResources(ctx context.Context, client connection.Client, pattern, filte
...
+			VersionName:     VersionName{Version: versionName},
+			UpdateTimestamp: version.UpdateTime.AsTime(),
+		}
+		(*result) = append((*result), resource)
+	}
+}
+
+func generateSpecHandler(result *[]ResourceInstance) func(*rpc.ApiSpec) {
+	return func(spec *rpc.ApiSpec) {
+		specName, _ := names.ParseSpec(spec.GetName())
```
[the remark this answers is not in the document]
shrutiparabgoogle: Since we are passing valid rpc objects to these handlers (retrieved through the list response), there won't be a case of invalid specNames, hence ignoring the error here. The only reason to call `names.ParseSpec()` is to convert `rpc.Spec` object into `names.Name` object. Makes sense?

cmd/registry/controller/parser.go
```diff
@@ -78,39 +110,51 @@ func extendDependencyPattern(
...
-	// We use stringsSplit()[-1] to extract only the API name
-	// apiPattern := regexp.MustCompile(`/apis/-`)
-	apiName := strings.Split(extractEntityName(groupKey, "api"), "/")
-	resourceName := strings.ReplaceAll(resourcePattern, "/apis/-",
-		fmt.Sprintf("/apis/%s", apiName[len(apiName)-1]))
-
-	versionName := strings.Split(extractEntityName(groupKey, "version"), "/")
-	resourceName = strings.ReplaceAll(resourceName, "/versions/-",
-		fmt.Sprintf("/versions/%s", versionName[len(versionName)-1]))
+	groupName, err := parsePattern(groupKey)
```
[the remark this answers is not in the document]
shrutiparabgoogle: groupKeys can only be resourceNames and not collections.  You can checkout the `getGroupKey()` function on line 232, which generates these keys. `parsePattern` is a common function which is reused here.

cmd/registry/controller/parser.go
shrutiparabgoogle: I agree the current approach is not the best with string replacement operations. the resourceName is a very generic pattern, it can be an Api, Spec, Version or Artifact. Currently, the way the controller works, it doesn't identify the type of a resource or a resourcePattern (hence the use of generic resource interface). 

However, we can always parse the pattern and figure out what kind it is. I would like to revisit this in a separate PR. I am guessing, changing this logic will need some changes to the interfaces we have defined. Plus if we are making changes to `names` package, then. this logic will be different.

shrutiparabgoogle:
> The wrapper types provided in `resource-name.go` and `resource.go` seem problematic to me. They seem to augment our existing types to be more generic but it makes for code that's hard to understand.
> 
> Maybe we can update the names package to be more friendly for generic resource handling.

If we can get the names package to be more generic then that would work best. However, from whaat I understand, everywhere else in the registry tool, the resources are identified with their type. This notion of a generic resource exists only in the controller. Plus, even if we get a generic package for resourceNames, I will still need to define an interface (`ResourceInstance` defined in `resource.go`) which stores the timestamp of the resources.

cmd/registry/controller/list.go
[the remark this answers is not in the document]
seaneganx: Right, it should never happen. What should happen if we accidentally introduce a bug in the future that results in an error here? I think calling `panic(err)` might be more appropriate than proceeding as if the error didn't occur.

cmd/registry/controller/parser.go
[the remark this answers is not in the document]
seaneganx: Okay. If we ever accidentally call this function with a collection instead of a resource name as expected, the function will return without error. If it's easy to validate that would be a good defensive measure.

cmd/registry/controller/parser.go
seaneganx: Yeah, we should think about how to address this in the names package itself. Definitely not in the scope of this PR. I would like our names package to provide good enough abstractions that client code like this doesn't need to create wrappers.

shrutiparabgoogle: I will revisit this and take another pass at the `names` package.

cmd/registry/controller/parser.go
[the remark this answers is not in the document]
shrutiparabgoogle: Done. Broke the `parsePattern()` into separate functions which individually parse Collections ane Names.

cmd/registry/controller/list.go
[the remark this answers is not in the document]
shrutiparabgoogle: Thanks for pushing for this.

Issue #389, opened 2021-11-12
Admin method to perform database migration


timburks:
Since [gorm supports automatic database migration](https://gorm.io/docs/migration.html), it seems possible to add an administrative API to automatically migrate an existing database.

It seems appropriate to add this to the Admin service and call it something like "MigrateDatabase".

timburks:
Done in #393

PR #393, opened 2021-11-16
Add a "migrate database" API.

timburks:
Addresses #389


timburks:
The new MigrateDatabase method returns a completed LRO object - allowing future implementations to return incompleted LROs for long-running migrations.

google/cloud/apigeeregistry/v1/admin_service.proto
```diff
@@ -105,6 +117,23 @@ message Status {
   string message = 1;
 }
 
+// Request message for MigrateDatabase.
+message MigrateDatabaseRequest {
+  // A string describing the kind of migration to perform.
+  // Currently only "auto" is recognized (and is the default if omitted).
+  string kind = 1;
```
seaneganx, line 124: Do we expect this to change frequently? [AIP-126](https://google.aip.dev/126) recommends enums unless we'll be frequently modifying the possible values here. 

As a caveat, I think we should use enum unless we think it will change frequently _and_ have clients using older versions of this proto to call newer versions of the admin code. We don't really have to worry about breaking changes to enum fields until there's somebody to break with the change.

server/registry/actions_migrate_database.go
```diff
@@ -0,0 +1,54 @@
...
+	"google.golang.org/genproto/googleapis/longrunning"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+// MigrateDatabase handles the corresponding API request.
+func (s *RegistryServer) MigrateDatabase(ctx context.Context, req *rpc.MigrateDatabaseRequest) (*longrunning.Operation, error) {
+	if req.Kind != "" && req.Kind != "auto" {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Unsupported migration kind '%s'", req.Kind))
```
seaneganx, line 31: Error messages typically aren't capitalized, and `status.Errorf` can format the message as a convenience. Note %q will wrap the parameter in quotes.

`status.Errorf(codes.InvalidArgument, "unsupported migration kind %q", req.Kind)`

google/cloud/apigeeregistry/v1/admin_service.proto
```diff
@@ -105,6 +117,23 @@ message Status {
...
+}
+
+// Metadata message for MigrateDatabase.
+message MigrateDatabaseMetadata {
+}
+
+// Response message for MigrateDatabase.
+message MigrateDatabaseResponse {
+  // A string describing the result of the migration.
+  string message = 1;
```
seaneganx, line 134: What do you expect this field to indicate in the future? Right now I see it's just set to "OK" but that doesn't seem to provide any more information than the response status code.

server/registry/internal/storage/gorm/client.go
```diff
@@ -116,9 +116,8 @@ func (c *Client) ensureTable(v interface{}) error {
 	return nil
 }
 
-// EnsureTables ensures that all necessary tables exist in the database.
-func (c *Client) EnsureTables() error {
-	entities := []interface{}{
+func (c *Client) entities() []interface{} {
```
seaneganx, line 119: Can this be moved into a package level variable now?

google/cloud/apigeeregistry/v1/admin_service.proto
timburks: A string value allows us to add options without changing the API - as you noted, adding enums are breaking changes and would require regenerating/rebuilding API clients. 

I'm not sure what we might agree is "frequent", but I think adding a custom database migration might be a change that we want to make very quickly (perhaps responding to a problem), so the less overhead associated with that, the better.

google/cloud/apigeeregistry/v1/admin_service.proto
timburks: I think a useful improvement would be to report whether anything changed in the migration.

server/registry/internal/storage/gorm/client.go
timburks: Sure, I'll do that.

server/registry/actions_migrate_database.go
timburks: OK, will update.

timburks: done, thanks!

server/registry/internal/storage/gorm/client.go
timburks: done

cmd/apg/admin_service.go
```diff
@@ -18,7 +18,8 @@ var AdminConfig *viper.Viper
 var AdminClient *gapic.AdminClient
 var AdminSubCommands []string = []string{
 	"get-status",
-	"list-projects",
+	"migrate-database",
+	"poll-migrate-database", "list-projects",
```
wangyh28, line 22: Minor, should `list-projects` on a separate line?

cmd/apg/migrate-database.go
```diff
@@ -0,0 +1,143 @@
...
+var MigrateDatabasePollOperation string
+
+func init() {
+	AdminServiceCmd.AddCommand(MigrateDatabaseCmd)
+
+	MigrateDatabaseCmd.Flags().StringVar(&MigrateDatabaseInput.Kind, "kind", "", "A string describing the kind of migration to...")
+
+	MigrateDatabaseCmd.Flags().StringVar(&MigrateDatabaseFromFile, "from_file", "", "Absolute path to JSON file containing request payload")
+
+	MigrateDatabaseCmd.Flags().BoolVar(&MigrateDatabaseFollow, "follow", false, "Block until the long running operation completes")
```
wangyh28, line 32: Within g3 LRO, we usually use `wait` to wait until LRO finish

cmd/apg/migrate-database.go
```diff
@@ -0,0 +1,143 @@
...
+				d := make(map[string]string)
+				d["operation"] = resp.Name()
+				s = d
+			}
+
+			printMessage(s)
+			return err
+		}
+
+		result, err := resp.Wait(ctx)
```
wangyh28, line 94: What's the default timeout on the WaitOperation? Are we considering a flag on customizing waiting time?

gapic/admin_client.go
```diff
@@ -360,6 +417,75 @@ func (c *adminGRPCClient) DeleteProject(ctx context.Context, req *rpcpb.DeletePr
...
+
+// MigrateDatabaseOperation returns a new MigrateDatabaseOperation from a given name.
+// The name must be that of a previously created MigrateDatabaseOperation, possibly from a different process.
+func (c *adminGRPCClient) MigrateDatabaseOperation(name string) *MigrateDatabaseOperation {
+	return &MigrateDatabaseOperation{
+		lro: longrunning.InternalNewOperation(*c.LROClient, &longrunningpb.Operation{Name: name}),
+	}
+}
+
+// Wait blocks until the long-running operation is completed, returning the response and any errors encountered.
```
wangyh28, line 433: Minor; Might be better to include `timeout` here as well, e.g. Wait blocks until the long-running operation is completed or timeout reached.

google/cloud/apigeeregistry/v1/admin_service.proto
```diff
@@ -47,6 +48,17 @@ service Admin {
     };
   }
 
+  // MigrateDatabase attempts to migrate the database to the current schema.
+  rpc MigrateDatabase(MigrateDatabaseRequest) returns (google.longrunning.Operation) {
+    option (google.api.http) = {
+      post: "/v1/migrate"
```
wangyh28, line 54: Minor; Would `/v1/migrateDatabase` would be more detailed and consistent with the rpc name?

ps: I'm not sure which way is better: /migrate or :migrate (as custom method), but I do found examples in [codesearch using /v1/migrate](https://source.corp.google.com/search?q=file:.*.proto$%20%22%2Fv1%2Fmigrate%22&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental)

google/cloud/apigeeregistry/v1/admin_service.proto
wangyh28: From on [AIP-180](https://google.aip.dev/180): Enum values may be freely added to enums which are only used in request messages. So I'd like to understand the `breaking change` part better.

Let's say we're using enum for `kind`, and we're adding a new enum value besides `auto` (e.g. `partial`); under this case, if a user calls the new version of the admin code from an old API client, would the user still be able to make successful MigrateDatabase request with `kind = auto`?

cmd/apg/admin_service.go
timburks: I would think so, but this is auto-generated and automatically-formatted code.

google/cloud/apigeeregistry/v1/admin_service.proto
timburks: @wangyh28 Yes, but the user wouldn't be able to pass the new `kind` value without regenerating/rebuilding their code. If we use a string field for `kind`, then it is just a configuration change (assuming that the code that calls the migration API is written to read a configured value for `kind`)

cmd/apg/migrate-database.go
timburks: I don't know -- Here's the source for the CLI generator https://github.com/googleapis/gapic-generator-go/tree/master/cmd/protoc-gen-go_cli.

google/cloud/apigeeregistry/v1/admin_service.proto
timburks: I'm ok with that - for consistency we could either use "migrate" for both or "[Mm]igrateDatabase" for both. Do you have a preference? (I think I slightly prefer the more explicit "migratedatabase" but it also seems that "database" is superfluous)

gapic/admin_client.go
timburks: Like the CLI, this is automatically-generated code (https://github.com/googleapis/gapic-generator) - I'm not sure what configuration options exist for that.

google/cloud/apigeeregistry/v1/admin_service.proto
timburks: @seaneganx If you've gotten a copy of [API Design Patterns](https://www.manning.com/books/api-design-patterns), you can read JJ's discussion of enums there, which is more strongly cautionary than the warnings in AIP-126. From the conclusion of Chapter 5: "Enumerations should generally be avoided, relying on string values instead, with validation done on the server side rather than the client side." That''s expanded earlier in section 5.5, "while [enumerations] might be wonderful things in languages like Java, their transplantation into the world of web APIs is often a mistake."

cmd/apg/migrate-database.go
timburks: SGTM - this is code generated by another project, so not something we can address here.

google/cloud/apigeeregistry/v1/admin_service.proto
wangyh28: I'm okay with both, but migrateDatabase is slightly clear from my perspective

gapic/admin_client.go
wangyh28: I see; didn't noticed it was auto-generated

cmd/apg/migrate-database.go
wangyh28: I haven't tried to use the migrate functionality yet, so not sure how long would it takes on the blocking behavior. This should be a good start, and we can add a new flag if we find use cases that need customizing waiting time in the future

cmd/apg/migrate-database.go
wangyh28: Ack. Would this be addressed separately or we'll use `--follow` for a while? I'm thinking when would be a good time to update data plane's provisioning template

google/cloud/apigeeregistry/v1/admin_service.proto
timburks: OK! I've changed that to "migrateDatabase"

cmd/apg/migrate-database.go
timburks: If you're using the `apg` tool in provisioning, I think using `--follow` would be much less overhead than changing the CLI generator. Also, this always returns completed operations and the PR doesn't include an implementation of the LRO service - we'll want to add that eventually.

wangyh28 (approved):

seaneganx (approved):

PR #394, opened 2021-11-18
Added composite actions to build containers

giteshk:
* Added configuration to build registry-tool and registry-linter ghcr.io images
* Broke down the steps to run processes in parallel
* Build composite action to build the various images
* added tags for branch/tags/latest/nightly


giteshk:
registry-linter step is failing because there is no public image at the moment for the registry-tool and that is a dependency.

Here is the build in my current repo: https://github.com/giteshk-org/registry/actions/runs/1474881451

timburks (approved):

.github/workflows/docker-publish.yml
```diff
@@ -28,71 +28,67 @@ on:
...
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - uses: ./.github/workflows/composite/build-push
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ env.USERNAME }}
+        password: ${{ env.PASSWORD }}
+        image: ${{ env.DOCKER_REPOSITORY_OWNER }}/registry-tool
+        context: .
+        file: containers/registry-tool/Dockerfile
```
timburks, line 63: Shouldn't this be "registry-tools"? (it's plural because it contains the `registry` tool and the automatically-generated `apg` tool)

timburks (changes requested):
I don't think we should rename "containers/registry-tools" to "containers/registry-tool"

.github/workflows/docker-publish.yml
giteshk: I changed it back , realized it has multiple tools

giteshk:
This build does not complete because of a docker image which has not yet been published on ghcr.io 

I have this completed build here for review : https://github.com/giteshk-org/registry/actions/runs/1518643638

timburks (approved):

PR #399, opened 2021-12-02
Fix GitHub container registry addresses

seaneganx:
I don't see any runs for this action in our [history](https://github.com/apigee/registry/actions) but it looks like it will fail if it ever runs because these addresses point to an available domain name gchr.io, not the ghcr.io domain used by GitHub's container registry.


seaneganx:
Looks like these are default values that are currently unused by our actions. Still we should correct them now so we don't get surprised with future changes that result in their use.

giteshk (approved):

PR #400, opened 2021-12-06
Add CRUD benchmarks for API methods

seaneganx:
After discussing some ideas for benchmarking the performance of our API
methods I wanted to see how Go's benchmark package would work. This
change includes some simple examples for the API resource's CRUD methods.

These benchmarks don't examine the response for correctness (aside from
an error check) because it might significantly affect the performance.
The benchmarks don't need to (and shouldn't) test the correctness of
responses because other test suites already check this.

The tests can executed using `go test -bench=. ./tests/benchmark/...` or
compiled using `go test -c ./tests/benchmark/...` and executed using the
resulting binary with the `test.bench=.` flag.

Example benchmark output:

```
goos: darwin
goarch: amd64
pkg: github.com/apigee/registry/tests/benchmark
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkCreateApi-12    	     152	   7968111 ns/op
BenchmarkGetApi-12       	     198	   6026086 ns/op
BenchmarkUpdateApi-12    	     157	   7667434 ns/op
BenchmarkDeleteApi-12    	      94	  11900588 ns/op
PASS
```

### DIFF
--- a/tests/benchmark/apis_test.go
+++ b/tests/benchmark/apis_test.go
@@ -1,0 +1,157 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package benchmark
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"github.com/google/uuid"
+)
+
+const (
+	rootResource = "projects/bench/locations/global"
+)
+
+func uniqueID() string {
+	return fmt.Sprintf("%.8s", uuid.New())
+}
+
+func BenchmarkCreateApi(b *testing.B) {
+	ctx := context.Background()
+	client, err := connection.NewClient(ctx)
+	if err != nil {
+		b.Fatalf("Setup: Failed to create client: %s", err)
+	}
+
+	reqs := make([]*rpc.CreateApiRequest, b.N)
+	for i := 0; i < b.N; i++ {
+		reqs[i] = &rpc.CreateApiRequest{
+			Parent: rootResource,
+			ApiId:  uniqueID(),
+			Api:    &rpc.Api{},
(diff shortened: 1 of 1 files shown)


tests/benchmark/apis_test.go
```diff
@@ -0,0 +1,157 @@
...
+			Api:    &rpc.Api{},
+		}
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		if _, err := client.CreateApi(ctx, reqs[i]); err != nil {
+			b.Errorf("CreateApi(%+v) returned unexpected error: %s", reqs[i], err)
+		}
+	}
```
timburks, line 56: I noticed that these tests don't clean up after themselves - after running the benchmarks once locally I found over 1600 APIs left in the bench project. Results shouldn't be dependent on the number of elements in the database but I think they probably are and would continue to be. Could we use [StopTimer()](https://godocs.io/testing#B.StopTimer) to stop time measurement before cleaning up at the end of each test?

tests/benchmark/apis_test.go
```diff
@@ -0,0 +1,157 @@
...
+	reqs := make([]*rpc.GetApiRequest, b.N)
+	for i := 0; i < b.N; i++ {
+		api, err := client.CreateApi(ctx, &rpc.CreateApiRequest{
+			Parent: rootResource,
+			ApiId:  uniqueID(),
+			Api:    &rpc.Api{},
+		})
+		if err != nil {
+			b.Fatalf("Setup: Failed to create API: %s", err)
+		}
```
timburks, line 75: Lines 68-75 are repeated in every test. Would you be open to replacing them with a function call that returns an API name? (not necessary in this PR, just as a design pattern)

tests/benchmark/apis_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"github.com/google/uuid"
+)
+
+const (
+	rootResource = "projects/bench/locations/global"
```
timburks, line 28: We might occasionally need to use a different value for this. Can the tests take arguments? We could also set it with an environment variable (that we replace with a default value if unset)

timburks (approved):
LGTM, I just have some questions about applications and scaling

tests/benchmark/apis_test.go
timburks: I agree with the `APG_REGISTRY_PROJECT` suggestion. We need something similar for the [registry-viewer](https://github.com/apigee/registry-viewer).

tests/benchmark/apis_test.go
timburks: SGTM.

timburks:
@seaneganx Thanks for these extra updates. This all LGTM for merging.

Issue #396, opened 2021-11-19
Admin method to obtain build info


timburks:
It would be useful to be able to query a server and obtain the git tag or commit hash corresponding to the build.

One way to do this would be to add a field to the Admin GetStatus response that includes the Go [BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo), which is easy to obtain at runtime and requires no special build configuration. This currently reports the main module version as "(devel)", but in go 1.18 this will [include commit hash and other information about VCS state](https://github.com/golang/go/issues/37475).

PR #403, opened 2021-12-08
Patch apg to make the Admin client use the Registry client configuration

timburks:
This eliminates the use of APG_ADMIN_* variables to configure the Admin service client, replacing them with APG_REGISTRY_* so that both services are configured with the same variables.

### DIFF
--- a/tools/GENERATE-APG.sh
+++ b/tools/GENERATE-APG.sh
@@ -30,3 +30,16 @@ protoc ${SERVICE_PROTOS[*]} \
   	--go_cli_opt='root=apg' \
   	--go_cli_opt='gapic=github.com/apigee/registry/gapic' \
   	--go_cli_out='cmd/apg'
+
+# Patch the generated CLI to use "APG_REGISTRY" as the prefix for
+# Admin service configuration variables. This causes the Admin service
+# client to be configured with the same variables that configure the
+# Registry service client.
+FILE=cmd/apg/admin_service.go
+sed -i.bak "s/APG_ADMIN/APG_REGISTRY/" "${FILE}"
+rm "${FILE}.bak"
+gofmt -w "${FILE}"
+if grep --quiet APG_ADMIN "${FILE}"; then
+  echo "Patching APG tool failed."
+  exit 1
+fi
--- a/auth/CLOUDRUN.sh
+++ b/auth/CLOUDRUN.sh
@@ -46,10 +46,3 @@ export APG_REGISTRY_TOKEN=$(gcloud auth print-identity-token ${APG_REGISTRY_CLIE
 
 # Calls don't use an API key.
 unset APG_REGISTRY_API_KEY
-
-# Duplicate the client configuration for the Admin service.
-unset APG_ADMIN_INSECURE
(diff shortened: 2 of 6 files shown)


tools/GENERATE-APG.sh
```diff
@@ -30,3 +30,16 @@ protoc ${SERVICE_PROTOS[*]} \
   	--go_cli_opt='root=apg' \
   	--go_cli_opt='gapic=github.com/apigee/registry/gapic' \
   	--go_cli_out='cmd/apg'
+
```
timburks, line 33: If we're ok with this patch, we could go further and replace "APG_REGISTRY" with "REGISTRY", but that's a bigger change that we would need to propagate through the repo and any client configuration scripts.

timburks: @shrutiparabgoogle I can imagine how that might happen in an authz configuration but it seems more likely (and better) for authz to give tokens varying sets of permissions. A more privileged token would have access to both services and a less-privileged one would only get to use the Registry service. On a much bigger scale, this is how Google auth tokens work - we get a token but it only works for the services we've enabled, and as we enable new services, that same token gets access to those new services.

seaneganx (commented):
Why do we want to do this? I would like to know we're solving an actual problem.

Generally I'm struggling to understand why we split the admin service from the registry service but keep them in the same binary and want them to use the same configuration. This change will make it difficult for anyone to work with these "separate" services if they're deployed at different addresses.

timburks:
@seaneganx Thanks for pointing out the lack of clarity. The split was to better organize and improve the maintainability of our protos - now the `Registry` service that is defined in this repo is exactly identical to the hosted version. This means that the `registry_service.proto` and `registry_models.proto` here will be identical to the ones in github.com/googleapis/googleapis - so any clients generated with either set of protos will be interchangeable.

Rather than just delete the rpcs that we removed from the `Registry` service, we moved them to `Admin` so that they could still be used when we and other users run our own instances of the OSS project and so that hosting teams can use the `Admin` rpcs to set up and manage hosted instances. 

Thus `Admin` is an administrative interface to the server that provides the `Registry` service. Considering https://github.com/apigee/registry/issues/396, it doesn't really make sense to me to run `Admin` with a different binary than `Registry`.

So although the GAPIC CLI generator allows for the `Admin` and `Registry` services to have their own configurations, it doesn't seem likely that these configurations would ever differ, and the scripts in the `auth` directory [currently just copy the configurations](https://github.com/apigee/registry/blob/99a996216c2806eb6f2eb93cdfcf95e28b8b9da8/auth/LOCAL.sh#L50). That seems like a nuisance, and  this makes it unnecessary. It's purely a client-side change that only affects the `apg` tool - the `registry` tool already uses `APG_REGISTRY_*` for both services (see [connection/client.go](https://github.com/apigee/registry/blob/99a996216c2806eb6f2eb93cdfcf95e28b8b9da8/connection/client.go#L38)).

shrutiparabgoogle (approved):

seaneganx (approved):
I think the summary is that we don't think anyone should do the thing this prevents, so removing the duplication in these scripts is worth it.

PR #404, opened 2021-12-09
List deployment artifacts

timburks:
This adds the ability to list artifacts that belong to deployments and verifies that deployments exist when artifacts are created. Thanks @krpate for finding these problems!

### DIFF
--- a/server/registry/actions_artifacts_test.go
+++ b/server/registry/actions_artifacts_test.go
@@ -197,10 +197,50 @@ func TestCreateArtifactResponseCodes(t *testing.T) {
 		want codes.Code
 	}{
 		{
-			desc: "parent not found",
+			desc: "parent project not found",
 			seed: &rpc.Project{Name: "projects/my-project"},
 			req: &rpc.CreateArtifactRequest{
 				Parent:     "projects/other-project/locations/global",
+				ArtifactId: "valid-id",
+				Artifact:   &rpc.Artifact{},
+			},
+			want: codes.NotFound,
+		},
+		{
+			desc: "parent api not found",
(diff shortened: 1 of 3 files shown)


server/registry/actions_artifacts_test.go
```diff
@@ -206,6 +206,46 @@ func TestCreateArtifactResponseCodes(t *testing.T) {
...
+				ArtifactId: "valid-id",
+				Artifact:   &rpc.Artifact{},
+			},
+			want: codes.NotFound,
+		},
+		{
+			desc: "parent version not found",
+			seed: &rpc.Project{Name: "projects/my-project"},
+			req: &rpc.CreateArtifactRequest{
+				Parent:     "projects/other-project/locations/global/apis/a/versions/v",
```
Krpate, line 223: In this test case, the api `a` itself doesn't exist. For me, it seems more like an `API not found` test case than `version not found`. 

Same for below cases.

Krpate (approved):

server/registry/actions_artifacts_test.go
timburks: That's true - the description is just a label here, and the returned error code is a general "NotFound" that doesn't specify what is missing. The implementation only verifies that the parent exists. If no parent exists, the CreateArtifact implementation might climb the ancestry to report the highest-level missing ancestor, but currently it doesn't do that.

seaneganx (approved):
I agree with Kritika about the tests. Please update them to seed the grandparent resource before testing the parent check. The update is trivial and reduces the possibility of false test passes.

LGTM otherwise.

PR #405, opened 2021-12-15
Update GAPIC config: 20s timeouts, additional retries

timburks:
Update GAPIC configuration to increase timeouts to 20s and retry two additional gRPC response codes (Abort, Cancelled).

In testing with hosted deployments, "Abort" codes are returned when prerequisite operations have failed to complete. Retrying these aborted calls generally succeeds. "Cancelled" codes are returned when calls are rate-limited. Again, retrying these callsl generally succeeds (with appropriate backoffs, which are built into the GAPICs).

The timeout increase works around issues where proxies preemptively expire cancel requests with timeouts that are considered too short.

### DIFF
--- a/gapic/grpc_service_config.json
+++ b/gapic/grpc_service_config.json
@@ -6,15 +6,17 @@
             "service": "google.cloud.apigeeregistry.v1.Registry"
           }
         ],
-        "timeout": "10s",
+        "timeout": "20s",
         "retryPolicy": {
           "initialBackoff": "0.200s",
           "maxBackoff": "10s",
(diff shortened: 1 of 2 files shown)


seaneganx (approved):
Note: Retrying cancelled calls might cause issues if we ever end up cancelling requests from the client.

timburks:
@seaneganx Good point - we'll need to figure out how to differentiate explicit client cancellations and the ones occurring from rate-limiting. // cc @krpate to confirm my understanding that rate-limiting is causing "Cancel" responses that we've been seeing.

PR #406, opened 2021-12-16
Add a --jobs flag to control the number of parallel bulk uploads.

timburks:


cmd/registry/cmd/upload/bulk/bulk.go
```diff
@@ -32,5 +32,7 @@ func Command(ctx context.Context) *cobra.Command {
 
 	cmd.PersistentFlags().String("project-id", "", "Project ID to use for each upload")
 	_ = cmd.MarkFlagRequired("project-id")
+
+	cmd.PersistentFlags().Int("jobs", 10, "Number of upload jobs to run simultaneously")
```
timburks, line 36: Originally I was thinking of doing this on the root but it seemed clearer to only have it on commands that use parallel workers. I expect that we'll want to duplicate this for all such commands.

cmd/registry/cmd/upload/bulk/openapi.go
```diff
@@ -50,12 +50,20 @@ func openAPICommand(ctx context.Context) *cobra.Command {
 				log.FromContext(ctx).WithError(err).Fatal("Failed to get client")
 			}
 
+			// create a queue for upload tasks and wait for the workers to finish after filling it.
+			jobs, err := cmd.Flags().GetInt("jobs")
```
seaneganx, line 54: Nit: I think `jc`, `jobCount`, or `poolSize` would be better in the context of this code.

seaneganx (approved):

cmd/registry/cmd/upload/bulk/openapi.go
timburks: Thanks, based on this I did some looking for other precedents. I found that Bazel and GNU Parallel both use `--jobs` consistently with make, and `go test` has a `--parallel` flag. Comparing `jobs` vs `parallel`, I prefer the shorter name and the more widespread usage of `make` (`bazel` is probably less-used than `go` but I'm not sure about `go test`). Also, `--jobs` is listed in the [GNU Coding Standards](https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table) and `--parallel` is not.

For the listed suggestions, `--jc` is an initialism and seems more like the single-character short forms of [POSIX options](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html). That same page describes "long options":

> GNU adds long options to these conventions. Long options consist of -- followed by a name made of alphanumeric characters and dashes. Option names are typically one to three words long, with hyphens to separate words. Users can abbreviate the option names as long as the abbreviations are unique. 

So (nit), `--job-count` and `--pool-size` would be more consistent with the POSIX standard. Since `--jobs` has strong precedent, I prefer it to the equivalent `--job-count`. I think `--pool-size` emphasizes an implementation detail (task pools) which isn't essential to parallel execution, so I prefer `--jobs` over that because it is more in what I expect to be the user's mental model.

I'll put this in our tool design doc and we can discuss it further.

PR #408, opened 2021-12-17
Add auth/HOSTED.sh to authenticate with Google-hosted services.

timburks:


auth/HOSTED.sh
```diff
@@ -0,0 +1,41 @@
...
+if ! [ -x "$(command -v gcloud)" ]; then
+  echo 'ERROR: This script requires the gcloud command. Please install it to continue.' >&2; return
+fi
+
+# Calls to the hosted service are secure.
+unset APG_REGISTRY_INSECURE
+
+# Get the service address.
+export APG_REGISTRY_AUDIENCES=https://apigeeregistry.googleapis.com
+export APG_REGISTRY_ADDRESS=${APG_REGISTRY_AUDIENCES#https://}:443
```
seaneganx, line 34: What does this do? A comment would be useful because bash syntax can be hard to search.

auth/HOSTED.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# The auth token is generated for the gcloud logged-in user.
+export APG_REGISTRY_CLIENT_EMAIL=$(gcloud config list account --format "value(core.account)")
```
seaneganx, line 37: Consider `ACCOUNT` instead of `EMAIL` if this is going to be used as a config value. That seems to be the terminology used by gcloud, although I think accounts might always be formatted as email addresses.

If it's just a variable meant to be used on the following line, let's keep it unexported.

seaneganx (approved):

auth/HOSTED.sh
timburks: Worth discussing; I've been looking for the original source that I used for this and haven't found anything using "EMAIL" in the variable name. That is consistently used in multiple files in the `auth` directory, so I prefer not to change it here, but wouldn't rule it out.

auth/HOSTED.sh
timburks: It removes the "https://" prefix. It's an obscure bash feature but I thought it's effect was clear from the context. It isn't really necessary in this file, I just used it for consistency with CLOUDRUN.sh. I've replaced it with something more explicit (and repetitive).

PR #411, opened 2021-12-30
Improve listing and deletion of deployments and related artifacts

timburks:
This adds missing deployment-related features to the registry tool and fixes a "leak" of artifact blobs that were not being deleted with their associated deployments.

### DIFF
--- a/cmd/registry/core/list.go
+++ b/cmd/registry/core/list.go
@@ -80,6 +80,35 @@ func ListAPIs(ctx context.Context,
 	return nil
 }
 
+func ListDeployments(ctx context.Context,
+	client *gapic.RegistryClient,
+	name names.Deployment,
+	filterFlag string,
+	handler DeploymentHandler) error {
+	request := &rpc.ListApiDeploymentsRequest{
+		Parent: name.Parent(),
+	}
+	filter := filterFlag
+	deploymentID := name.DeploymentID
+	if deploymentID != "" && deploymentID != "-" {
+		filter = "deployment_id == '" + deploymentID + "'"
+	}
+	if filter != "" {
+		request.Filter = filter
+	}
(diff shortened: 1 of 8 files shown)


cmd/registry/core/list.go
```diff
@@ -80,6 +80,35 @@ func ListAPIs(ctx context.Context,
...
 
+func ListDeployments(ctx context.Context,
+	client *gapic.RegistryClient,
+	name names.Deployment,
+	filterFlag string,
+	handler DeploymentHandler) error {
+	request := &rpc.ListApiDeploymentsRequest{
+		Parent: name.Parent(),
+	}
+	filter := filterFlag
```
seaneganx, line 91: Can we avoid this? Maybe just rename the argument instead of making a new variable.

cmd/registry/core/list.go
```diff
@@ -80,6 +80,35 @@ func ListAPIs(ctx context.Context,
...
+	deploymentID := name.DeploymentID
+	if deploymentID != "" && deploymentID != "-" {
+		filter = "deployment_id == '" + deploymentID + "'"
+	}
+	if filter != "" {
+		request.Filter = filter
+	}
```
seaneganx, line 98: `request.Filter` is an empty string by default because we don't set it when we declare the request, so overwriting it with another empty string is a no-op that doesn't need to be avoided with this check. We can set this field when the request is declared, like how we set the parent.

seaneganx (approved):

cmd/registry/core/list.go
timburks: Thanks, noted in #420 - this pattern is repeated for different resource types throughout this file and fixing it seems complex enough to warrant a separate PR (and it looks like there's also a bug to be fixed).

Issue #412, opened 2022-01-04
Admin method to obtain database information


timburks:
It would be useful, particularly during testing, to be able to query a server and obtain database statistics including the number of entities (rows) in all tables and the amount (bytes) of storage used. 

One way to do this would be to expand the Admin GetStatus response to include this information in additional fields.

timburks:
I'm not sure that we need the number bytes of storage used (or what we might do with that information), so let's exclude that for now.

PR #415, opened 2022-01-05
Add a new admin API method to return database information.

timburks:
This adds a GetStorage RPC to the admin service that returns a list of tables in the backend storage and the number of entries stored in each table. It is separate from GetStatus because it adds a small overhead (some database queries) and might be worth extending in the future to include other information such as schemas, index descriptions, and bytes of storage required for each entity type.

Addresses #412


server/registry/internal/storage/gorm/client.go
```diff
@@ -349,3 +353,38 @@ func (c *Client) GetRecentDeploymentRevisions(ctx context.Context, offset int32,
 func (c *Client) Migrate(kind string) error {
 	return c.db.AutoMigrate(entities...)
 }
+
+func (c *Client) GetStorage() (*rpc.Storage, error) {
```
seaneganx, line 357: Consider splitting this into separate methods, e.g. `DatabaseName() (string, error)`, `TableNames() ([]string, error)`, and `RowCount(table string) int`. The RPC handler can use these to build a response, consistent with how we implement other methods.

Each individual method is easier to understand than one larger and more vaguely named `GetStorage()` method, and it decrease coupling between the storage implementation and our API by keeping RPCs out of the storage package.

On the other hand, I can see value in your current approach where the storage layer responds with RPC messages. If we consistently did that, the storage models would become internal to the storage layer and the RPCs would no longer have to manage that type conversion.

server/registry/actions_storage_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	req := &emptypb.Empty{}
+
+	resp, err := server.GetStorage(ctx, req)
+	if err != nil {
+		t.Fatalf("GetStorage(%+v) returned error: %s", req, err)
+	}
+
+	// Ensure that we get the set of tables that we expect.
+	// Tables should be returned in alphabetical order.
```
seaneganx, line 38: This isn't mentioned in the API documentation. IMO we shouldn't implement or rely on it unless we make it a feature. If it's just for reliable comparison in tests we can use [cmpopts.SortSlices](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices).

server/registry/actions_storage_test.go
```diff
@@ -0,0 +1,47 @@
...
+	want := []string{"apis", "artifacts", "blobs", "deployment_revision_tags", "deployments", "projects", "spec_revision_tags", "specs", "versions"}
+	got := make([]string, 0)
+	for _, c := range resp.Collections {
+		got = append(got, c.Name)
```
seaneganx, line 42: Are row counts affected by other tests running on the same database? If possible we should test it, but I understand if the environment isn't isolated enough to do that reliably.

seaneganx (approved):

server/registry/actions_storage_test.go
timburks: Good point - these numbers come from the database and so they are global across projects.

server/registry/actions_storage_test.go
timburks: OK, SGTM - I'll remove the sort and fix the test as you suggest.

server/registry/internal/storage/gorm/client.go
timburks: Good points - I've restructured this to use the methods that you suggested and build the response in the action handler. I'm not sure that we need to keep the proto models out of the storage layer, but it doesn't seem worthwhile to bring them in for a small feature like this.

server/registry/actions_storage_test.go
timburks: Backtracking on this, I realized that the table ordering could be handled by the SQL server with a trivial change on our side, and I think that consistent ordering makes the results more human-readable. So I'm suggesting that this API is for more than just testing - I think human users could also use it to diagnose problems in deployed instances such as accidentally dropped or renamed tables.

To the point about documentation, I've added a comment in the protos indicating that collection names are alphabetized. Thanks!

PR #416, opened 2022-01-06
Add Build Information to the Admin GetStatus RPC

timburks:
Addresses #396. This uses Go build tags to include Go 1.18-only features when available (when Go 1.18 or later is used to build `registry-server`).

This builds on #415 and should be merged after that.


google/cloud/apigeeregistry/v1/admin_models.proto
```diff
@@ -25,6 +25,71 @@ option java_multiple_files = true;
...
+  // A setting (key-value pair) associated with the build.
+  message Setting {
+    // The key used to store the setting.
+    string key = 1;
+    // The value of the setting.
+    string value = 2;
+  }
+
+  // Settings associated with the build.
+  repeated Setting settings = 5;
```
seaneganx, line 63: I think this would be better represented as a map.

server/registry/internal/buildinfo/buildinfo-1.18.go
```diff
@@ -0,0 +1,60 @@
...
+
+package buildinfo
+
+import (
+	"runtime/debug"
+
+	"github.com/apigee/registry/rpc"
+)
+
+func Module(m *debug.Module) *rpc.BuildInfo_Module {
```
seaneganx, line 25: This shouldn't be exported.

google/cloud/apigeeregistry/v1/admin_models.proto
```diff
@@ -25,6 +25,71 @@ option java_multiple_files = true;
...
+}
+
+// Status represents the status of the service.
+message Status {
+  // A string describing the status.
+  string message = 1;
+
+  // Information about the build of the server.
+  BuildInfo buildinfo = 2;
```
seaneganx, line 72: Shouldn't this be `build_info`? Maybe just `build` is enough detail though.

google/cloud/apigeeregistry/v1/admin_models.proto
```diff
@@ -25,6 +25,71 @@ option java_multiple_files = true;
...
+  // A module used to create the build.
+  message Module {
+    // The module path.
+    string path = 1;
+    // The module version.
+    string version = 2;
+    // The module checksum.
+    string sum = 3;
+    // An optional replacement for the module
+    Module replacement = 4;
```
seaneganx, line 45: Do we need this? It seems to cover a very specific situation where the module containing the main package is replaced with a different module, but I can't imagine why we would ever do that or want to see it in the build info from a running server.

seaneganx (changes requested):
Please remove `buildinfo-prev.go` and the storage changes from #415 (until it's merged). Otherwise LGTM.

google/cloud/apigeeregistry/v1/admin_models.proto
timburks: I might have been overindexing on the naming in the Go runtime/debug package. I think both of your suggestions are better names and prefer `build` for brevity.

google/cloud/apigeeregistry/v1/admin_models.proto
timburks: Yes (if I understand you), replacing the main module would be weird.

This message is also used to represent dependencies, and I think that's what led to this [field's inclusion in the runtime/debug package](https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/runtime/debug/mod.go;drc=987e4e892325ea91a15a34f7a858ab007fea39c4;l=34) - and I think it's useful here to potentially identify dependencies that are introduced/changed with replace statements.

google/cloud/apigeeregistry/v1/admin_models.proto
timburks: Agreed, I was following the runtime/debug package and missed this. Changed, and this nicely improves the JSON-formatted output.

server/registry/internal/buildinfo/buildinfo-1.18.go
timburks: Thanks, fixed.

timburks:
@seaneganx I'm unsure what you meant by "Please remove buildinfo-prev.go" - is there a concern about using build tags to enable additional 1.18-only buildinfo? Admittedly this PR was initially a mess to review, but it's much better now with #415 merged.

seaneganx (approved):
I misunderstood the purpose of `buildinfo-prev.go`, thinking it was a previous version of the file accidentally checked into the PR. Now I see it's a different build command for earlier Go versions. LGTM

PR #418, opened 2022-01-06
Exclude child artifacts from the query when listing artifacts

### LINKED ISSUE #417: Registry API: Listing API artifacts returns artifacts that belong to the API's child deployments
Script to produce the issue:

```
apg admin create-project --project_id my-project
apg registry create-api --parent projects/my-project/locations/global --api_id my-api
apg registry create-api-deployment --parent projects/my-project/locations/global/apis/my-api --api_deployment_id my-deployment
apg registry create-artifact --parent projects/my-project/locations/global/apis/my-api/deployments/my-deployment --ar
tifact_id my-artifact
apg registry list-artifacts --parent projects/my-project/locations/global/apis/my-api
```

The expected result is an empty collection because the API doesn't have any artifacts. Instead, the artifact created under the child deployment resource is returned. To fix this, we need to exclude deployment artifacts from the query similar to [how we exclude version artifacts](https://github.com/apigee/registry/blob/29025b9b11ad5d7dd5fe06c899feba9d6a7f1cfa/server/registry/internal/storage/artifacts.go#L195).

seaneganx:
Child artifacts need to be excluded from the query or else they will be returned (incorrectly). Fixes #417.

The removed requirements aren't necessary because they exclude a subset of the resources already excluded by other requirements. For example, there is no need to require spec_id to be empty when we already require version_id to be empty.

The added requirements to ListSpecArtifacts and ListVersionArtifacts aren't necessary for correct behavior. Those are an optimization to reduce the number of artifacts needing to be filtered out in-memory before returning the collection of artifacts.


server/registry/internal/storage/artifacts.go
```diff
@@ -145,7 +147,6 @@ func (d *Client) ListVersionArtifacts(ctx context.Context, parent names.Version,
 
 func (d *Client) ListDeploymentArtifacts(ctx context.Context, parent names.Deployment, opts PageOptions) (ArtifactList, error) {
 	q := d.NewQuery(gorm.ArtifactEntityName)
-	q = q.Require("SpecID", "")
```
timburks, line 148: Do you think these superfluous Require()s are costing us anything? I think the performance impact is negligible, whereas knowing exactly when an extra Require() is beneficial (line 50) seems to require tricky analysis (at least I had to think a bit - and read down to line 97 - to see how it helped).

timburks (approved):
LGTM - some followup testing would be good for this but for now we could just track that with an issue.

server/registry/internal/storage/artifacts.go
seaneganx: I don't think there's any cost, so I put them back in.

PR #422, opened 2022-01-07
Remove the GAPIC patch that sends Authorization tokens with insecure requests.

timburks:
As discussed [here](https://github.com/apigee/registry/issues/52#issuecomment-1007822734), I think this patch is no longer needed.


timburks:
When this is merged, let's close #52.

seaneganx (approved):

PR #426, opened 2022-01-11
Conformance command: Add `mime_type` check

shrutiparabgoogle:
Logic to check mime_types in styleguide and match with mime_type of the spec

Fixes https://github.com/apigee/registry/issues/427


shrutiparabgoogle:
This is obsolete. Created this instead: https://github.com/apigee/registry/pull/430

PR #431, opened 2022-01-14
Follow AIP-135 guidance for cascading deletion

### LINKED ISSUE #414: Registry API: Add "force" field for cascading delete and AIP-135 conformance
According to [AIP-135](https://google.aip.dev/135#cascading-delete), deletions of resources with child resources should be denied with FAILED_PRECONDITION errors unless a boolean field named "force" is present and true in the request.

Currently the API implementation performs these "cascading deletions" by default. We think this should be changed to match the above requirements.

seaneganx:
* Add "force" field to Delete requests in protos
* Update expected protoc version to 3.19.3
* Regenerate files from updated protos
* Update test seeder to ignore nil resources
* Follow AIP-135 guidance for cascading deletion
* Use transactions for data consistency on deletion
* Add tests for cascading deletion

Fixes #414.


README.md
```diff
@@ -70,7 +70,7 @@ The entry point for the Registry API server itself is
 The following tools are needed to build this software:
 
 - Go 1.17 (recommended) or later.
-- protoc, the Protocol Buffer Compiler, version 3.18.1.
+- protoc, the Protocol Buffer Compiler, version 3.19.3.
```
timburks, line 73: Why the bump? Could we instead make changes like this in standalone PRs?

timburks (commented):
I'll look at this more closely after #428 is merged - could we bump the protoc version in a separate PR?

seaneganx:
>I'll look at this more closely after #428 is merged - could we bump the protoc version in a separate PR?

I opened #432 to update the protoc version. The reason I updated the version is because my machine's protoc version upgraded and I couldn't easily downgrade it to rebuild the protos.

PR #432, opened 2022-01-14
Update scripts to use protoc version 3.19.3

seaneganx:


timburks (approved):
Thanks for separating this - I noticed some minor churn in the generated .pb.go files, which I think might be because you might have used gofmt on the results instead of goimports... or possibly you didn't run either one. 

I think this only matters for the sake of minimizing churn, but I've been using goimports to normalize generated Go code. Maybe we should do this in our GENERATE scripts.

(not a blocking concern)

seaneganx:
Running `golangci-lint run --fix` uses the same config ([.golangci.yaml](https://github.com/apigee/registry/blob/main/.golangci.yaml)) as our GitHub merge precheck, but that's specifically configured to ignore our generated directories. That said, I didn't run any formatters on the generated code.

I'm fine with either formatting those or ignoring them, but let's update the config to match our expectation if we want it formatted.

PR #434, opened 2022-01-14
Follow AIP-135 guidance for cascading deletion

### LINKED ISSUE #414: Registry API: Add "force" field for cascading delete and AIP-135 conformance

seaneganx:
* Add "force" field to Delete requests in protos
* Regenerate files from updated protos
* Update test seeder to ignore nil resources
* Follow AIP-135 guidance for cascading deletion
* Use transactions for data consistency on deletion
* Add tests for cascading deletion

Fixes #414.


server/registry/internal/storage/delete.go
```diff
@@ -21,86 +21,171 @@ import (
...
+				return err
+			}
+
+			count += op.RowsAffected
 		}
-	}
 
-	return nil
+		if count > 1 && !cascade {
+			return status.Errorf(codes.FailedPrecondition, "cannot delete child resources in non-cascading mode")
```
timburks, line 50: Nice - assuming this aborts the transaction. Can we be sure that when only one row is affected, it has to correspond to the resource at the top of the hierarchy? (I think probably yes under normal conditions, but if something gets messed up, probably not)

timburks (approved):
Looks good. I like the implementation but realized that it depends on an invariant assumption that we make about the database: we assume that every resource's parent exists. Our creation and deletion functions maintain that, but if something outside of our control goes wrong, this could get messy. WDYT about adding a consistency checker to the admin API?

server/registry/internal/storage/delete.go
seaneganx: Before calling these database deletes we do an existence check on the resource in the RPC handlers. Unless something happens to the project between that transaction and this one, we know it exists at this point. Then unless gorm is incorrectly reporting affected rows, we know one of the affected rows is the deleted project.

I've been thinking about keeping the database operations for each API call in a single transaction (for a later PR). In that case, the first operation in the transaction would be the existence check.

timburks: SGTM, thanks for clarifying.

PR #440, opened 2022-01-19
Require "global" as the location when parsing project names with location

timburks:
We noticed that CreateAPI calls were accepting arbitrary location values, where all others required that the location be "global". In the future, we may expand our support to allow arbitrary locations, but for consistency, this corrects the regex used to parse located projects so that the location value must be "global".

Verified manually with:
```
apg registry create-api --parent projects/openapi/locations/invalid --api_id test
```


Krpate (approved):

timburks:
Merging - @seaneganx FYI

Issue #451, opened 2022-01-20
Make use of `exec.Command` more testable


shrutiparabgoogle:
With our current use of `exec.Command` there is no direct way to write unit tests. We have some functional tests which execute real shell commands, but we are limited here with what binaries are available to use during runtime in github automated tests.

This needs a refactoring of our use `exec.Command` right now. The plan is to create a common interface for shell commands so that we can pass a mock implementation while testing the execution of these commands. `exec.Command` is currently used for linters and controller. This refactor will help test both these use cases.

shrutiparabgoogle:
Created a draft PR to demonstrate the POC for testing `exec.Command` logic https://github.com/apigee/registry/pull/454
Handing off to @seaneganx for his efforts on overall testing strategy.

PR #452, opened 2022-01-20
Unit tests for conformance package

shrutiparabgoogle:
* Added unit tests for `conformance-task.go` and `linter.go`
* Moved some common test helpers `InitReport()` and `InitRuleReportGroups()` to `test-helpers.go` in package `conformance`.
* I have added some comments in the PR to explain the major changes happening in the code.

TODOs to follow:
* Tests for `invokeLinter()`. https://github.com/apigee/registry/issues/451


cmd/registry/cmd/compute/conformance_test.go
```diff
@@ -33,62 +33,9 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/testing/protocmp"
+	"github.com/apigee/registry/cmd/registry/conformance"
 )
 
-func initReport(t *testing.T) *rpc.ConformanceReport {
```
shrutiparabgoogle, line 38: Moved these helper functions to `test-helpers.go`

cmd/registry/conformance/conformance-task.go
```diff
@@ -107,40 +110,77 @@ func (task *ComputeConformanceTask) Run(ctx context.Context) error {
...
-	err = task.computeConformanceReport(ctx, root, conformanceReport)
-	if err != nil {
-		log.Errorf(ctx,
-			"Failed to compute the conformance for spec %s: %s",
-			task.Spec.GetName(),
-			err,
-		)
-		return err
+	guidelineReportsMap := make(map[string]int)
+	for _, metadata := range task.LintersMetadata {
```
shrutiparabgoogle, line 116: Moved the for loop out of `computeConformanceReport` out here. This was done for better testability. We can now pass different combinations of `linterResponse` and `metadata` to `computeConformanceReport` and have unit test coverage.

cmd/registry/conformance/conformance-task.go
```diff
@@ -107,40 +110,77 @@ func (task *ComputeConformanceTask) Run(ctx context.Context) error {
...
+		}
+
+		task.computeConformanceReport(ctx, conformanceReport, guidelineReportsMap, linterResponse, metadata)
 	}
 
 	return task.storeConformanceReport(ctx, conformanceReport)
 }
 
-func (task *ComputeConformanceTask) computeConformanceReport(
+func (task *ComputeConformanceTask) invokeLinter(
```
shrutiparabgoogle, line 133: Moved `invokeLinter` from `linter.go` to here. Making invokeLinter a function of `ComputeConformanceTask` will open the possibility to mock this function while testing the `Run()` function of this task.

cmd/registry/conformance/conformance-task.go
```diff
@@ -162,14 +202,17 @@ func (task *ComputeConformanceTask) computeConformanceReport(
 
 				// Check if the guideline report for the guideline which contains this rule
 				// has already been initialized. If it hasn't then create one.
-				guidelineReport, ok := guidelineReportsMap[guideline.GetId()]
+				reportIndex, ok := guidelineReportsMap[guideline.GetId()]
```
shrutiparabgoogle, line 205: `guidelineReportsMap` now stores only the index of the `guidelineReport` in the `guidelineGroup` and not the whole object itself.

Example:
```
conformanceReport{
    ...
    GuidelineReportGroups: []*rpc.GuidelineReportGroup{
         &rpc.GuidelineReportGroup{...Status:UNSPECIFIED...},
         &rpc.GuidelineReportGroup{...Status:PROPOSED...},
         &rpc.GuidelineReportGroup{
             ...
             Status: "ACTIVE",
             GuidelineReports: []*rpc.GuidelineReport{
                 &rpc.GuidelineReport{GuidelineId: "test-guideline-id-1", ...}},
                 &rpc.GuidelineReport{GuidelineId: "test-guideline-id-2", ...}},
             ...
        }
    }
}
```

The `guidelineReportsMap` will store the index and not the whole `guidelineReport` object. The map will look like this for the above example:
```
"test-guideline-id-1": 0
"test-guideline-id-2": 1
```
We need to know the position of that particular `guidelineReport` in the corresponding status group. Then append the new `RuleReports` under that particular guideline.

shrutiparabgoogle:
@timburks to review conformance logic and test cases.
@seaneganx for reviewing tests for best practices.

cmd/registry/conformance/conformance-task.go
shrutiparabgoogle: I might be mistaken about my understanding of mocking in golang,  it works differently than my understanding from python background.

timburks (approved):

seaneganx (commented):
It looks like the task tests are missing a lot of non-trivial work being done because of our direct call to `computeConformanceReport` instead of testing the task using its `Run` function. Aside from the external linter problem, are there any other reasons for this approach?

This approach also has the cost of tightly coupling our tests with this particular implementation of the task (we can't modify `computeConformanceReport` without updating all of our tests).

cmd/registry/conformance/conformance-task_test.go
```diff
@@ -0,0 +1,561 @@
+package conformance
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

cmd/registry/conformance/linter_test.go
```diff
@@ -0,0 +1,282 @@
+package conformance
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

cmd/registry/conformance/conformance-task_test.go
```diff
@@ -0,0 +1,561 @@
...
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/apigee/registry/rpc"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"google.golang.org/protobuf/testing/protocmp"
+	// "github.com/apigee/registry/connection"
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

cmd/registry/conformance/linter_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+			if test.wantErr {
+				if err == nil {
+					t.Errorf("Expected GenerateLinterMetadata() to return an error")
+				}
+			} else {
+				if err != nil {
+					t.Errorf("Unexpected error from GenerateLinterMetadata(): %s", err)
+				}
+			}
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

cmd/registry/conformance/linter_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+			opts := cmp.Options{
+				protocmp.Transform(),
+				cmpopts.SortSlices(func(a, b string) bool { return a < b }),
+				cmp.AllowUnexported(linterMetadata{}),
+				cmp.AllowUnexported(ruleMetadata{}),
+				cmpopts.SortMaps(func(a, b string) bool { return a < b }),
+			}
+			if !cmp.Equal(test.wantMetadata, gotMetadata, opts) {
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done

cmd/registry/cmd/compute/conformance_test.go
shrutiparabgoogle: I started with the normal struct approach, constructing each testcase got super long since we had to initialize all the `status` and `severity` values. This was my attempt in avoiding that repeated initialization code and maintain only what is required in the testcases. 

However, I realize now that this approach is making it more complicated and affecting the readability of the tests. I can revert to my original method.

cmd/registry/conformance/conformance-task.go
shrutiparabgoogle: [This](https://github.com/apigee/registry/pull/454) is my plan for testing `exec.Command` usage in. our codebase.  
The PR. is still WIP, it currently contains tests for package `controller`, I am planning to add tests for `invokeLinter` in the similar way.

We can defer this discussion of testing `invokeLinter()` to that PR.

cmd/registry/conformance/test-helpers.go
```diff
@@ -0,0 +1,61 @@
+package conformance
```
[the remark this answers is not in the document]
shrutiparabgoogle: Not valid anymore.

cmd/registry/conformance/test-helpers.go
```diff
@@ -0,0 +1,61 @@
...
+			},
+			{
+				Status:           rpc.Guideline_DISABLED,
+				GuidelineReports: []*rpc.GuidelineReport{},
+			},
+		},
+	}
+}
+
+func InitRuleReportGroups(t *testing.T) []*rpc.RuleReportGroup {
```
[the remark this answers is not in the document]
shrutiparabgoogle: Not valid anymore.

cmd/registry/conformance/test-helpers.go
```diff
@@ -0,0 +1,61 @@
+package conformance
+
+import (
+	"testing"
+
+	"github.com/apigee/registry/rpc"
+)
+
+func InitReport(t *testing.T) *rpc.ConformanceReport {
+	t.Helper()
```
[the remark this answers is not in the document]
shrutiparabgoogle: Not valid anymore.

cmd/registry/cmd/compute/conformance_test.go
```diff
@@ -117,11 +64,11 @@ func TestConformance(t *testing.T) {
 			conformancePath: filepath.Join("testdata", "styleguide.yaml"),
 			getPattern:      "projects/conformance-test/locations/global/apis/petstore/versions/1.0.0/specs/openapi.yaml/artifacts/conformance-openapitest",
 			wantProto: func() *rpc.ConformanceReport {
-				conformance := initReport(t)
-				conformance.Name = "projects/conformance-test/locations/global/apis/petstore/versions/1.0.0/specs/openapi.yaml/artifacts/conformance-openapitest"
-				conformance.StyleguideName = "openapitest"
+				report := conformance.InitReport(t)
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done, got rid of test-helpers.go

```
Why did we choose to use a fixed length GuidelineReportGroups list with enum-specific indices instead of having one field for each status group? The latter seems like a much less complicated way to represent the same structure.
```
I agree with your point. I am not completely aware of what factors were into play while designing the original protos. Can we have a separate discussion on this, it feels out of scope for this PR.

cmd/registry/conformance/conformance-task_test.go
```diff
@@ -0,0 +1,561 @@
...
+func TestInitializeConformanceReport(t *testing.T) {
+	specName := specName
+	styleguideId := styleguideId
+	want := &rpc.ConformanceReport{
+		Name:           fmt.Sprintf("%s/artifacts/conformance-%s", specName, styleguideId),
+		StyleguideName: styleguideId,
+		GuidelineReportGroups: []*rpc.GuidelineReportGroup{
+			{
+				Status:           rpc.Guideline_STATUS_UNSPECIFIED,
+				GuidelineReports: make([]*rpc.GuidelineReport, 0),
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

cmd/registry/conformance/conformance-task_test.go
```diff
@@ -0,0 +1,561 @@
...
+
+// This test will catch any changes made to the original severity values.
+func TestInitializeGuidelineReport(t *testing.T) {
+	guidelineId := styleguideId
+	want := &rpc.GuidelineReport{
+		GuidelineId: guidelineId,
+		RuleReportGroups: []*rpc.RuleReportGroup{
+			{
+				Severity:    rpc.Rule_SEVERITY_UNSPECIFIED,
+				RuleReports: make([]*rpc.RuleReport, 0),
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

cmd/registry/cmd/compute/conformance_test.go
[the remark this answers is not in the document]
seaneganx: Right, I agree that shouldn't be changing in this PR.

cmd/registry/cmd/compute/conformance_test.go
shrutiparabgoogle: Not valid anymore

shrutiparabgoogle:
> It looks like the task tests are missing a lot of non-trivial work being done because of our direct call to `computeConformanceReport` instead of testing the task using its `Run` function. Aside from the external linter problem, are there any other reasons for this approach?

The tests for `conformance-tasks` are still incomplete. If you look at the structure of `conformance-task.go`, 
- `Run()` is the main method of the command which starts execution of the task.
- Inside `Run()`, there are calls to `invokeLinter()` and `computeConformanceReport()`. This PR includes tests for `computeConformanceReport()`.
- The goal here is to add test cases for both `computeConformanceReport()` and `invokeLinter()`, which cover all the code paths in those functions (Success and Failure cases). Once we have enough coverage for both of these functions, testing `Run()` becomes simpler, we have to test `Run()` only based on what result these two functions return, tests for `Run()` will essentially cover how it is handling failures (and success) from `computeConformance()` and `invokeLinter()`.

I should have mentioned earlier that these tests are not complete yet, this PR was already quiet long, so I will be adding tests in iterations. Next PRs will include tests for `invokeLinter()` and then `Run()`.

> 
> This approach also has the cost of tightly coupling our tests with this particular implementation of the task (we can't modify `computeConformanceReport` without updating all of our tests).

Isn't that expected with unit tests? Everytime we change the logic for `computeConformanceReport()`, the tests need to be updated for new code paths. These tests are also a way of documenting code behavior, they will help understand the behavior of the code for anyone new who interacting with the codebase.

seaneganx:
Thanks for the clarifications.

>tests for `Run()` will essentially cover how it is handling failures (and success) from `computeConformance()` and `invokeLinter()`.

Does this mean the `Run()` tests will be using some kind of test double to create the situations where it's handling fail/success from those private methods?

>>This approach also has the cost of tightly coupling our tests with this particular implementation of the task (we can't modify computeConformanceReport without updating all of our tests).

>Isn't that expected with unit tests? Everytime we change the logic for computeConformanceReport(), the tests need to be updated for new code paths.

I think it's expected that the interface will be tightly coupled with the tests. My concern is about tight coupling to the current implementation. Private/unexported methods like `invokeLinter` are much more likely to be replaced or removed in the future than behaviors and interface for `ComputeConformanceTask.Run`.

Generally speaking, I think this comes down to the pros and cons of testing private/unexported code. Searching online you will find mixed opinions ([Stack Overflow](https://softwareengineering.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods)) about whether or not to do it. Conveniently I just realized we made the Software Engineering at Google book freely available online. See [Chapter 12](https://abseil.io/resources/swe_at_google.2.pdf#page=259) for unit testing guidance.

shrutiparabgoogle:
> Thanks for the clarifications.
> 
> > tests for `Run()` will essentially cover how it is handling failures (and success) from `computeConformance()` and `invokeLinter()`.
> 
> Does this mean the `Run()` tests will be using some kind of test double to create the situations where it's handling fail/success from those private methods?

We can test `Run()` with specific inputs which will cause one of those private methods to fail/succeed.

For example, how `Run()` is tested in this PR: 
```
func (task *ExecCommandTask) Run(ctx context.Context) error {
	// The monitoring metrics/dashboards are built on top of the format of the log messages here.
	// Check the metric filters before making any changes to the format.
	// Location: registry/deployments/controller/dashboard/*
	logger := log.FromContext(ctx).WithFields(map[string]interface{}{
		"action": fmt.Sprintf("{%s}", task.Action.Command),
		"taskID": fmt.Sprintf("{%s}", task.TaskID),
	})

	err := task.executeCommand(ctx, logger)
	if err != nil {
		return err
	}

	logger.Debug("Successful Execution:")
	return nil
}
```

The `Run()` function makes a call to private method `task.executeCommand(ctx, logger)` and it is tested as follows: https://github.com/apigee/registry/pull/454/files#diff-509ea8a8192e6c1c5a56974f0e47c8bcb103951cdfca6e95d9b4d9601365dbc6R66

This is just a proposal and still open to discussion on what might be the best approach.

cmd/registry/conformance/linter_test.go
```diff
@@ -0,0 +1,292 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+
+			got, err := GenerateLinterMetadata(test.styleguide)
+
+			if test.wantErr && err == nil {
+				t.Errorf("Expected GenerateLinterMetadata() to return an error")
```
seaneganx, line 275: Should this be `Fatalf`? If we expected an error and didn't find one, the test case doesn't seem like it's set up to make a useful diff later in the test.

cmd/registry/conformance/linter_test.go
```diff
@@ -0,0 +1,292 @@
...
+import (
+	"testing"
+
+	"github.com/apigee/registry/rpc"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+var rules = []*rpc.Rule{
```
seaneganx, line 26: Can we declare these separately in a var block with descriptive names for each rule? As long as they're only being used as "drop-in" values (whose contents don't actually matter) I think it's fine that they're not in-lined in the test. Still, it would be helpful to have a general idea what `rule[0]` actually is, e.g. `noRefSiblingsRule`.

cmd/registry/conformance/conformance-task_test.go
```diff
@@ -0,0 +1,601 @@
...
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+const specName = "projects/demo/locations/global/apis/petstore/versions/1.0.0/specs/openapi.yaml"
+const styleguideId = "openapi-test"
+
+// This test will catch any changes made to the original status values.
+func TestInitializeConformanceReport(t *testing.T) {
+	specName := specName
+	styleguideId := styleguideId
```
seaneganx, line 34: Do these serve a purpose? It seems unnecessary.

seaneganx (approved):
Some final comments.

cmd/registry/conformance/linter_test.go
shrutiparabgoogle: Makes sense.

cmd/registry/conformance/linter_test.go
shrutiparabgoogle: Done.

cmd/registry/conformance/conformance-task_test.go
shrutiparabgoogle: Done.

PR #457, opened 2022-02-01
Add tests and decompression support for GetArtifactContents 

seaneganx:
I noticed we didn't have tests for this RPC method, so I copied the cases from `TestGetApiSpecContents` and adjusted them for the artifact resource type. One of the cases failed, which led me to find that `GetArtifactContents` wasn't decompressing gzipped artifact contents as it [claims to support](https://github.com/apigee/registry/blob/61c5bc7df3a3ddbdb9587ae184b5d26a2931b39e/google/cloud/apigeeregistry/v1/registry_service.proto#L328). This PR adds that functionality and some tests for the method.

### DIFF
--- a/server/registry/actions_artifacts.go
+++ b/server/registry/actions_artifacts.go
@@ -17,6 +17,7 @@ package registry
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/apigee/registry/rpc"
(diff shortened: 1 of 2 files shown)


timburks (approved):
This is great, thanks for catching this @seaneganx !

PR #467, opened 2022-02-15
Replace "reference implementation" with "core implementation" in README.

timburks:
Also include results of running tools/pretty.sh to format all READMEs.


earth2marsh (approved):
lgtm

Issue #464, opened 2022-02-09
Registry controller: Support advanced patterns in manifest definition


shrutiparabgoogle:
**Current support:**
manifest definition currently supports patterns which use the same level of matching in dependencies and action strings.
It enforces two conditions as follows:
- All of the dependencies should start with the same level of `$resource.{api|version|spec|artifact}` pattern or none.
   Valid dependency patterns:
   `$resource.spec/artifacts/complexity`, `$resource.spec/artifacts/lint-spectral`
   `apis/-/versions/-/specs`, `apis/-/versions/-`
   However, a combination of the above two patterns is considered invalid:
   `$resource.spec` `artifacts/apihub-styleguide`
   This pattern needs support now that we have incorporated styleguide definitions. 

**Desired support:**
The desired manifest entry for using styleguide definitions with controller looks as follows:
```
pattern: "apis/-/versions/-/specs/-/artifacts/conformance-receipt"
receipt: true
dependencies:
- pattern: "$resource.spec"
- pattern: "artifacts/apihub-styleguide"
action: "registry compute conformance $resource.spec"
```

Make modifications in the [pattern validation](https://github.com/apigee/registry/blob/main/cmd/registry/controller/parser.go#L191) step and possibly controller logic to support such patterns.

shrutiparabgoogle:
A valid pattern can be defined as follows:
1. There can be a combination of the presence and absence of $resource references in the dependencies or action string.
2. The generated resource pattern should always end with a resource name. It cannot end with a “-”.
3. There cannot be references to entities which are not present in the original resource pattern. Eg: for a resource pattern apis/-/versions/-/specs/- a $resource.artifact reference cannot be used anywhere.

PR #471, opened 2022-03-04
Controller: Use parent to generate names for non-existent target resources.

shrutiparabgoogle:
Partially Fixes https://github.com/apigee/registry/issues/464

The validation pattern will be implemented in upcoming PRs (https://github.com/apigee/registry/pull/477)

Summary of changes:
* Change logic to use parent while generating names for non-existent target resources.
* Some refactoring in `generateActions`.


cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,218 @@ func generateActions(
...
-			}
+	return actions, nil
+}
+
+func evaluateAction(
+	targetResourceName ResourceName,
+	targetResourceTime time.Time,
+	dependencyMaps []map[string]time.Time,
+	generatedResource *rpc.GeneratedResource,
+	createMode bool) (*Action, error) {
```
seaneganx, line 313: This seems like it should be split into separate functions for each supported "mode" because 1) it serves a different purpose in each mode, and 2) callers need to provide an arbitrary `targetResourceTime` argument to be ignored in create mode.

At first glance, it seems like create mode is:
* Get the entity key for each dependency.
* Validate that each dependency is in the dependency maps
* Generate an action for the resource.

and update mode is:
* Get the entity key for each dependency.
* Validate that each dependency is in the dependency maps.
* Generate an action for the resource if and only if it's stale.

These seem like they can be split into separate functions that are individually much simpler.

seaneganx (commented):
Some preliminary comments. I haven't looked much at the tests or thought much about potential bugs. I don't really understand what's happening with the "visited" set.

Is it skipping the check for missing artifacts if we update any artifacts under the same parent?  I can imagine a scenario where a resource needs some artifacts updated and others created.

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,218 @@ func generateActions(
...
+		action := &Action{
+			Command:           cmd,
+			GeneratedResource: targetResourceName.String(),
+			RequiresReceipt:   generatedResource.Receipt,
+		}
+		return action, nil
 	}
 
-	return actions, nil
+	return nil, nil
```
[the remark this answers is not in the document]
shrutiparabgoogle: Thanks for this suggestion. Done.

shrutiparabgoogle:
> Some preliminary comments. I haven't looked much at the tests or thought much about potential bugs. I don't really understand what's happening with the "visited" set.
> 
> Is it skipping the check for missing artifacts if we update any artifacts under the same parent? I can imagine a scenario where a resource needs some artifacts updated and others created.

Since we have now decided (to be implemented in next PR) to fix the resource name for `generatedResource` and not allow `-`, each manifest entry will create/update only one artifact for a parent. Example:
```
generated_resources:
- pattern: "apis/-/versions/-/specs/-/artifacts/complexity"
   dependencies:
   - pattern: "$resource.spec"
   action: "registry compute complexity $resource.spec"
- pattern: "apis/-/versions/-/specs/-/artifacts/spectral-lint"
   dependencies:
   - pattern: "$resource.spec"
   action: "registry compute lint $resource.spec --linter spectral"
```

In the above example, the first entry is looking at only `complexity` artifacts whereas the second entry is looking for only `spectral-lint` artifacts under a spec. As for the visited map, the visited parents map is maintained per manifest entry (each entry in generated_resources). Hence, the parents which are marked as visited in the first entry will not affect the parents in the second entry. Thus, even if a `complexity` artifact was generated for a spec `apis/petstore/versions/1.0.0/specs/openapi.yaml`, this spec will still be checked for `lint` artifacts when the controller starts processing the next entry.

cmd/registry/controller/controller.go
```diff
@@ -61,30 +62,21 @@ func processManifestResource(
 	ctx context.Context,
 	client connection.Client,
 	projectID string,
-	resource *rpc.GeneratedResource) ([]*Action, error) {
+	generatedResource *rpc.GeneratedResource) ([]*Action, error) {
```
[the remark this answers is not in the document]
seaneganx: Yeah `resource` wasn't great either. I called this a nit because I don't really have any ideas aside from `gr` but that might be too short.

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,249 @@ func generateActions(
...
+		}
+
+		// All the dependencies should be present to generate an action.
+		if maxUpdateTime, ok := dMap[entityKey]; ok {
+			if maxUpdateTime.After(targetResourceTime) {
+				takeAction = true // Take action if dependency timestamp is later than resource timestamp
+			}
+		} else {
+			takeAction = false
+			break
```
seaneganx, line 356: This should `return false` directly for clarity.

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,249 @@ func generateActions(
...
+		entityKey, err := getEntityKey(dependency.Pattern, targetResourceName)
+		if err != nil {
+			// This means that there is error in the pattern definition, hence return
+			return false, fmt.Errorf("cannot match resource with dependency. Error: %s", err.Error())
+		}
+
+		// All the dependencies should be present to generate an action.
+		if maxUpdateTime, ok := dMap[entityKey]; ok {
+			if maxUpdateTime.After(targetResourceTime) {
+				takeAction = true // Take action if dependency timestamp is later than resource timestamp
```
seaneganx, line 352: I think this should be `return true` but that changes the functionality of this loop because we're combining the validation (are all dependencies are present?) with the check itself (are any dependencies fresher than the target?). Can we separate the validation step?

IMO the validation should either have its own function, or at least its own loop in this one. The dependency lists are small and everything is in-memory so I have no concern about the performance of two loops instead of one.

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,249 @@ func generateActions(
...
+func needsUpdate(
+	targetResourceName ResourceName,
+	targetResourceTime time.Time,
+	dependencyMaps []map[string]time.Time,
+	generatedResource *rpc.GeneratedResource,
+	createMode bool) (bool, error) {
+	takeAction := false
+	//Check if all the dependencies exist in the map.
+	for i, dependency := range generatedResource.Dependencies {
+		dMap := dependencyMaps[i]
```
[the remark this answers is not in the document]
shrutiparabgoogle: I agree that this relationship should be clearly visible through code. 
Currently, the "magic" is done in `processManifestResource()`, where for each dependency pattern, there is a map maintained in the same order with entities as keys.

Example:
```
dependencies:
- pattern: "$resource.spec"
- pattern: "$resource.version/artifacts/deployed-spec"
```

The corresponding maps will be as follows:
```
dependencyMaps := []*map(string,time){
  map(string, time){
    "apis/petstore/versions/1.0.0/specs/openapi.yaml": "00:00:00",
    "apis/petstore/versions/1.0.1/specs/openapi.yaml": "00:00:00"
  },
 map(string, time){
    "apis/petstore/versions/1.0.0": "00:00:00",
    "apis/petstore/versions/1.0.1": "00:00:00"
  },
}
```

cmd/registry/controller/controller.go
shrutiparabgoogle: I would prefer to keep these two checks in the same loop since it will be easier to understand all the validation conditions in one place. 

Refactored to separate the two checks so that it's more readable and less confusing.

cmd/registry/controller/controller.go

cmd/registry/controller/controller.go
```diff
@@ -244,7 +241,7 @@ func generateCreateActions(
 	if strings.HasSuffix(parentName, "locations/global") {
 		// Return if this parent was already visited.
 		if visited[parentName] {
-			return actions, nil
+			return nil, nil //[]*Action{}, nil
```
seaneganx, line 244: Nit: Given how [common](https://go.dev/doc/effective_go#slices) `nil` slices are in Go, I don't think this needs a comment. Otherwise I think we should just return an empty slice instead of adding a comment explaining the `nil` slice.

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,246 @@ func generateActions(
...
+		entityKey, err := getEntityKey(dependency.Pattern, targetResourceName)
+		if err != nil {
+			// This means that there is error in the pattern definition, hence return
+			return false, fmt.Errorf("cannot match resource with dependency. Error: %s", err.Error())
+		}
+
+		// All the dependencies should be present to generate an action.
+		if _, ok := dMap[entityKey]; !ok {
+			takeAction = false
+			break
```
seaneganx, line 377: Another case to `return false`. We don't need the `takeAction` variable in this function.

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,246 @@ func generateActions(
...
 	return actions, nil
 }
+
+func needsUpdate(
+	targetResourceName ResourceName,
+	targetResourceTime time.Time,
+	dependencyMaps []map[string]time.Time,
+	generatedResource *rpc.GeneratedResource,
+	createMode bool) (bool, error) {
+	//Check if all the dependencies exist in the map.
```
seaneganx, line 336: The loop does more than just this. IMO we should remove this (and the similar one in `needsCreate`) unless there's something to say that isn't clear from the loop code/comments.

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,246 @@ func generateActions(
...
+
+		takeAction, err := needsCreate(
+			targetResourceName,
+			dependencyMaps,
+			generatedResource,
+		)
+
+		if err != nil {
+			log.Errorf(ctx, "%s", err)
+			continue
```
seaneganx, line 310: Why don't we return these errors? In the special case we return them. It seems like these only happen if the manifest has invalid fields, so shouldn't we fail in that case?

cmd/registry/controller/controller.go
```diff
@@ -144,107 +136,249 @@ func generateActions(
...
+
+	// If parent is a project, we can't list projects since this is registry client command.
+	// Since the manifest definition is scoped  only for a particular project,
+	// there will be only one target resource in this case.
+	// There are two cases where this might happen:
+	// 1. Target resource is a project level artifact "projects/demo/locations/global/artifacts/serach-index"
+	//    extracted parent will be "projects/demo/locations/global"
+	// 1. Target resource is an api "projects/demo/locations/global/apis/petstore"
+	//    extracted parent will be "projects/demo/locations/global"
+	if strings.HasSuffix(parentName, "locations/global") {
```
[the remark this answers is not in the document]
shrutiparabgoogle: One more future optimization in my mind is to converge the `updateActions` and `createActions` methods, where we fetch the parent only if the target resource is absent and not go through all the parents.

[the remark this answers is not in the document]
seaneganx: Suppose this is listing a few thousand specs. Will there be a perceptible difference in performance by looping over it again?

There are other much slower operations (e.g. unnecessary server requests) that we can avoid but choose not to because it's fast enough as-is. Here it matters much less, so I think we should favor code clarity and maintainability.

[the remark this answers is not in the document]
shrutiparabgoogle: Yes, looping over twice is a concern but I don't have any proof regarding the performance impact. I am completely in favor of code clarity and maintainability, I am just saying that I would like to refactor this in a separate effort where I can look for options to converge both the `parent` data types and also converge the `create` and `update` functions. Hence, the tracking issue: https://github.com/apigee/registry/issues/485

Unrelated, but for multiple server requests, the optimization plan is to introduce caching. Hence, the code right now is doing multiple calls to the registry with exactly the same query, but it should be solved with caching.

seaneganx (approved):

PR #478, opened 2022-03-04
When testing GetStatus, exclude Go 1.18+ fields.

timburks:
The "Build" field is only returned for Go 1.18. When it is returned from a server built with Go 1.18, the test fails unless we exclude this field. It also contains compiler-dependent values, so we should probably always exclude it.


seaneganx (changes requested):

server/registry/actions_status_test.go
```diff
@@ -34,6 +34,7 @@ func TestGetStatus(t *testing.T) {
 	}
 
 	got, err := server.GetStatus(ctx, req)
+	got.Build = nil // remove values that are not returned by server builds < Go 1.18
```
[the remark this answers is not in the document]
timburks: Updated.

server/registry/actions_status_test.go
```diff
@@ -38,7 +38,13 @@ func TestGetStatus(t *testing.T) {
...
 
-	if !cmp.Equal(want, got, protocmp.Transform()) {
+	opts := cmp.Options{
+		protocmp.Transform(),
+		// Ignore fields that are build-dependent or excluded from Go 	< 1.18.
+		protocmp.IgnoreFields(&rpc.Status{}, "build"),
+	}
+
+	if !cmp.Equal(want, got, opts) {
 		t.Errorf("GetStatus(%+v) returned unexpected diff (-want +got):\n%s", req, cmp.Diff(want, got, protocmp.Transform()))
```
seaneganx, line 48: Did you intentionally use different options for the diff and equal? IMO they should use the same options.

seaneganx (approved):

server/registry/actions_status_test.go
timburks: No, thanks for catching that. Will fix.

PR #479, opened 2022-03-07
Update Registry-linter docker file to use a specific version of spectral

giteshk:
Additionally added compiled proto file to Registry server container


containers/registry-server/Dockerfile
```diff
@@ -43,6 +45,7 @@ RUN apk add --no-cache ca-certificates
 
 # Copy the binary to the production image from the builder stage.
 COPY --from=builder /app/registry-server /registry-server
+COPY --from=builder /app/deployments/envoy/proto.pb /proto.pb
```
timburks, line 48: Why is this needed here? Usually it's an input to envoy, but I don't think there's an envoy instance in this container. (is there?)

giteshk: This is needed when we test the setup using istio and have a sidecar proxy for grpc transcoding.

timburks: So do you have the sidecar read this out of the container?

timburks (approved):

shrutiparabgoogle (approved):

containers/registry-server/Dockerfile
giteshk: When I look at the file for the hosted solution. The files are copied into an empty volume that is shared between the running registry server pod and the istio filter.

PR #480, opened 2022-03-07
Benchmark test for CRUD operations on Api/Api Version/Api spec/Artifact

giteshk:
The benchmark tests can be used against any running registry service 

To run it against a local running service you can run

```
source auth/LOCAL.sh
export REGISTRY_PROJECT_IDENTIFIER=bench
apg admin create-project --project_id=$REGISTRY_PROJECT_IDENTIFIER
go test --bench=Create ./tests/benchmark --benchtime=10x
go test --bench=Update ./tests/benchmark --benchtime=10x
go test --bench=Get ./tests/benchmark --benchtime=10x
go test --bench=List ./tests/benchmark --benchtime=10x
go test --bench=Delete ./tests/benchmark --benchtime=10x
```

If you wish to create 5 versions for every API set `export REGISTRY_VERSION_COUNT=5`

To run this test against hosted version
```
source auth/HOSTED.sh
export REGISTRY_PROJECT_IDENTIFIER=<Registry project name>
apg admin create-project --project_id=$REGISTRY_PROJECT_IDENTIFIER
go test --bench=Create ./tests/benchmark --benchtime=10x
go test --bench=Update ./tests/benchmark --benchtime=10x
go test --bench=Get ./tests/benchmark --benchtime=10x
go test --bench=List ./tests/benchmark --benchtime=10x
go test --bench=Delete ./tests/benchmark --benchtime=10x
```


tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+}
+
+func getNumberOfVersions() int {
+	strNumberOfVersions, exists := os.LookupEnv("REGISTRY_VERSION_COUNT")
+	numberOfVersions, err := strconv.Atoi(strNumberOfVersions)
+	if err != nil || !exists {
+		numberOfVersions = 1
+	}
+	return numberOfVersions
+}
```
seaneganx, line 49: Nit: Function names starting with "get" aren't very common in Go. Usually functions like `getFoo` are shortened to just `foo`.

In cases like this where the value isn't readily available, the function name can usually include a different verb to indicate it's coming from somewhere else, e.g. `fetchFoo`, `generateBar`, etc. For this particular case I recommend something like `loadVersionCount`.

**More importantly**: I think we should fail the test if the input count is invalid to communicate the problem to the user. We want them to know we're not proceeding with whatever they tried to set the value to.

```
func loadVersionCount(t *testing.T) {
	t.Helper()

	s, ok := os.LookupEnv("REGISTRY_VERSION_COUNT")
	if !ok {
		return 1 // Default to a single version.
	}

	c, err := strconv.Atoi(s)
	if err != nil {
		t.Fatalf("Setup: Invalid REGISTRY_VERSION_COUNT %q: count must be an integer", s)
	}

	return c
}
```

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+func getRootResource() string {
```
seaneganx, line 34: These functions should accept a parameter `b *testing.B` so they can be marked as test helpers and fail the test if necessary.

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+		apiNamePrefix = "benchtest"
+	}
+	apiNameStartOffset, valueExists := os.LookupEnv("API_NAME_START_OFFSET")
+	if valueExists {
+		intVar, _ := strconv.Atoi(apiNameStartOffset)
+		i += intVar
+	}
+
+	return fmt.Sprintf("%s-%d", apiNamePrefix, i)
+}
```
seaneganx, line 63: Why do we need these to be customizable?

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+
+func getApiName(i int) string {
+	apiNamePrefix, valueExists := os.LookupEnv("API_NAME_PREFIX")
+	if !valueExists {
+		apiNamePrefix = "benchtest"
+	}
+	apiNameStartOffset, valueExists := os.LookupEnv("API_NAME_START_OFFSET")
+	if valueExists {
+		intVar, _ := strconv.Atoi(apiNameStartOffset)
```
seaneganx, line 58: This is another case where I think we should fail the test when the input is invalid, rather than defaulting to something that is unlikely to match the user's intentions.

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+	client := gapicClient.GrpcClient()
+
+	b.StopTimer()
+	b.ResetTimer()
+
+	for i := 1; i <= b.N; i++ {
+		if i%1000 == 0 {
+			//Sleep 10s every 1000 iterations to reduce server load
+			time.Sleep(10 * time.Second)
+		}
```
seaneganx, line 132: Why do we need to reduce server load? Why does it need 10s to do that, as opposed to something much quicker like 1s, 100ms, etc?

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+				ApiVersion: &rpc.ApiVersion{
+					DisplayName: fmt.Sprintf("v%d", j),
+					Description: fmt.Sprintf("v%d of %s", j, apiId),
+				},
+				ApiVersionId: fmt.Sprintf("v%d", j),
+			}
+			// Only take reading for the first version
+			if j == 1 {
+				b.StartTimer()
+			}
```
seaneganx, line 147: Why do we only measure the first version? Why do we continue creating other versions after that?

.github/workflows/composite/benchmark/action.yml
```diff
@@ -0,0 +1,71 @@
...
+
+    - name: Install
+      run: go install ./...
+      shell: bash
+
+    - name: Start the Registry Server with Postgres
+      run: registry-server &
+      shell: bash
+      env:
+        REGISTRY_DATABASE_CONFIG: ${{ inputs.dbconnection }}
```
seaneganx, line 46: When you run the server without a config file it uses a [default configuration](https://github.com/apigee/registry/blob/40b8c96b63f88fd908eb77f98b919eab0f8a9370/cmd/registry-server/main.go#L76) that connects to a SQLite database. If we need this to run against a different database you will need to create a config file that references this environment variable.

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+		if i%1000 == 0 {
+			//Sleep 10s every 1000 iterations to reduce server load
+			time.Sleep(10 * time.Second)
+		}
+
+		apiId := getApiName(i)
+
+		for j := 1; j <= getNumberOfVersions(); j++ {
+
+			buf, err := readAndGZipFile(filepath.Join("testdata", "openapi1.yaml"))
```
seaneganx, line 180: This seems unnecessarily expensive. With some of our examples this file will be re-read from disk 10,000 times during the benchmark. Why not read it once at the start?

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+	zw, _ := gzip.NewWriterLevel(&buf, gzip.BestCompression)
+	_, err := zw.Write(fileBytes)
+	if err != nil {
+		return nil, err
+	}
+	if err := zw.Close(); err != nil {
+		return nil, err
+	}
+	return &buf, nil
+}
```
seaneganx, line 79: Can we gzip the files that are being checked into the `testdata/` directory instead? Doing it here seems like unnecessary work.

seaneganx (commented):
There are a lot of new features in here that I don't understand. Can you expand the documentation to explain each feature and why it's useful?

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,895 @@
...
+	"github.com/apigee/registry/rpc"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+func getRootResource() string {
+	registryProject, valueExists := os.LookupEnv("REGISTRY_PROJECT_IDENTIFIER")
+	if valueExists {
+		return fmt.Sprintf("projects/%s/locations/global", registryProject)
+	}
+	return "projects/bench/locations/global"
```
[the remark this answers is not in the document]
giteshk: I addressed this.

tests/benchmark/crud_test.go
giteshk: Say I ran the create script and created 100 APIs. 
I wanted to add 100 more APIs to an existing environment without having to delete and rerun the script with 200 APIs

This was Needed when I started seeing errors while trying to adding 2000 APIs.
so I could break down the create into smaller chunks

tests/benchmark/crud_test.go
giteshk: fixed

tests/benchmark/crud_test.go
giteshk: I found the script would error out if I ran it to create 2000 APIs. 
Needed the pause to finish creates sucessfully.

tests/benchmark/crud_test.go
giteshk: I wanted to create 3 versions and 3 specs. But the benchmark would add up time of all 3 versions which is how it is reported.

To make sure our reports are accurate we only measure the first pass for revisions and specs but let the scripts create all versions.

.github/workflows/composite/benchmark/action.yml
giteshk: This is a template action which has all the steps for benchmark tests.
I found that running the test against sqlite was not reporting accurate results. 
If you look at benchmark.yml you will see the postgres DB.

tests/benchmark/crud_test.go
giteshk: done

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,920 @@
...
+}
+func pauseAfter1000Iterations(i int) {
+	if i%1000 == 0 {
+		/*
+			Pause 5 seconds every 1000 iterations to reduce chances of failure on
+			CREATE, UPDATE, DELETE
+		*/
+		time.Sleep(5 * time.Second)
+	}
+}
```
seaneganx, line 100: Can we replace this with a simpler throttle? Making a file-level variable with a fixed duration to sleep between _every_ iteration seems much easier to understand.

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,920 @@
...
+	gapicClient, err := connection.NewClient(ctx)
+	if err != nil {
+		b.Fatalf("Setup: Failed to create client: %s", err)
+	}
+	client := gapicClient.GrpcClient()
+	b.StopTimer()
+	b.ResetTimer()
+
+	for i := 1; i <= b.N; i++ {
+		pauseAfter1000Iterations(i)
```
seaneganx, line 466: Why are we pausing between GET calls? The comment in the pause function says we only need throttling for mutations.

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,920 @@
...
+				Description: fmt.Sprintf("Description for %s", apiId),
+			},
+		}
+		b.StartTimer()
+		_, err := client.CreateApi(ctx, req)
+		b.StopTimer()
+		/**
+		Go Benchmarking always runs all the benchmark tests once before it runs
+		for the duration specified. So the first row always returns error.
+		*/
```
seaneganx, line 131: Do you have a link to any documentation about this? Why does it cause an error?

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,920 @@
...
+	s, ok := os.LookupEnv("API_NAME_START_OFFSET")
+	if ok {
+		offset, err := strconv.Atoi(s)
+		if err != nil {
+			b.Fatalf("Setup: Invalid API_NAME_START_OFFSET %q: offset must be an integer", s)
+		}
+		i += offset
+	}
+
+	return fmt.Sprintf("%s-%d", prefix, i)
```
seaneganx, line 76: >Say I ran the create script and created 100 APIs. I wanted to add 100 more APIs to an existing environment without having to delete and rerun the script with 200 APIs. This was Needed when I started seeing errors while trying to adding 2000 APIs. so I could break down the create into smaller chunks

I understand it was helpful during development, but why are we including it now? I'm not understanding the benefit to having custom prefixes _and_ offsets. If I want to add 100 more APIs, then I can pick a new prefix and re-run the benchmarks on the same project.

tests/benchmark/crud_test.go
seaneganx: Right, so let's talk about my second question.

>Why do we continue creating other versions after that?

This is supposed to be benchmarking CreateApiVersion alone. Each iteration of the benchmark loop should only create one ApiVersion. Why are we creating three?

If I'm understanding correctly, BenchmarkCreateApiVersion a benchmark _and_ setup code for BenchmarkCreateApiSpec. Is that correct?

Tests should work correctly regardless of the order they execute in. We should be able to run `go test -bench=Get ./tests/benchmark/...` to successfully measure the performance of every Get method.

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,920 @@
...
+		b.Fatalf("Setup: Failed to create client: %s", err)
+	}
+	client := gapicClient.GrpcClient()
+
+	b.StopTimer()
+	b.ResetTimer()
+	buf, err := readAndGZipFile(filepath.Join("testdata", "openapi1.yaml"))
+	if err != nil {
+		b.Errorf("BenchmarkCreateApiSpec could not read openapi1.yaml : %s", err)
+	} else {
```
seaneganx, line 195: This else can be removed. https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

tests/benchmark/testdata/openapi1.yaml
```diff
@@ -0,0 +1,125 @@
+# Copyright 2021 Google LLC
```
seaneganx, line 1: This file can be renamed to `openapi.yaml` now.

seaneganx (changes requested):
Thanks for splitting this into smaller parts Gitesh. Still, there are a lot of new features and changes to how the benchmarks work, so I have a lot of questions and concerns.

Many of my concerns are about the accuracy of the benchmarks. We should strive to keep the benchmarks as simple as possible because the cost of accidentally producing incorrect measurements are high. I think there's a lot we can do to reduce complexity here.

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,920 @@
...
+	if ok {
+		offset, err := strconv.Atoi(s)
+		if err != nil {
+			b.Fatalf("Setup: Invalid API_NAME_START_OFFSET %q: offset must be an integer", s)
+		}
+		i += offset
+	}
+
+	return fmt.Sprintf("%s-%d", prefix, i)
+}
```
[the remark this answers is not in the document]
giteshk: Great recommendation !  made it so much simpler to handle

tests/benchmark/crud_test.go
giteshk: I made it a parameter that can be passed.

tests/benchmark/crud_test.go
giteshk: Removed the throttling from the GET / LIST calls

tests/benchmark/crud_test.go
giteshk: Added this info in the commit message: 

https://github.com/golang/go/issues/32051
The issue seems to be fixed for --benchtime=1x but does not fix --benchtime=100X

tests/benchmark/crud_test.go
giteshk: Say I want to create 10k APIs to test.
I can run the create scripts 5 times with 2k APIs created every time.
Once the 10k APIs are created, we can now run the GET and LIST benchmarks on the 10k APIs and they follow the same pattern for api names (test-1000 test-5000 , test-10000)

tests/benchmark/testdata/openapi1.yaml
giteshk: done

tests/benchmark/crud_test.go
giteshk: You are correct
> If I'm understanding correctly, BenchmarkCreateApiVersion a benchmark and setup code for BenchmarkCreateApiSpec. Is that correct?

I think you are right in terms of we don't need to iterate through every api spec version for the UPDATE/GET calls
That really does not give us any metrics and is wasted execution time. I changed the logic for Update and gets to only operate on the first spec of the first revision of every api.

I left the looping over api versions on the Deletes, to clean everything that was created during the tests.

giteshk (commented):
Thank you for your taking the time to review @seaneganx .
I have made the changes requested and replied to your comments.

seaneganx (changes requested):
The most important thing I've said in the past few rounds of feedback is this: Tests must be self-contained. Here's a list of statements on this topic that should all be true.

* Running an individual benchmark should work without any manual setup.
* Benchmarks should work regardless of the order they are executed in.
* Running a benchmark multiple times repeatedly should not affect its performance.
* Benchmarks should not leave any resources behind when they finish successfully.

Here's how I recommend you structure the benchmarks to achieve this. 

1. Set up the server according to the benchmark parameters.
2. In the benchmark loop, measure the performance of the operation being tested.
3. After the benchmark loop is complete, clean up all the resources created by the test.

timburks:
Hi, I was looking into this today to try to better understand how it got stuck, and I think we're probably trying to do too much in an initial PR. I'll go back and add some comments here-and-there, but I think it might be more productive to restart with something very minimal at first that we later expand to measure more things.

tests/benchmark/benchmark_helpers.go
```diff
@@ -0,0 +1,661 @@
...
+)
+
+var (
+	registryProject string
+	rootResource    string
+)
+
+func init() {
+	flag.StringVar(&registryProject, "registry_project", "bench", "Name of the Registry project")
+	rootResource = fmt.Sprintf("projects/%s/locations/global", registryProject)
```
timburks, line 35: This isn't doing what is probably intended -- because it is evaluated in the `init()` function, the `registryProject `variable still has its default value, so "rootResource" is always "projects/bench/locations/global". 

I found that inside the actual benchmarking function, e.g. `BenchmarkListApis`, `registryProject` does hold the value set by the flag, so `rootResource` could be constructed there either directly or by calling a helper function that refers to the `registryProject` variable.

tests/benchmark/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+    export ITERATION_SIZE=10
+    export REGISTRY_PROJECT_IDENTIFIER=bench
+    export API_VERSIONS_COUNT=3
+    apg admin create-project --project_id=$REGISTRY_PROJECT_IDENTIFIER
+
+    go test --bench=Create ./tests/benchmark \
+        --registry_project=$REGISTRY_PROJECT_IDENTIFIER \
+        --version_count=$API_VERSIONS_COUNT \
+        --benchtime=${ITERATION_SIZE}x --timeout=0
```
timburks, line 15: I was initially puzzled by this and subsequent invocations, wondering what was actually getting called for each value of the `--bench` argument. I found that it's treated as a regex that is used to match benchmark test function names, and since none of the test function names include "Create", "Update", etc., only the `--bench=List` invocation below calls an actual test function. That's probably because this was "design" for an implementation that is coming, but at least for clarity in review, I think it would be good to have the README instructions match the implementation.

tests/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return
+	}
+	operations := [5]string{"Create", "Update", "Get", "List_pagesize_20", "Delete"}
+
+	for _, operation := range operations {
+		b.Run(fmt.Sprintf("%s", operation), func(b *testing.B) {
+			for i := 1; i <= b.N; i++ {
+				apiId := getApiName(i)
+				switch operation {
```
timburks, line 35: This is an interesting structure (almost like a state machine?) which I think might be intended to provide some flexibility that isn't being used yet (since `operations` above is a fixed array). If we wanted a way to repeatedly do a series of things in a standard wrapper, I'd suggest having an array of Go [anonymous functions](https://gobyexample.com/closures) instead, but before doing that I would probably just write one instance in a simpler, possibly more tedious way and abstract from there.

tests/benchmark/benchmark_helpers.go
```diff
@@ -0,0 +1,661 @@
...
+		b.Fatalf("Setup: Failed to create client: %s", err)
+		return err, nil, nil
+	}
+	client := gapicClient.GrpcClient()
+	b.StopTimer()
+	b.ResetTimer()
+	return nil, client, ctx
+}
+
+func CreateApi(apiId string, b *testing.B, record bool, client rpc.RegistryClient, ctx context.Context) *rpc.Api {
```
timburks, line 55: I didn't find any calls of this function where `record` is set to `true`. I tend to think that this is more granularity than necessary -- I would instead just make sure that measurement is off during initialization and teardown, which could be done with calls that start/stop the timer in the main test function. Why might we keep this?

tests/benchmark/crud_test.go
```diff
@@ -0,0 +1,841 @@
...
+func pauseAfter1000Iterations(i int, b *testing.B) {
+	b.Helper()
+	if i%1000 == 0 {
+		/*
+			Pause 5 seconds every 1000 iterations to reduce chances of failure on
+			CREATE, UPDATE, DELETE
+		*/
+		time.Sleep(time.Duration(pauseDuration * time.Second.Nanoseconds()))
+	}
+}
```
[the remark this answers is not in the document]
timburks: These pauses make me uncomfortable because they suggest that we're accepting and working around some odd performance that we don't fully understand (and might need to fix).

timburks:
See #631 which contains some initial benchmarks with smaller scope.

timburks:
@giteshk feel free to reopen this if needed

PR #488, opened 2022-03-10
Update manifest validate function to support advanced patterns

shrutiparabgoogle:
Fixes: https://github.com/apigee/registry/issues/464

### DIFF
--- a/cmd/registry/cmd/upload/artifact.go
+++ b/cmd/registry/cmd/upload/artifact.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"strings"
 
 	"github.com/apigee/registry/cmd/registry/controller"
 	"github.com/apigee/registry/cmd/registry/core"
@@ -109,12 +108,11 @@ func buildManifestArtifact(ctx context.Context, parent string, jsonBytes []byte)
 		return nil, err
 	}
 	errs := controller.ValidateManifest(ctx, parent, m)
-	if len(errs) > 0 {
-		errStrings := make([]string, 0)
+	if count := len(errs); count > 0 {
 		for _, err = range errs {
-			errStrings = append(errStrings, err.Error())
+			log.FromContext(ctx).WithError(err).Error("Manifest error")
 		}
-		return nil, fmt.Errorf("manifest definition contains errors: %s", strings.Join(errStrings, "\n"))
+		return nil, fmt.Errorf("manifest definition contains %d error(s): see logs for details", count)
 	}
 
 	artifactBytes, err := proto.Marshal(m)
--- a/cmd/registry/cmd/upload/artifact.go
+++ b/cmd/registry/cmd/upload/artifact.go
@@ -84,7 +84,7 @@ func buildArtifact(ctx context.Context, parent string, filename string) (*rpc.Ar
 	var artifact *rpc.Artifact
 	switch header.Kind {
 	case "Manifest", patch.ManifestMimeType:
-		artifact, err = buildManifestArtifact(ctx, jsonBytes)
+		artifact, err = buildManifestArtifact(ctx, parent, jsonBytes)
 	case "TaxonomyList", "google.cloud.apigeeregistry.v1.apihub.TaxonomyList":
 		artifact, err = buildTaxonomyListArtifact(ctx, jsonBytes)
 	case "Lifecycle", "google.cloud.apigeeregistry.v1.apihub.Lifecycle":
@@ -101,16 +101,17 @@ func buildArtifact(ctx context.Context, parent string, filename string) (*rpc.Ar
 	return artifact, nil
 }
 
-func buildManifestArtifact(ctx context.Context, jsonBytes []byte) (*rpc.Artifact, error) {
+func buildManifestArtifact(ctx context.Context, parent string, jsonBytes []byte) (*rpc.Artifact, error) {
 	m := &rpc.Manifest{}
 	err := protojson.Unmarshal(jsonBytes, m)
 	if err != nil {
 		return nil, err
 	}
-	err = validateManifest(ctx, m)
-	if err != nil {
-		return nil, err
+	isValid := controller.ValidateManifest(ctx, parent, m)
+	if !isValid {
+		return nil, fmt.Errorf("Manifest definition contains errors")
 	}
+
 	artifactBytes, err := proto.Marshal(m)
 	if err != nil {
 		return nil, err
@@ -119,20 +120,6 @@ func buildManifestArtifact(ctx context.Context, jsonBytes []byte) (*rpc.Artifact
 		Contents: artifactBytes,
 		MimeType: core.MimeTypeForMessageType("google.cloud.apigeeregistry.v1.controller.Manifest"),
 	}, nil
-}
-
-func validateManifest(ctx context.Context, m *rpc.Manifest) error {
-	isValid := true
-	for _, resource := range m.GeneratedResources {
-		if err := controller.ValidateResourceEntry(resource); err != nil {
-			log.FromContext(ctx).WithError(err).Errorf("Invalid manifest entry %v", resource)
-			isValid = false
-		}
-	}
-	if !isValid {
-		return fmt.Errorf("manifest contains errors")
-	}
-	return nil
 }
 
 func buildTaxonomyListArtifact(ctx context.Context, jsonBytes []byte) (*rpc.Artifact, error) {
--- a/cmd/registry/cmd/upload/artifact.go
+++ b/cmd/registry/cmd/upload/artifact.go
@@ -109,7 +109,7 @@ func buildManifestArtifact(ctx context.Context, parent string, jsonBytes []byte)
 	}
 	isValid := controller.ValidateManifest(ctx, parent, m)
 	if !isValid {
-		return nil, fmt.Errorf("Manifest definition contains errors")
+		return nil, fmt.Errorf("manifest definition contains errors")
 	}
 
 	artifactBytes, err := proto.Marshal(m)
--- a/cmd/registry/cmd/upload/artifact.go
+++ b/cmd/registry/cmd/upload/artifact.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"strings"
 
 	"github.com/apigee/registry/cmd/registry/controller"
 	"github.com/apigee/registry/cmd/registry/core"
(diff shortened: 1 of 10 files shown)


cmd/registry/controller/parser.go
```diff
@@ -157,39 +157,75 @@ func resourceNameFromParent(
 	return resource, nil
 }
 
-func ValidateResourceEntry(resource *rpc.GeneratedResource) error {
-	var group string
+func validateGeneratedResourceEntry(parent string, generatedResource *rpc.GeneratedResource) []error {
+	totalErrors := make([]error, 0)
```
seaneganx, line 161: Nit: I usually see `errs` used for this pattern.

cmd/registry/controller/parser.go
```diff
@@ -157,39 +157,75 @@ func resourceNameFromParent(
...
 	if err != nil {
-		return err
+		totalErrors = append(totalErrors, err)
 	}
-	// no $resource reference present
-	if len(entity) == 0 {
-		return nil
+	for _, r := range references {
+		if !validateEntityReference(parsedTargetResource, r.entityType) {
+			totalErrors = append(totalErrors, fmt.Errorf("Invalid reference in action: %s", generatedResource.Action))
```
seaneganx, line 220: Nit: errors shouldn't be capitalized.

cmd/registry/controller/controller.go
```diff
@@ -31,6 +31,20 @@ type Action struct {
 	RequiresReceipt   bool
 }
 
+func ValidateManifest(ctx context.Context, parent string, manifest *rpc.Manifest) bool {
+	isValid := true
+	for _, resource := range manifest.GeneratedResources {
+		log.FromContext(ctx).Debugf("Validating entry: %v", resource)
+		if errs := validateGeneratedResourceEntry(parent, resource); errs != nil {
```
seaneganx, line 38: We should always consider `nil` slices the same as an empty slice to avoid confusion and subtle bugs, so we shouldn't be checking `errs != nil` here. Length checks are preferred.

shrutiparabgoogle:
Bumping this up for review.

cmd/registry/controller/resource-name.go
```diff
@@ -22,113 +22,113 @@ import (
...
-	GetApi() string
-	String() string
-	GetParent() string
+type resourceName interface {
+	getArtifact() string
+	getSpec() string
+	getVersion() string
+	getApi() string
+	string() string
+	getParent() string
```
seaneganx, line 31: I don't think we need to unexport these methods if we're already unexporting the entire struct. Particularly, we probably still want this type to implement [fmt.Stringer](https://go.dev/tour/methods/17).

Also a nitpick since these are being updated: [getters in Go don't usually include the word "get" in the method name](https://go.dev/doc/effective_go#Getters), so these can be renamed to `Artifact() string`, `Spec() string`, etc.

The same advice applies to the rest of this file and `resource.go`.

cmd/registry/controller/controller.go
```diff
@@ -31,6 +31,20 @@ type Action struct {
...
+			for _, err := range errs {
+				log.FromContext(ctx).WithError(err).Errorf("Invalid manifest entry")
```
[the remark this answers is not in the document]
seaneganx: >The original thinking behind this was, because this validate function is used outside of the package (in upload manifest and upload artifact commands) where the general pattern is usually returning a single error and a response. That error is then just logged as is without any processing.

IMO we should give those commands control over their log output. If we return multiple errors they can choose to loop and log each one, replace the detailed errors with a summary, ignore the errors, etc. It gives them flexibility.

>Another reason to log errors here is so that we can specifically mention which errors belong to which manifest entry, with the log entry Validating entry %v. If we pass a list of errors to the calling function, it will have to traverse through the whole list and log a combined list of errors without identifying which entry it belongs to (which we can solve by generating more informative errors)

Anything we can log in this function can be returned as an error with the same information. We can give access to any level of detail that callers might want, e.g. validation for the entire manifest, or specific parts.

>but the purpose of this function is to just return a yes/no answer to the callee.

Why do callers need that? It looks like returning false will just result in an error being propagated back to the root command, leading to an error with very little detail about the true cause. This seems like another reason to return errors instead.

cmd/registry/controller/controller.go
```diff
@@ -41,9 +55,9 @@ func ProcessManifest(
 	for _, resource := range manifest.GeneratedResources {
 		log.Debugf(ctx, "Processing entry: %v", resource)
 
-		err := ValidateResourceEntry(resource)
-		if err != nil {
-			log.FromContext(ctx).WithError(err).Debugf("Skipping resource: %q invalid resource pattern", resource)
+		errs := validateGeneratedResourceEntry(fmt.Sprintf("projects/%s/locations/global", projectID), resource)
```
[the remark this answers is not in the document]
seaneganx: Skipping invalid entries sounds fine to me. I still think it can be valuable to pre-validate the entire manifest before processing though.

For example, we can log any issues with the manifest entries and remove them from the list of entries before processing it. That ensures the errors aren't mixed in with all the other non-error logs and can simplify the processing code.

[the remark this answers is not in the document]
shrutiparabgoogle: Makes sense, updated the code.

cmd/registry/controller/controller.go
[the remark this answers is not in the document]
shrutiparabgoogle: Done. Now `manifest upload` and `artifact upload` commands are logging the errors

cmd/registry/controller/parser.go
```diff
@@ -157,39 +159,84 @@ func resourceNameFromParent(
...
+
+func validateGeneratedResourceEntry(parent string, generatedResource *rpc.GeneratedResource) []error {
+	errs := make([]error, 0)
+	parsedTargetResource, err := parseResourcePattern(
+		fmt.Sprintf("%s/%s", parent, generatedResource.Pattern))
+
+	// Check that the target resource pattern should be a valid pattern.
+	// Return for errors in target resource pattern since we caan't verify action and dependencies based off an incorrect pattern.
+	if err != nil {
+		errs = append(errs, fmt.Errorf("Invalid pattern for generatedResource %v, %s", generatedResource.Pattern, err))
```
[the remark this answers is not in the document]

cmd/registry/controller/parser.go
```diff
@@ -157,39 +159,84 @@ func resourceNameFromParent(
...
+		return errs
+	}
+	// Check that generatedResource pattern doesn't end with a "-".
+	// We require a name for the target resource.
+	if strings.HasSuffix(parsedTargetResource.string(), "/-") {
+		errs = append(errs, fmt.Errorf("Invalid generatedResource pattern: %q, it should end with a name and not a \"-\"", generatedResource.Pattern))
```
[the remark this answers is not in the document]

cmd/registry/controller/resource-name.go
shrutiparabgoogle: Done.

cmd/registry/controller/parser.go
```diff
@@ -130,7 +131,7 @@ func extendDependencyPattern(
 
 func resourceNameFromParent(
 	resourcePattern string,
-	parent string) (ResourceName, error) {
+	parent string) (resourceName, error) {
```
seaneganx, line 134: Nit: This seems like it can be reduced to one line now, since it's shorter than `ValidateManifest` below.

`func resourceNameFromParent(resourcePattern, parent string) (resourceName, error)`

cmd/registry/controller/parser.go
```diff
@@ -157,39 +158,78 @@ func resourceNameFromParent(
 	return resource, nil
 }
 
-func ValidateResourceEntry(resource *rpc.GeneratedResource) error {
-	var group string
+func ValidateManifest(ctx context.Context, parent string, manifest *rpc.Manifest) []error {
+	totalErrors := make([]error, 0)
+	for _, resource := range manifest.GeneratedResources {
+		if errs := validateGeneratedResourceEntry(parent, resource); len(errs) != 0 {
+			for _, err := range errs {
```
seaneganx, line 165: Nit: This length check isn't necessary because `range` handles `nil` or empty lists gracefully.

cmd/registry/controller/parser.go
```diff
@@ -157,39 +158,78 @@ func resourceNameFromParent(
...
+	return totalErrors
+}
+
+func validateGeneratedResourceEntry(parent string, generatedResource *rpc.GeneratedResource) []error {
+	errs := make([]error, 0)
+	parsedTargetResource, err := parseResourcePattern(
+		fmt.Sprintf("%s/%s", parent, generatedResource.Pattern))
+
+	// Check that the target resource pattern should be a valid pattern.
+	// Return for errors in target resource pattern since we caan't verify action and dependencies based off an incorrect pattern.
```
seaneganx, line 179: Nit: Replace `caan't` with `can't`.

cmd/registry/controller/parser.go
```diff
@@ -157,39 +158,78 @@ func resourceNameFromParent(
...
+	if err != nil {
+		errs = append(errs, fmt.Errorf("invalid pattern for generatedResource %v, %s", generatedResource.Pattern, err))
+		return errs
```
seaneganx, line 182: Nit: Consider declaring the single element slice as it's returned to make it clearer that this is the only error that will be returned. `errs` can be declared before the loop where multiple errors might actually be returned.

```
return []error{
    fmt.Errorf("invalid pattern for generatedResource %v, %s", generatedResource.Pattern, err),
}
```

Same goes for the suffix check below.

cmd/registry/cmd/upload/artifact.go
```diff
@@ -101,16 +102,21 @@ func buildArtifact(ctx context.Context, parent string, filename string) (*rpc.Ar
...
-	err = validateManifest(ctx, m)
-	if err != nil {
-		return nil, err
+	errs := controller.ValidateManifest(ctx, parent, m)
+	if len(errs) > 0 {
+		errStrings := make([]string, 0)
+		for _, err = range errs {
+			errStrings = append(errStrings, err.Error())
+		}
+		return nil, fmt.Errorf("manifest definition contains errors: %s", strings.Join(errStrings, "\n"))
```
seaneganx, line 117: How does this error appear to users? I don't often see multi-line error messages or log entries.

Consider logging the errors here and return an error summarizing the issue for the main command to include in the fatal log. This might present better to users even though it pushes logging deeper into the command implementation.

```
if count := len(errs); count > 0 {
	for _, err := range errs {
		log.FromContext(ctx).WithError(err).Error("Manifest error")
	}
	return fmt.Errorf("manifest definition contains %d error(s): see logs for details", count)
}
```

seaneganx (approved):
Some final comments for optional changes. LGTM overall.

cmd/registry/controller/parser.go

cmd/registry/controller/parser.go

cmd/registry/controller/parser.go

cmd/registry/controller/parser.go

cmd/registry/cmd/upload/artifact.go
shrutiparabgoogle: Thanks, I think this is a better approach.

PR #490, opened 2022-03-11
Build registry controller ghcr.io image

giteshk:
Sync docker files from the hosted version

### DIFF
--- a/containers/registry-server/Dockerfile
+++ b/containers/registry-server/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


shrutiparabgoogle:
Looks good. I have only one question.

I agree with keeping `registry-controller` in sync with the hosted version. 
The initial purpose of `registry-tools` and `registry-linter` images was to demonstrate the pattern of layering. The idea was that anyone can take our `registry-tools` as the base image and add necessary tools on top to have their own custom controller. This was shown through `registry-linter`. I am open to other ways of showing this possibility, maybe through documentation? What are your thoughts on demonstrating that possibility?

giteshk:
@shrutiparabgoogle, We will leave the registry-linter in place. This will serve as a good example of extending the controller capabilities.

PR #491, opened 2022-03-14
Add support for uploading local spec files using "registry apply"

timburks:
This refines the "registry apply" command to support uploading of API specs from local files. This reads both individual files (useful for OpenAPI and Discovery specs) and directories of files (useful for proto-based specs, which are typically multifile). Directories are uploaded as zip archives.


cmd/registry/patch/spec.go
```diff
@@ -84,26 +86,55 @@ func applyApiSpecPatch(
...
 		if err != nil {
 			return err
 		}
-		body, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			return err
-		}
-		if strings.Contains(spec.Data.MimeType, "+gzip") {
-			body, err = core.GZippedBytes(body)
+		if u.Scheme == "http" || u.Scheme == "https" {
```
seaneganx, line 95: Consider using a switch statement for this.

cmd/registry/patch/spec.go
```diff
@@ -84,26 +86,55 @@ func applyApiSpecPatch(
...
+			if strings.Contains(spec.Data.MimeType, "+gzip") {
+				body, err = core.GZippedBytes(body)
+				if err != nil {
+					return err
+				}
+			}
+			req.ApiSpec.Contents = body
+		} else if u.Scheme == "file" {
+			// remove leading slash from path; we load from paths relative to the working directory
+			path := strings.TrimLeft(u.Path, "/")
```
seaneganx, line 113: Is this supposed to trim multiple leading slashes? If not, let's use TrimPrefix.

seaneganx (approved):

cmd/registry/patch/spec.go
timburks: Great suggestion, thanks - that will let people "force" global paths if they want. Done.

cmd/registry/patch/spec.go
timburks: SGTM, changed.

Issue #445, opened 2022-01-20
Task improvements for the registry tool


seaneganx:
### Background

The registry tool makes heavy use of the [worker pool](https://github.com/apigee/registry/blob/6201f82be8768ba690b3e32446ec43bf7c8a39c0/cmd/registry/core/tasks.go#L36) for concurrent execution of subtasks associated with various commands. Currently if a task returns an error, the worker pool will terminate the command entirely. Some tasks [intentionally avoid returning errors](https://github.com/apigee/registry/blob/6201f82be8768ba690b3e32446ec43bf7c8a39c0/cmd/registry/cmd/annotate/annotate.go#L171) to prevent this.

[...]

PR #494, opened 2022-03-16
Remove unused and simple single-use code from libraries

seaneganx:
I've been noticing unused code throughout our libraries and decided to remove it. Most is old and seemingly forgotten code, while some might actually end up being added again in the future if we invest in those features. All of it is either trivial to add back or has room benefit from being rewritten if it becomes useful in the future.

Whenever the code being removed was exported I checked against the registry-experimental repo to make sure it wasn't in use there.


cmd/registry/patch/api.go
```diff
@@ -33,7 +33,6 @@ type ApiData struct {
 	RecommendedDeployment string           `yaml:"recommendedDeployment,omitempty"`
 	ApiVersions           []*ApiVersion    `yaml:"versions,omitempty"`
 	ApiDeployments        []*ApiDeployment `yaml:"deployments,omitempty"`
-	Artifacts             []*Artifact      `yaml:"artifacts,omitempty"`
```
timburks, line 36: This one is unfortunate - I have WIP that will put it right back.

cmd/registry/cmd/upload/csv.go
```diff
@@ -57,7 +58,14 @@ func csvCommand(ctx context.Context) *cobra.Command {
 				log.FromContext(ctx).WithError(err).Fatal("Failed to get client")
 			}
 
-			core.EnsureProjectExists(ctx, adminClient, projectID)
+			if _, err := adminClient.UpdateProject(ctx, &rpc.UpdateProjectRequest{
```
timburks, line 61: Do you think this is clearer? On the left I know exactly what is happening from the function name and on the right I have to parse 8 lines of text.

timburks (commented):
Thanks for going through all this! It feel like "omnibus" legislation, which makes it hard to review as a single unit. I'm not confident to merge it or to say that we wouldn't want to merge it, I just want to spend more time reviewing and testing first. Individual PRs would be easier to evaluate, but I also don't want to add that work to you -- just explaining my slow review.

cmd/registry/cmd/upload/csv.go
seaneganx: I think they're both pretty clear, but I agree this is less concise so it loses some points there. On the other hand, there is value in being able to see what is happening in detail without having to look at a separate function. I consider this change a toss-up for readability overall.

We can discuss whether it would be better in a _local_ helper function (again though, I think it's a toss-up), but there are more important questions to address first.

- Should this code be here at all? None of our other commands create a project because we're encouraging flat organization of APIs into a single project per user.
- Given how many occurrences there are of similar code throughout the registry tool, shouldn't we have a more robust library instead of several mostly overlapping helper functions scattered throughout the codebase? I mention this in #445.

cmd/registry/patch/api.go
seaneganx: Just reverted the commit to avoid churn. I incorrectly assumed this would end up in our pile of TODOs for a while still.

seaneganx:
>Thanks for going through all this! It feel like "omnibus" legislation, which makes it hard to review as a single unit. I'm not confident to merge it or to say that we wouldn't want to merge it, I just want to spend more time reviewing and testing first.

Are you checking the individual commits? There's very substance in this PR given that almost everything being touched is dead code. I intentionally only made small adjustments to live code, e.g. in-lining helpers that are only used once, in-lining local variables where I thought it improved readability, etc.

What specifically do you think needs testing for this PR?

cmd/registry/cmd/upload/csv.go
timburks: OK, I see what you were doing as chipping away at core (which makes sense). I can see value in making this a local helper function for clarity but let's not make that change now. I'm also happy to see a "robust library" emerge, but would wait for a clear need for that. The "mostly overlapping helper functions" don't strike me as a large problem yet (when they become so, we can create the library)

cmd/registry/core/labeling.go
```diff
@@ -47,27 +47,3 @@ func (l *Labeling) Apply(m map[string]string) (map[string]string, error) {
 	}
 	return m, nil
 }
-
-// UpdateMap updates a map containing labels or annotations to be modified.
-func UpdateMap(m map[string]string,
```
timburks, line 52: LGTM - just noting, this seems to have been replaced by the function above it.

cmd/registry/core/types.go
```diff
@@ -35,14 +35,6 @@ func ProtobufMimeType(compression string) string {
 	return fmt.Sprintf("application/x.protobuf%s", compression)
 }
 
-// TODO: tighten these up, possibly using regular expressions.
-
-// IsAsyncAPIv2 returns true if a MIME type represents an AsyncAPI v2 spec.
-func IsAsyncAPIv2(mimeType string) bool {
```
timburks, line 41: hopefully returning soon, but no big deal to recreate

cmd/registry/cmd/apply/apply_test.go
```diff
@@ -78,7 +78,7 @@ func TestApply(t *testing.T) {
 		}
 		api := project.Api("registry")
 		// TODO: Prefer using the API client directly instead of relying on application code that isn't directly related to this test.
-		_, err = core.GetAPI(ctx, registryClient, api, func(message *rpc.Api) {
+		err = core.GetAPI(ctx, registryClient, api, func(message *rpc.Api) {
```
timburks, line 81: Not a blocker, and I've done this too - but this change (overall in this file) seems big for this PR.

cmd/registry/patch/api.go
timburks: No worries, I should have left that alone - it's easy enough to add back.

timburks (approved):
Thanks again for breaking this out into small commits - that made this much easier to review. I might have preferred for one or two the bigger-impact changes to be separate PRs but that's for future consideration only. Onward! :)

PR #495, opened 2022-03-16
Fix filtering overwrite bug in listing helpers

### LINKED ISSUE #420: Registry tool: review filter usage in core/list.go
Reviews of #411 pointed out some possibly-redundant variables, and when I looked at the code, I think we may be losing filter arguments in certain cases in situations like [this](https://github.com/apigee/registry/blob/29025b9b11ad5d7dd5fe06c899feba9d6a7f1cfa/cmd/registry/core/list.go#L36), which occurs throughout [core/list.go](https://github.com/apigee/registry/blob/main/cmd/registry/core/list.go).

seaneganx:
Fixes #420.

### DIFF
--- a/cmd/registry/core/list.go
+++ b/cmd/registry/core/list.go
@@ -26,26 +26,21 @@ import (
 func ListProjects(ctx context.Context,
 	client *gapic.AdminClient,
 	name names.Project,
-	filterFlag string,
+	filter string,
 	handler ProjectHandler) error {
-	request := &rpc.ListProjectsRequest{}
-	filter := filterFlag
-	projectID := name.ProjectID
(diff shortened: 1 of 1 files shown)


timburks (approved):
Thanks, so this fix is to move the item-specific filter to the client side? I think that's fine until (if ever) we find that it's a problem (due to too much API traffic).

seaneganx:
>Thanks, so this fix is to move the item-specific filter to the client side?

That was the plan because I didn't realize how simply we can add an extra condition to user provided filters. If we find any cases where appending ` && something_id == 'some_id'` to the filters doesn't work as intended then we can try a more robust approach.

PR #496, opened 2022-03-16
Support errors in listing handlers

### LINKED ISSUE #487: Registry tool: Improve error handling for core/list.go
It's difficult to handle errors while listing resources using the helpers in core/list.go because the callback functions don't return errors. When error handling is necessary, we use [complicated workarounds](https://github.com/apigee/registry/blob/3612f11a4962082b2989f2925bc19052aac5e065/cmd/registry/patch/api.go#L73).

One possible solution is to adopt a pattern similar to [fs.WalkDir](https://pkg.go.dev/io/fs#WalkDir), where the callback function can return an error to immediately halt the operation and return the error to the caller.

seaneganx:
I didn't want to make any decisions about which errors should be stopping the loops, so everything is kept as it was behaving before this change. At some point we should switch some of the `return nil` statements to `return err`.

Fixes #487.


timburks:
LGTM - I see that handler errors terminate listings. I agree with that, assuming that handlers with warnings would just log them and not return errors.

timburks (approved):

Issue #485, opened 2022-03-08
Registry controller: Refactor generateCreateActions()


shrutiparabgoogle:
As pointed out in this comment: https://github.com/apigee/registry/pull/471#discussion_r821969034
There is scope for removing redundancy in `generateCreateAction()`. 
Currently, the type of `parent` is different in the two cases a `string` vs a `resourceInstance()`. 
It will involve refactoring of the underlying functions in order to converge the `types` in both cases and reuse the same code.
Hence, this issue is tracking separately the refactoring efforts required.

Two tasks as part of this refactor:
* Converge the type of `parent` in `generateCreateAction()`.
* Converge `generateCreateAction()` and `generateUpdateAction()` into using same logic, fetch parents for only non-existent target resources.

PR #501, opened 2022-03-23
Refactor controller code for CreateActions

shrutiparabgoogle:
Refactor controller code:
* Reuse the `takeAction` loop in `generateCreateActions()` as suggested [here](https://github.com/apigee/registry/issues/485)
* Changed `resourceName.Parent() string` to `resourceName.ParentName() resourceName`.
* Use a list of non-visited parents while generating non-existent resources.

Fixes: https://github.com/apigee/registry/issues/485


cmd/registry/controller/controller.go
```diff
@@ -232,73 +233,51 @@ func generateCreateActions(
...
 		if err != nil {
-			return nil, err
-		} else if !takeAction {
-			return nil, nil
+			return nil, fmt.Errorf("Internal error: Invalid visited map, %s", err)
 		}
-
-		cmd, err := generateCommand(generatedResource.Action, targetResourceName.String())
+		// Construct a filter to fetch only non-visited parents.
+		filter := fmt.Sprintf("%s.all(n, !(name.contains(n)))", visited_json)
```
seaneganx, line 271: What's happening here? I think this is worth explaining in more detail in the code comment.

cmd/registry/controller/controller.go
```diff
@@ -232,73 +233,51 @@ func generateCreateActions(
...
-	if strings.HasSuffix(parentName, "locations/global") {
+
+	parentName := parsedResourcePattern.ParentName()
+	switch parentName.(type) {
+	case projectName:
+		// If parent is a project, we can't list projects since this is registry client command.
+		// Since the manifest definition is scoped  only for a particular project,
+		// there will be only one target resource in this case.
+		// There are two cases where this might happen:
+		// 1. Target resource is a project level artifact "projects/demo/locations/global/artifacts/serach-index"
```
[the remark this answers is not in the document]

cmd/registry/controller/controller.go
```diff
@@ -232,73 +233,51 @@ func generateCreateActions(
...
 
-		takeAction, err := needsCreate(
-			targetResourceName,
-			dependencyMaps,
-			generatedResource,
-		)
-
+	default:
+		// If parent resource is not a project, then go through all the non-visited parents.
+		visited_json, err := json.Marshal(visited)
```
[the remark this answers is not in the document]

cmd/registry/controller/controller.go
```diff
@@ -232,73 +233,51 @@ func generateCreateActions(
...
 		if err != nil {
-			return nil, err
-		} else if !takeAction {
-			return nil, nil
+			return nil, fmt.Errorf("Internal error: Invalid visited map, %s", err)
```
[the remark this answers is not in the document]

cmd/registry/controller/controller.go
shrutiparabgoogle: Added more explanation, let me know if it is clear enough to understand.

cmd/registry/controller/controller.go
```diff
@@ -232,73 +233,64 @@ func generateCreateActions(
...
+		//	"projects/demo/locations/global/apis/example-api1": true,
+		//  "projects/demo/locations/global/apis/example-api2": true,
+		// }
+		// We make use of `e.all(x,p)` macro as defined here: https://github.com/google/cel-spec/blob/master/doc/langdef.md#macros
+		// The CEL filter for listing non-visited parents will be as follows:
+		// '{
+		//	"projects/demo/locations/global/apis/example-api1": true,
+		//  "projects/demo/locations/global/apis/example-api2": true,
+		// }.all(n, !(name.contains(n)))'
+		filter := fmt.Sprintf("%s.all(n, !(name.contains(n)))", visitedJSON)
```
seaneganx, line 284: I think the explanation is clear. Do we use `contains` instead of `==` because we want to ignore all resources whose parent has been visited? Should we be using `startsWith` instead?

It doesn't seem like we need to use `json.Marshal` here either. That was one of the things that confused me when I first looked at the code because I wasn't sure how a serialized map would work with the `all` macro. Building the JSON array using string processing can avoid the possibility of an error, as a bonus.

This can be captured in a separate function to help make the code a bit more self-documenting and reduce the noise in this func.

```
// Constructs a CEL filter to exclude resources with visited parents.
// The filter excludes resources whose `name` property starts with any of the visited parent names.
// 
// For example, consider a visited map of parents which are apis:
// {
//     "projects/demo/locations/global/apis/example-api1": true,
//     "projects/demo/locations/global/apis/example-api2": true,
// }
//
// The resulting CEL filter will be:
// ["projects/demo/locations/global/apis/example-api1","projects/demo/locations/global/apis/example-api2"].all(prefix, !name.startsWith(prefix))
//
// Note: The `bool` values in the input map are ignored. The filter will use every map key. 
func excludeVisitedParents(v map[string]bool) string {
	// Wrap each string with quotes and join them with commas to build a JSON string array.
	jsonStrings := make([]string, 0, len(v))
	for parent := range v {
		jsonStrings = append(jsonStrings, fmt.Sprintf("%q", parent))
	}
	return fmt.Sprintf("[%s].all(parent, !name.startsWith(parent))", strings.Join(jsonStrings, ","))
}
```

seaneganx (approved):

cmd/registry/controller/controller.go
shrutiparabgoogle: Thanks for this suggestion.

Created a separate function as suggested. 
For your question on the use of `contains`, yes, the original intention was to ignore all the resources with parents as prefixes. However, after doing another pass at it, I realized that there is no case right now where, the `List` fetches children resources of the parents which are stored in the visited map.  The expectation is that the values stored in the visited map should be full resource names and not prefixes to any of the parents. Use of `contains` opens the possibility for bugs in the logic which might go unnoticed.

I am going to revert this to ==` operation for now, we can always change it to `startsWith` if we feel the need for it in the future.

PR #503, opened 2022-03-26
Add support for more apihub resources, add YAML import of API-owned artifacts.

timburks:
This adds support for display settings and references, including YAML apply and export. APIs can now be applied/exported with certain child artifacts (notably references), and this can be extended to include additional artifact types as needed.

Artifacts can be imported as subcollections of APIs in YAML files. This allows certain annotations like "references" to be easily specified in external source input ([example](https://github.com/apigee/registry/pull/503/commits/cc5b1db5473c40b9126b3a51258ace076dda1ab6#diff-1378506d005890cc95520c4803ac9e77b157e6d6b157b2097d4f4d89e3cdacde)). This artifact import mechanism is generic - any type of artifact can be attached at any level. This doesn't always make sense, but typing and enforcement seems better left for later. This means that artifact lists in YAML are polymorphic, which makes marshalling and unmarshalling a bit more tricky (see patch/artifact.go for the implementation) - but this generality allows new artifact types to be easily supported.

In a previous PR, I added some code that passed cmp.Diff() arguments in the wrong order. I believe that's fixed now.


timburks:
Let's merge #510 before this to reduce the complexity of this PR.

cmd/registry/patch/apply.go
```diff
@@ -20,25 +20,31 @@ import (
...
-			return applyFile(ctx, client, path, parent)
+			return applyFile(ctx, client, p, parent)
 		})
 }
 
 func applyFile(ctx context.Context, client connection.Client, fileName, parent string) error {
+	if !strings.HasSuffix(fileName, ".yaml") {
+		return nil
+	}
+	log.FromContext(ctx).Infof("Importing %s", fileName)
```
seaneganx, line 47: This seems to me like it should be part of the directory walking code. It's just another property of the file/path that we ignore during the walk.

cmd/registry/patch/artifact.go
```diff
@@ -43,46 +48,99 @@ func ExportArtifact(ctx context.Context, client *gapic.RegistryClient, message *
...
+func (a *Artifact) MarshalYAML() (interface{}, error) {
+	type exportable struct {
+		Header `yaml:",inline"`
+		Data   ArtifactData `yaml:"data"`
+	}
+	return &exportable{
+		Header: a.Header,
+		Data:   a.Data,
+	}, nil
+}
```
seaneganx, line 120: What is this?

cmd/registry/patch/artifact_references.go
```diff
@@ -0,0 +1,102 @@
...
+		reference := Reference{
+			ID:          t.Id,
+			DisplayName: t.DisplayName,
+			Category:    t.Category,
+			Resource:    t.Resource,
+			URI:         t.Uri,
+		}
+		referenceListData.References = append(referenceListData.References, reference)
+	}
+	return referenceList, nil
```
seaneganx, line 101: Consider reorganizing this so the return value can be constructed as one literal.

```
refs := make([]Reference, len(value.References))
for i, ref := range value.References {
	refs[i] = Reference{
		ID:          ref.Id,
		DisplayName: ref.DisplayName,
		Category:    ref.Category,
		Resource:    ref.Resource,
		URI:         ref.Uri,
	}
}

return &Artifact{
	Header: Header{
		ApiVersion: RegistryV1,
		Kind:       "ReferenceList",
		Metadata: Metadata{
			Name: artifactName.ArtifactID(),
		},
	},
	Data: &ReferenceListData{
		DisplayName: value.DisplayName,
		Description: value.Description,
		References: refs,
	},
}, nil
```

Same advice for similar functions.

seaneganx (commented):
Aside from the one piece of mystery code (see "What is this?") this LGTM.

cmd/registry/cmd/upload/artifact.go
```diff
@@ -101,6 +105,38 @@ func buildArtifact(ctx context.Context, parent string, filename string) (*rpc.Ar
 	return artifact, nil
 }
 
+func buildDisplaySettingsArtifact(ctx context.Context, jsonBytes []byte) (*rpc.Artifact, error) {
+	m := &rpc.DisplaySettings{}
+	err := protojson.Unmarshal(jsonBytes, m)
+	if err != nil {
```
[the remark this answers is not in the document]
timburks: OK, I've updated that one. I'll watch for this, but note that it's inconsistently applicable - for example, the code right below it can't be collapsed:
```
	artifactBytes, err := proto.Marshal(m)
	if err != nil {
		return nil, err
	}
```
because that takes artifactBytes out of scope for the subsequent expressions.

cmd/registry/core/zip.go
```diff
@@ -101,23 +102,24 @@ func UnzipArchiveToMap(b []byte) (map[string][]byte, error) {
 // ZipArchiveOfPath reads the contents of a path into a zip archive.
 // The specified prefix is stripped from file names in the archive.
 // Based on an example published at https://golangcode.com/create-zip-files-in-go/
-func ZipArchiveOfPath(path, prefix string) (buf bytes.Buffer, err error) {
+func ZipArchiveOfPath(path, prefix string, recursive bool) (buf bytes.Buffer, err error) {
 	zipWriter := zip.NewWriter(&buf)
 	defer zipWriter.Close()
 
-	_ = filepath.Walk(path, func(p string, info os.FileInfo, err error) error {
+	err = filepath.WalkDir(path, func(p string, entry fs.DirEntry, err error) error {
```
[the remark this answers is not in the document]
timburks: True. I don't love this optimization because the resulting code is
```
return buf, filepath.WalkDir...
```
and this expects readers to know that buf is returned-by-reference and thus the returned value is modified by the expression that follows it. I think people *should* know that but it feels like a Go interview question. Happy to change it if you feel strongly, though.

cmd/registry/core/zip.go
```diff
@@ -101,23 +102,24 @@ func UnzipArchiveToMap(b []byte) (map[string][]byte, error) {
...
 		if err != nil {
 			return err
-		}
-		if info.IsDir() {
-			return nil
+		} else if entry.IsDir() && (recursive || (p == path)) {
+			return nil // Do nothing for the directory, but still walk its contents.
+		} else if entry.IsDir() {
+			return filepath.SkipDir // Skip the directory and contents.
```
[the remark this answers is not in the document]
timburks: Updated - I also swapped the order of the !recursive and p != path checks to make it more intuitive (for me)...
"if this is a directory that is not the top-level directory and we're not recursively expanding" then skip its contents

cmd/registry/patch/apply.go
timburks: Thanks, looking up at the code above it, it seems that putting this check there would excessively complicate that code (and cause it to diverge from the WalkDir in the zip builder), so I'll leave this for now but am not opposed to revisiting this.

cmd/registry/patch/artifact.go
timburks: I found that this was necessary to properly export artifact YAMLs. You'll see that in the struct tags above (on line 31), the tag for "Data" is a dash ("-"). This prevents the importer from directly importing this value -- which it can't do because this value could be a struct of multiple types. See the StackOverflow link on line 84 for more on that.

Now, with "-" in the struct tag for Data, YAML exports will skip that field. Unlike imports, YAML exports have no problems with polymorphic fields, so we can get the field to export by setting its struct tag back to "data". Rather than mess with reflection or whatever is needed to tweak the struct tag, it seemed better to just copy the artifact into a new structure with appropriate tags and then export it.

I'm happy to do this in a better way if there is one, but I think having polymorphic Data inside artifacts will be a requirement (it's a Kubernetes resource pattern that I think someone was trying to follow when they asked that question on SO). I think in a future revision, we might want to have this code "float" up into a general reader of wrapped types that reads all of the YAMLs (APIs, Artifacts, etc) in one place in the code.

cmd/registry/cmd/upload/artifact.go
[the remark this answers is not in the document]
seaneganx: Right, it's not always possible so `err` sometimes needs to be function scoped. When we can though, it's good to reduce scope.

cmd/registry/patch/artifact.go
seaneganx: I think this is worth explaining in a code comment. It's not obvious that we're implementing an interface that the `yaml` package is using at marshal time.

cmd/registry/patch/artifact_references.go
timburks: Done.

cmd/registry/patch/artifact.go
timburks: OK, done!

seaneganx (approved):

Issue #492, opened 2022-03-15
Registry Server: Improve the performance of ListApis


giteshk:
The response time of the ListApis gets progressively worse with the increate in the number of APIs.

This is because we fetch all the rows and then apply the filters.

Improving this API call will have positive effects on the usability of this project.

some initial benchmark results for a setup with Registry server running on GKE with Cloud Postgresql

| Number of APIs | Number of Versions|Response time in ms/op|
|-|-|-|
|100|1|160|
|500|1|300|
|1000|1|436|
|2500|1|864|
|100|3|69|
|500|3|270|
|1000|3|418|
|2500|3|885|

seaneganx:
Is there much variation in performance between runs? I'm surprised to see a 70% difference in response time for Listing APIs when they have a different number of versions. It's surprising because we don't read the versions table at any point during a ListAPIs call.

With larger numbers of APIs the performance seems to stabilize, reducing from 70% (100 APIs) to 11% (500 APIs) to 4% (1000 APIs) to 2% (2500 APIs). This makes me suspicious of the accuracy for those measurements with lower API counts. What happens if we run the benchmarks multiple times and average the results?

seaneganx:
>For the Listing API call with no Filters the page size was set to (Number of APIs)/10

I'm concerned about how meaningful our results are when we have so many variables in the benchmarks. Increasing the page size likely causes a decrease in performance per call, and so does increasing the number of specs. If we see a decrease, how do we know which variable is causing it? IMO this page size should be constant - either a configurable constant if we want to measure the effect of increasing page sizes, or a fixed constant if we don't.

giteshk:
Agreed.. I  changed the page size to **20** last week after for the benchmark tests to ensure the only variable is the number of API Specs.  Thanks for pointing this out.

PR #504, opened 2022-03-30
Optimize memory usage for database listing methods

### LINKED ISSUE #148: Registry API: Optimize memory usage for database listing operations
Listing operations work by reading the entire table into an iterator then processing resources one at a time until a page is filled or the entire table has been read.

The iterator struct could be modified to fetch smaller batches of resources and automatically attempt to fetch another batch after all of its resources are accessed. The iterator behavior would not change, but the number of resources that are unnecessarily read from the database and stored in memory would be reduced.

seaneganx:
Fixes #148 and affects #492.


server/registry/internal/storage/list.go
```diff
@@ -25,10 +26,16 @@ import (
...
+// limit returns the database page size to use for a listing request.
+func limit(opts PageOptions) int {
+	// Without filters, read exactly enough rows to fill the page,
+	// plus an extra row to check if another page exists.
+	if opts.Filter == "" {
+		return int(opts.Size) + 1
+	}
+
+	// When filters are present, we should read larger pages.
+	return 500
```
seaneganx, line 38: In a simulation of a practical ListApis request (page_size=`10`, filter=`api_id == 'my-api'`), having this case results in ~64ms response time on my local machine, as opposed to ~187ms when the database always uses `page_size + 1` to choose its database page size.

timburks: What (if anything) do we guarantee about the page sizes that get returned when filters are present? Is the "500" here because we only make one fetch per request?... so if none of those results pass the filter, could we return an empty page?

server/registry/internal/storage/list.go
```diff
@@ -130,76 +131,80 @@ var apiFields = []filtering.Field{
 	{Name: "labels", Type: filtering.StringMap},
 }
 
-func (c *Client) ListApis(ctx context.Context, parent names.Project, opts PageOptions) (ApiList, error) {
+func (c *Client) ListApis(ctx context.Context, parent names.Project, opts PageOptions) (*rpc.ListApisResponse, error) {
```
timburks, line 134: Was this change necessary to make this optimization? It seems like an orthogonal refactoring.

timburks:
@seaneganx Could you add a note somewhere (here is fine) explaining the new approach and any user-observable changes?

server/registry/internal/storage/list.go
seaneganx: I'm not totally sure I understand your question, so here's a full explanation of why this function exists.

When a database page is loaded, the list methods loop through it and ignore any rows that don't match the filter. If the response isn't filled up to the requested size, then another page is loaded from the database. This repeats until the response is full or the table has nothing left to read.

Without filters, we can load exactly the resources we need in one query. On the other hand, in cases where a filter only matches one row (or a few) in the entire table, we usually need to do a full table scan with `ceil(table_size/page_size)` queries to find all the matches. This can be a very large number of queries if the page size is too small on a large table. 

This `limit` function lets us compute a "smart" database page size for each request. As long as it returns positive integers, the listing methods will work correctly. The choice of 500 for filtered requests is somewhat arbitrary, aside from us wanting it to be large enough to avoid making too many database queries per request.

server/registry/internal/storage/list.go
seaneganx: It wasn't necessary to include. At some point while experimenting with possible listing algorithms I decided to make this change because it was convenient for something I was working on.

I can remove it if you want, but I think returning proto models from the storage package is better than returning the database models.

seaneganx:
>Could you add a note somewhere (here is fine) explaining the new approach and any user-observable changes in the response values? (e.g. Could results include uneven page sizes?)

There are no user-observable changes. This is purely an implementation change. An explanation of the approach is copied below.

>I'm not totally sure I understand your question, so here's a full explanation of why this function exists.

>When a database page is loaded, the list methods loop through it and ignore any rows that don't match the filter. If the response isn't filled up to the requested size, then another page is loaded from the database. This repeats until the response is full or the table has nothing left to read.

>Without filters, we can load exactly the resources we need in one query. On the other hand, in cases where a filter only matches one row (or a few) in the entire table, we usually need to do a full table scan with ceil(table_size/page_size) queries to find all the matches. This can be a very large number of queries if the page size is too small on a large table.

>This limit function lets us compute a "smart" database page size for each request. As long as it returns positive integers, the listing methods will work correctly. The choice of 500 for filtered requests is somewhat arbitrary, aside from us wanting it to be large enough to avoid making too many database queries per request.

server/registry/internal/storage/list.go
seaneganx: I split the tag removal into #508 and dropped the storage interface changes.

server/registry/internal/storage/list.go
```diff
@@ -57,43 +69,43 @@ func (c *Client) ListProjects(ctx context.Context, opts PageOptions) (ProjectLis
...
 			if err != nil {
-				return response, status.Error(codes.Internal, err.Error())
+				return ProjectList{}, err
+			} else if !match {
+				token.Offset++
+				continue
+			}
+
+			if len(response.Projects) == int(opts.Size) {
+				response.Token, err = encodeToken(token)
```
timburks, line 99: I'm unsure if this warrants a comment but it was the least-clear part to me...
// If we get here then we've read too much to fit in one page, so we return a token and everything up to this point.

timburks (approved):
Thank you! This seems very clear and looks correct. I don't think we need measurements to prove that this helps, but it would be interesting to take a retrospective look at list performance after this is merged.

server/registry/internal/storage/list.go
seaneganx: Right, it's definitely the most "clever" part of the code. I'll leave it as-is since we haven't had comments on it up to this point, and don't really want to add the same comment in so many different places. It's probably worth adding if we refactor this file into one general-purpose listing method.

PR #508, opened 2022-03-31
Remove revision_tags field on specs/deployments

### LINKED ISSUE #506: Registry server: Remove `revision_tags` field from ApiSpec and ApiDeployment models
Supporting this field has negative performance impact for several API methods, and increases maintenance burden for the core server. This field is nice to have, but we don't see any practical use case for it, and so we plan to remove it. If a clear need for it arises we can always add it back.

Most importantly, ListApiSpecs and ListApiDeployments retrieve a full page of resources before querying the `tags` table for all the tags associated with resources in the response, then assign those tags to their matching resources. In other cases like GetApiSpec and GetApiDeployment, a smaller impact is caused by one database query.

seaneganx:
Fixes #506.


server/registry/internal/storage/list.go
```diff
@@ -998,42 +998,3 @@ func artifactMap(artifact models.Artifact) (map[string]interface{}, error) {
 		"size_bytes":  artifact.SizeInBytes,
 	}, nil
 }
-
```
timburks, line 1001: It's unfortunate that we now don't have any way to list tags (other than raw database queries). Nonetheless, I think this is the right change to make now, and it's really helpful to have this in one commit in case we want to quickly find this code to revive it.

timburks (approved):
Thank you!

PR #510, opened 2022-04-01
Upgrade from yaml.v2 to yaml.v3 throughout the project.

timburks:
This will mainly affect YAML import/export and is preparation for other changes related to that. Since marshalling changed slightly in yaml.v3, this includes regenerated yaml test files and an internal marshalYAML function that sets the indent width to 2 (the default is 4).

Let's merge this before #503.


cmd/registry/cmd/apply/testdata/lifecycle.yaml
```diff
@@ -6,32 +6,31 @@ data:
...
-    displayOrder: 4
-  - id: deprecated
-    displayName: Deprecated
-    description: API not recommended for new consumers
-    displayOrder: 5
-  - id: retired
-    displayName: Retired
-    description: API no longer available for use
-    displayOrder: 6
+    - id: concept
```
timburks, line 9: Notice the change to dictionary indentation - this is really the only reason we had to regenerate this YAML test data. (The default indent width also changed to 4 but we set it back to 2 with the custom `marshalYAML()` in `patch/yaml.go`)

go.mod
```diff
@@ -44,6 +42,7 @@ require (
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/subosito/gotenv v1.2.0 // indirect
 	gopkg.in/ini.v1 v1.62.0 // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
timburks, line 45: I haven't looked into what's bringing this in, hoping it's gogo/protobuf (since we're eliminating that in a separate PR)

cmd/registry/patch/yaml.go
```diff
@@ -0,0 +1,30 @@
...
+package patch
+
+import (
+	"bytes"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Prefer this to yaml.Marshal because it uses tighter 2-space indentation.
+func marshalYAML(a interface{}) ([]byte, error) {
```
seaneganx, line 24: Can we reduce the responsibility of this helper? IMO the important part is that we use a common encoder. We only need a unified encoding format, not an abstraction of the `yaml` package's interface.

```
func yamlEncoder(dst io.Writer) *yaml.Encoder {
	enc := yaml.NewEncoder(dst)
	enc.SetIndent(2)
	return enc
} 
```

seaneganx (approved):

cmd/registry/patch/yaml.go
timburks: OK! done

PR #524, opened 2022-04-08
Support listing revisions across collections

### LINKED ISSUE #500: Support nested queries of resource revisions
Right now you can query lists of nested resources using `-`s in the parent name. For example, you can get a list of all versions from all APIs in the system by setting `parent` to `projects/my-project/locations/global/apis/-` in the `ListApiVersionRequest`.

It would be nice to have the ability to query a list of all `ApiSpec` revisions in a similar way to prevent the need for multiple nested calls to get a full revision history of an API.

seaneganx:
Fixes #500

### DIFF
--- a/server/registry/internal/storage/list.go
+++ b/server/registry/internal/storage/list.go
@@ -515,20 +515,48 @@ func (c *Client) ListSpecRevisions(ctx context.Context, parent names.Spec, opts 
 		return SpecList{}, status.Errorf(codes.InvalidArgument, "invalid page token %q: %s", opts.Token, err.Error())
 	}
 
+	if parent.ProjectID != "-" && parent.ApiID != "-" && parent.VersionID != "-" && parent.SpecID != "-" {
+		if _, err := c.GetSpec(ctx, parent); err != nil {
+			return SpecList{}, err
+		}
+	} else if parent.ProjectID != "-" && parent.ApiID != "-" && parent.VersionID != "-" && parent.SpecID == "-" {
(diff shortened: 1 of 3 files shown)


server/registry/internal/storage/list.go
```diff
@@ -515,20 +515,48 @@ func (c *Client) ListSpecRevisions(ctx context.Context, parent names.Spec, opts
 		return SpecList{}, status.Errorf(codes.InvalidArgument, "invalid page token %q: %s", opts.Token, err.Error())
 	}
 
+	if parent.ProjectID != "-" && parent.ApiID != "-" && parent.VersionID != "-" && parent.SpecID != "-" {
```
timburks, line 518: Can we briefly explain what is happening here? (in a comment block above this line?) I know we've used this pattern before, but I find that the differences between != and == in this list of conditions are hard to visually parse, and without explanation this feels like a puzzle to work out (I think the answer is "complain about the lowest-level ancestor that doesn't exist")

timburks (approved):

timburks:
Does "registry list projects/p/locations/global/apis/-/versions/-/specs/-@-" now list all revisions in a project?

server/registry/internal/storage/list.go
seaneganx: Done.

seaneganx:
>Does registry list projects/p/locations/global/apis/-/versions/-/specs/-@- now list all revisions in a project?

Yes, it looks like it does. I just manually verified.

Issue #535, opened 2022-04-19
Registry tool should have a version command


giteshk:
Registry Tool should display the tool build version. 
This would enable users to compare versions and update if needed.

giteshk:
Released

Issue #526, opened 2022-04-11
Create binary releases for registry server and registry tools


giteshk:
Create binary releases for:
1. Registry Tool

Platforms to support
1. Linux
2. MacOS
3. Windows

timburks:
Looking at the untarred releases, I think it might be helpful (and scalable) to have everything in a directory. Something like this:
```
- registry-0.4.6
  - README.md
  - LICENSE
  - auth
     - HOSTED.sh
     - HOSTED.bat
  - bin
     - registry
```
   


```

timburks:
I think a more generally-named "README.md" makes sense to explain the distribution.

Also, that MacOS quarantining is so annoying (I just tried a build on my MacBook) that I think we should call more attention to the fix. Maybe `README-MacOS.md`.

timburks:
I've been looking into #525 and don't think it will have a quick fix. Could we add `apg` to future distributions?

giteshk:
Released

PR #537, opened 2022-04-20
#526 : Release binaries for Registry Tools

giteshk:
1. Added a release workflow to generate Registry tool binaries (#526)
2. Added a version to the registry tools command (#535)


giteshk:
Here is a sample release with the binaries looks like:

https://github.com/giteshk-org/registry/releases/tag/v0.3.13

theganyo (approved):
LGTM

You may want to update the README to let people know that a binary is available in the releases instead of having to build.

Also, you may want to mention somewhere that on MacOS you need to unquaranteen the binary before you can run it. This can be done via `xattr -d com.apple.quarantine registry`.

timburks:
Are we following an existing convention for naming the release files?

If there's no strong precedent, I'd like to follow the convention used by [protoc releases](https://github.com/protocolbuffers/protobuf/releases)
```
registry-3.2.0-osx-x86_64.zip
registry-3.2.0-linux-x86_64.zip
registry-3.2.0-win64.zip
```

theganyo:
No particular precedent, I didn't override the goreleaser default. 
(FWIW, it looks like [Istio](https://github.com/istio/istio/releases/tag/1.13.3) uses yet another slightly different approach as well, eg. `istio-1.13.3-linux-amd64.tar.gz`)

timburks:
I see - goreleaser is the precedent then - anything built with goreleaser defaults would use these conventions? If so, this LGTM.

.goreleaser.yml
```diff
@@ -0,0 +1,51 @@
...
+  - amd64
+  env:
+  - CGO_ENABLED=0
+  ldflags:
+  - -s -w -X "github.com/apigee/registry/cmd/registry/cmd.Version={{.Env.RELEASE_VERSION}}"
+
+archives:
+- id: default
+  format: tar.gz
+  replacements:
```
timburks, line 34: For the sake of predictability, I would rather not make these replacements (going with the defaults seems more aligned with what other projects would do)

.goreleaser.yml
```diff
@@ -0,0 +1,51 @@
...
+    darwin: macOS
+    amd64: 64-bit
```
timburks, line 36: For the sake of predictability, I would rather not make these replacements (going with the defaults seems more aligned with what other projects would do)

auth/HOSTED.bat
```diff
@@ -0,0 +1,20 @@
+
```
timburks, line 1: I think this needs a license header.

cmd/registry/cmd/root.go
```diff
@@ -32,10 +32,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var Version = "dev"
```
timburks, line 35: Is this being overridden somewhere in the build process?

timburks: Here? 
```
ldflags:
  - -s -w -X "github.com/apigee/registry/cmd/registry/cmd.Version={{.Env.RELEASE_VERSION}}"
```

giteshk: Yes. we are pulling the RELEASE_VERSION from the tag name in the release.yml workflow file

seaneganx: I think we should add a comment to explain the purpose here.

auth/HOSTED.bat
seaneganx: +1

seaneganx (approved):
LGTM after open comments are addressed.

cmd/registry/cmd/root.go
timburks: Agreed, when I first saw this in the code I thought it was unfinished.

giteshk: comment added

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Prepare
+        id: prep
+        run: |
+          echo ::set-output name=version::${GITHUB_REF#refs/tags/}
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.17
```
timburks, line 39: Can you update this to 1.18 ?

.goreleaser.yml
```diff
@@ -33,7 +33,6 @@ archives:
   format: tar.gz
   replacements:
     darwin: macOS
```
timburks, line 35: can we remove this also?

timburks (approved):

PR #542, opened 2022-04-22
Merge the apg generated CLI into the registry tool.

timburks:
<strike>(For discussion only)</strike>

This contains a set of <strike>manual</strike> automated modifications to the generated `apg` source code that merges the `apg` tool into the `registry` tool. The `apg registry` subcommands are available under `registry rpc`and the `apg admin` subcommands are available as `registry rpc admin`, where the `admin` subcommand is hidden (`registry help` won't show it). Following this pattern, we might add the Provisioning service calls under `registry rpc provisioning`.

<strike>
Notes:
- The `apg registry` subcommands are under `registry .` to indicate that these are not the "main" subcommands and to minimize typing (alternately, we could use `registry .registry` but that would be painful to type).
- The `apg admin` subcommands are under `registry .admin` because "admin" is the name of the service, and this isn't a frequently-used service (so typing a bit more for this seems ok). Since cobra makes hiding commands easy, I hid the command, but we might not want to do that.
- If we added `registry .provisioning`, I think we probably would not want it to be hidden.
</strike>


seaneganx:
I think something like `registry call` is worth considering, e.g. `registry call get-api`. Other services like provisioning and admin can be subcommands, e.g. `registry call admin get-project`, `registry call provisioning get-foo`.

It's an extra 3-4 characters per command, but keeps all the API client commands in one location with a clear purpose.

shrutiparabgoogle:
Another option to explore is:

If we merge the two tools, there will be two types of operations: 
- API client operations -> interacting with the resources through the API (registry, admin and provisioning)
-  Control operations -> operations like `resolve`, `uploading` definitions, `apply`.

So we can split the high level command options into these categories:
- `registry client` -> includes all `apg registry` commands
- `registry admin` -> includes all `apg admin` commands
- `registry provision` -> includes all provisioning API commands
- `registry ctl` -> includes all the current commands from the registry tool (this will be similar to `kubectl` or `apigeectl`)
    The commands will look like `registry ctl resolve ...`, `registry ctl apply ...`, `registry ctl compute ...`

timburks:
Thanks for all the feedback and offline discussion. Based on our conversations, I've reorganized this as follows:

- First, all generated RPC-related subcommands are now under `rpc`. This is similar to `call` proposed by @seaneganx but has the added specificity of saying that we're making an rpc call, which is mainly to alert users that these are low-level operations.
- All of the registry service commands are directly under `rpc`, e.g. `registry rpc list-apis`.
- The admin service commands are under `registry rpc admin`, which is hidden.
- In the near future, provisioning subcommands will be under `registry rpc provisioning` (or `registry rpc provision`) and will also be hidden.

`admin` and `provision` are hidden to discourage casual usage.

We've also discussed renaming the `registry` tool to `regctl`, which will be addressed in a separate PR.

Note that this PR includes manual modifications to the `apg` generated code. The biggest change is the package name, which should be easy to automatically patch.

seaneganx:
>Note that this PR includes manual modifications to the apg generated code. The biggest change is the package name, which should be easy to automatically patch.

The harder part to patch seems to be the removal of viper config from each service's init functions. I would like to see that added to GENERATE-APG.sh with an explanation about why it's necessary. I agree that package rename should be an easy replacement.

Can we also move the generated code into `cmd/registry/cmd/rpc` to be consistent with other subcommands?

timburks:
> The harder part to patch seems to be the removal of viper config from each service's init functions. I would like to see that added to GENERATE-APG.sh with an explanation about why it's necessary. I agree that package rename should be an easy replacement.
> Can we also move the generated code into cmd/registry/cmd/rpc to be consistent with other subcommands?

Yes, I think both of these should be done - I'd like to make those behind-the-scenes changes in a followup PR.

timburks:
@seaneganx the CLI-related patches are now automated.

cmd/registry/cmd/root.go
```diff
@@ -57,6 +60,23 @@ func Command() *cobra.Command {
 	cmd.AddCommand(list.Command())
 	cmd.AddCommand(upload.Command())
 	cmd.AddCommand(vocabulary.Command())
+	cmd.AddCommand(RPCCommand())
+	return cmd
+}
+
+func RPCCommand() *cobra.Command {
```
seaneganx, line 67: Can this be moved into `cmd/registry/cmd/rpc` too? I think it would provide a nicer boundary between code that we fully own/control, and code that requires knowledge of the GAPIC generator output.

seaneganx (changes requested):

tools/GENERATE-CLI.sh
```diff
@@ -29,17 +29,33 @@ protoc ${SERVICE_PROTOS[*]} \
...
+# to avoid confusion. These flags are not supported by the non-generated subcommands
+# of the main tool.
+REGISTRY_SERVICE=cmd/registry/cmd/rpc/registry_service.go
+sed -i.bak "s/^\W*RegistryServiceCmd.PersistentFlags.*$//g" "${REGISTRY_SERVICE}"
+sed -i.bak "s/^\W*RegistryConfig\.Bind.*$//g" "${REGISTRY_SERVICE}"
+
+# Remove local connection configuration flags from the generated Admin service CLI.
+ADMIN_SERVICE=cmd/registry/cmd/rpc/admin_service.go
+sed -i.bak "s/^\W*AdminServiceCmd.PersistentFlags.*$//g" "${ADMIN_SERVICE}"
+sed -i.bak "s/^\W*AdminConfig\.Bind.*$//g" "${ADMIN_SERVICE}"
```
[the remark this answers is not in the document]
timburks: Those work on linux - updated.

tools/GENERATE-CLI.sh
```diff
@@ -29,17 +29,33 @@ protoc ${SERVICE_PROTOS[*]} \
...
 # Admin service configuration variables. This causes the Admin service
 # client to be configured with the same variables that configure the
 # Registry service client.
-FILE=cmd/apg/admin_service.go
-sed -i.bak "s/APG_ADMIN/APG_REGISTRY/" "${FILE}"
-rm "${FILE}.bak"
-gofmt -w "${FILE}"
-if grep --quiet APG_ADMIN "${FILE}"; then
+sed -i.bak "s/APG_ADMIN/APG_REGISTRY/" "${ADMIN_SERVICE}"
+gofmt -w "${ADMIN_SERVICE}"
```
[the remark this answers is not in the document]
timburks: SGTM - done.

cmd/registry/cmd/root.go
timburks: It could, but I am reluctant to do that because all the other go files in the `rpc` directory are automatically generated. I've left this as-is, or would be fine to put the rpc-related code in a separate file if you prefer.

seaneganx: I have the same feeling of reluctance, wanting to have clear separation of the generated code from the handwritten integration code.

On the other hand, I have a similar feeling leading my suggestion to move this function. I want to have a clear separation of code that we control from code that relies on the generated commands. This feels more important to me because it affects where the errors will appear if the integration breaks on regen.

There's a middle path that might balance these desires: where we have package `cmd/registry/cmd/rpc` with a single file defining the `rpc` subcommand, and a second package holding the generated code. It would make for a very lightweight package, but I don't think there's anything wrong with that.

timburks: I wouldn't strongly object to this in a future PR, but think we're talking about diminishing returns now, particularly since the new package would be <20 lines of code.

seaneganx (approved):

timburks:
Thanks @seaneganx , I'll do that in the update that adds the Provisioning API.

PR #543, opened 2022-04-22
consolidate protoc version, fail fast in make

theganyo:
A few minor improvements that will hopefully help:

1. consolidated the places that the protoc version was declared into a single file
2. `make all` will fail fast with an error if `protoc` isn't present on the path
3. made the `tool/*.sh` files that were not executable, executable


theganyo:
Oh, another thing I noticed is that we're using `go install`... but there seems to be an issue on MacOS where the GOPATH/bin is not automatically set on the PATH. I worked around this on my box by running this:

`export PATH=$PATH:$(go env GOPATH)/bin`

I didn't address this in the PR, but it could potentially be addressed in the scripts or mentioned in the docs.

tools/PROTOC-VERSION.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+#
+# Copyright 2020 Google LLC. All Rights Reserved.
```
timburks, line 3: nit: 2022

tools/FETCH-PROTOC.sh
```diff
@@ -27,12 +27,12 @@ case "$(arch)" in
   ;;
 esac
 
-export VERSION="3.19.3"
+. tools/PROTOC-VERSION.sh
```
timburks, line 30: Is "source" equivalent to "." ? We use that elsewhere in the project (maybe everywhere else?)

timburks (approved):
Thanks!

Could you verify that tools/pretty.sh isn't getting stomped by this PR? I'm puzzled as to why it's listed as an empty file among the other changes.

Also, I think it would be fine to update the protoc version but would like to do that in a standalone PR.

tools/FETCH-PROTOC.sh
timburks: Thanks - If you don't mind, let's go for consistency (short term) and do the replacement as a unit.

theganyo: I could do that, but unfortunately, that will break it... it's the reason why I used `.` in the first place.

theganyo: Well, actually, now that I think about it... it might've been because I used `sh xxxx` to execute something. I'd have to go back a re-test to see what may or may not break...

theganyo:
Also: This PR was not associated with an issue. Would you like me to open an issue that this PR closes?

tools/FETCH-PROTOC.sh
theganyo: I've opened the [issue](https://github.com/apigee/registry/issues/545). If you'd prefer, I can remove this change from this issue and can fix it in the other. Or, I can remove this entirely and you can close the issue if you just don't like it.

theganyo:
Opened related [issue for protoc version](https://github.com/apigee/registry/issues/544)

PR #547, opened 2022-04-25
Format APIHUB-README.md with ./tools/pretty.sh

timburks:
Let's run `./tools/pretty.sh` to automatically format any edits to markdown files.


giteshk (approved):
We should add https://github.com/github/super-linter to our workflow to lint all file types

PR #548, opened 2022-04-25
Update protoc version to v3.20.1

### LINKED ISSUE #544: Update protoc version to v3.20.1
Latest Protoc is v3.20.1: https://github.com/protocolbuffers/protobuf/releases.

Related: Only protoc versions 3.20.0+ have binary releases for MacOS on M1/ARM, so this will improve MacOS installation.

theganyo:
Fixes #544


timburks:
https://github.com/apigee/registry/blob/30c1d99aaefe83e061e6abd299ee6d8aeb3ebb01/deployments/envoy/Dockerfile#L1

.github/workflows/docker-publish.yml
```diff
@@ -92,3 +96,4 @@ jobs:
         image: ${{ env.DOCKER_REPOSITORY_OWNER }}/registry-envoy
         context: .
         file: deployments/envoy/Dockerfile
+        build-args: PROTOBUF_VERSION=${{ env.PROTOC_VERSION }}
```
timburks, line 99: This seems fragile since it would only set the version properly when we run the workflow. I wonder - could we get the protoc version within the Dockerfile by reading tools/PROTO_VERSION.sh ?

theganyo: See line # 88. I'm setting from the file.

theganyo: Now that we're overriding it, I wonder if we should remove the default PROTOBUF_VERSION ARG value from the Dockerfile? Or just update it to match?

timburks: I was wondering why it couldn't be set here: https://github.com/apigee/registry/blob/30c1d99aaefe83e061e6abd299ee6d8aeb3ebb01/deployments/envoy/Dockerfile#L1

timburks: We could just run ./tools/FETCH-PROTOC.sh in the Dockerfile

theganyo: Yes, I thought of that, but wasn't sure you wanted to get rid of the PROTOBUF_VERSION ARG. Now that I think of it, though, there's probably no reason to keep it. Ok, I'll do as you suggested.

timburks: Let's get rid of the arg, yes. I don't think it's helping anywhere.

theganyo: Done.

deployments/envoy/Dockerfile
```diff
@@ -1,17 +1,13 @@
...
 
 WORKDIR /app
 
-RUN apt-get update -y && apt-get install curl unzip git make -y && \
-    PB_REL="https://github.com/protocolbuffers/protobuf/releases" && \
-    curl -LO $PB_REL/download/v${PROTOBUF_VERSION}/protoc-${PROTOBUF_VERSION}-linux-x86_64.zip && \
-    unzip protoc-${PROTOBUF_VERSION}-linux-x86_64.zip -d /usr/local
+RUN apt-get update -y && apt-get install curl unzip git make -y
 COPY . .
+RUN ./tools/FETCH-PROTOC.sh
```
timburks, line 10: beautiful! :)

timburks (approved):

PR #549, opened 2022-04-26
Move details for running registry-server to cmd/registry-server/README.md

timburks:
This moves the details of various ways to run `registry-server` to cmd/registry-server/README.md, which makes the top-level README more readable and sets a pattern that we can follow to add more detail to cmd/registry/README.md.


theganyo (approved):
I like it, I was thinking that the README was a bit much. Would it make sense to have a link to the registry-server/README in the top-level README?

README.md
```diff
@@ -63,7 +63,8 @@ Two command-line interfaces are included:
   Go GAPIC library to support additional API management tasks.
 
 The entry point for the Registry API server itself is
-[cmd/registry-server](cmd/registry-server).
+[cmd/registry-server](cmd/registry-server). For more on running the server, see
+[cmd/registry-server/README.md](cmd/registry-server/README.md).
```
timburks, line 67: @theganyo added per your suggestion

PR #552, opened 2022-04-26
Add support for compute score command

shrutiparabgoogle:
Summary of changes:
* Added support for command `registry compute score`.
* TODOs:
   - timestamp based check for artifact dependencies in score calculation
   - support for dry_run flag
   - support for RollUp score
   - Support for json artifacts


cmd/registry/scoring/expression.go
```diff
@@ -0,0 +1,145 @@
...
+		return unmarshalAndStruct(contents, &rpc.Score{})
+	case "google.cloud.apigeeregistry.v1.scoring.ScoreCard":
+		return unmarshalAndStruct(contents, &rpc.ScoreCard{})
+	// TODO: Add support for JSON artifacts
+	default:
+		return nil, fmt.Errorf("unsupported artifact type: %s", messageType)
+	}
+}
+
+func unmarshalAndStruct(contents []byte, message proto.Message) (*structpb.Struct, error) {
```
seaneganx, line 125: It doesn't look like we use the `Struct` type for anything except converting it into a map. Can we update this to return `map[string]interface{}`? 

We can replace the protojson unmarshal step with a regular JSON unmarshal to skip converting to/from the struct type if our end goal is the map. I think working with maps will make the tests a lot cleaner too.

cmd/registry/scoring/expression_test.go
```diff
@@ -0,0 +1,277 @@
...
+						FilePath: "openapi.yaml",
+						Problems: []*rpc.LintProblem{
+							{
+								Message: "lint-error",
+							},
+						},
+					},
+				},
+			},
+			wantValue: 1,
```
seaneganx, line 158: Since the type of this value is also part of the test, consider making this `int(1)` to make it clear we expect a normal `int` instead of `int64` or other variants.

cmd/registry/scoring/score_test.go
```diff
@@ -0,0 +1,648 @@
...
+			defer adminClient.Close()
+
+			test.setup(ctx, registryClient, adminClient)
+
+			gotValue, gotErr := processScoreFormula(ctx, registryClient, test.formula, test.resource)
+			if gotErr != nil {
+				t.Errorf("processScoreFormula() returned unexpected error: %s", gotErr)
+			}
+
+			if gotValue != nil && test.wantValue != gotValue.(int) {
```
seaneganx, line 300: This will panic if the returned value isn't an `int`. If we want to give more info, there should be type cast with an error check before this point, otherwise we can drop the `.(int)` and just compare `wantValue` and `gotValue` directly without converting them from `interface{}`.

cmd/registry/scoring/score_test.go
```diff
@@ -0,0 +1,648 @@
...
+
+			if !test.wantErr && gotErr != nil {
+				t.Errorf("matchResourceWithTarget() returned unexpected error: %s", gotErr)
+			}
+		})
+	}
+}
+
+func TestProcessScoreFormula(t *testing.T) {
+	tests := []struct {
```
seaneganx, line 224: There's only one test in this table. Unless you have immediate plans to add tests here, consider removing the table and having the test case on its own in the function.

cmd/registry/scoring/score.go
```diff
@@ -0,0 +1,366 @@
...
+	if fmt.Sprintf("%T", targetPatternName) != fmt.Sprintf("%T", resourceName) {
+		return fmt.Errorf("resource %s doesn't match target pattern %v", resourceName.String(), targetPattern)
+	}
+
+	// Check if the individual entities match
+	switch targetPatternName.(type) {
+	case patterns.SpecName:
+		// type casting because we need access to "Name" field of patterns.SpecName to compare specific IDs
+		specPattern := targetPatternName.(patterns.SpecName)
+		specResource := resourceName.(patterns.SpecName)
```
seaneganx, line 84: How certain are we that these are always the same type? The resource name isn't type checked, and will panic if it doesn't match.

seaneganx (commented):
Most of these comments are just nitpicks, but there are some more important comments in `scoring/expression.go` that I think will simplify the code quite a bit.

cmd/registry/scoring/expression.go
```diff
@@ -0,0 +1,145 @@
...
+		expectedType := nativeTypeMap["bool"]
+		nativeVal, err := out.ConvertToNative(expectedType)
+		if err != nil {
+			return nil, fmt.Errorf("failed converting output from type %s to %s: %s", out.Type().TypeName(), expectedType, err)
+		}
+		return nativeVal, nil
+	default:
+		return nil, fmt.Errorf("evaluating expression %q generated an unexpected output type %s: should be one of [int, float, bool]", expression, out.Type().TypeName())
+
+	}
```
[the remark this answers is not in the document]
shrutiparabgoogle: Thanks for this suggestion. Done.

cmd/registry/scoring/expression.go
shrutiparabgoogle: Thanks for this suggestion. Done!

cmd/registry/scoring/expression_test.go
shrutiparabgoogle: Done. With the updated code using `ref.Val.Value()`, evaluateScoreExpression will return either an `int64`, `float64` or `bool`.

cmd/registry/scoring/score_test.go
```diff
@@ -0,0 +1,648 @@
...
+		{
+			desc: "artifact pattern error",
+			targetPattern: &rpc.ResourcePattern{
+				Pattern: "apis/-/versions/-/specs/-/artifacts/lint-spectral",
+			},
+			resourceName: "projects/pattern-test/locations/global/apis/petstore/versions/1.0.0/specs/openapi.yaml/artifacts/lint-spectral",
+			wantErr:      true,
+		},
+		{
+			desc: "target and resource mismatch",
```
shrutiparabgoogle, line 199: test case for targetPattern and resource mismatch

cmd/registry/scoring/score.go
```diff
@@ -0,0 +1,366 @@
...
+		Uri:            definition.GetUri(),
+		UriDisplayName: definition.GetUriDisplayName(),
+		DefinitionName: fmt.Sprintf("%s/artifacts/%s", project, definition.GetId()),
+	}
+
+	// Set the Value field according to the type
+	switch definition.GetType().(type) {
+	case *rpc.ScoreDefinition_Integer:
+		// Score proto expects int32 type
+		value := int32(0)
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

cmd/registry/scoring/score.go
```diff
@@ -0,0 +1,366 @@
...
+		// Convert scoreValue to appropriate type
+		// evaluateScoreExpression can return either a float or int value.
+		// Both are valid for an integer.
+		if intVal, ok := scoreValue.(int64); ok {
+			value = int32(intVal)
+		} else if floatVal, ok := scoreValue.(float64); ok {
+			value = int32(floatVal)
+		} else {
+			return nil, fmt.Errorf("failed typecheck for output: expected either int or float64 got %s", scoreValue)
+		}
```
seaneganx, line 239: I think a type switch is easier to visually parse for something like this.

```
switch v := scoreValue.(type) {
case int64, float64:
	value = int32(v)
default:
	return nil, fmt.Errorf("failed typecheck for output: expected either int64 or float64 got %T %s", v, v)
}
```

Also, I think the error message can be more specific about the types it expects, and show what type it received (rather than just the value).

Same for percentages below.

cmd/registry/scoring/score.go
```diff
@@ -0,0 +1,366 @@
...
+			} else {
+				displayValue = "true"
+			}
+		} else {
+			if configuredDisplay := definition.GetBoolean().GetDisplayFalse(); len(configuredDisplay) > 0 {
+				displayValue = configuredDisplay
+			} else {
+				displayValue = "false"
+			}
+		}
```
seaneganx, line 328: I think this can be improved by flattening it out.

```
var displayValue string
if t := definition.GetBoolean().GetDisplayTrue(); boolVal && t != "" {
	displayValue = t
} else if f := definition.GetBoolean().GetDisplayFalse(); !boolVal && f != "" {
	displayValue = f
} else {
	displayValue = strconv.FormatBool(boolVal)
}
```

cmd/registry/scoring/expression.go
```diff
@@ -0,0 +1,119 @@
...
+	case "google.cloud.apigeeregistry.v1.scoring.ScoreCard":
+		return unmarshalAndMap(contents, &rpc.ScoreCard{})
+	// TODO: Add support for JSON artifacts
+	default:
+		return nil, fmt.Errorf("unsupported artifact type: %s", messageType)
+	}
+}
+
+func unmarshalAndMap(contents []byte, message proto.Message) (map[string]interface{}, error) {
+	mapValue := make(map[string]interface{}, 0)
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

cmd/registry/scoring/expression_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			contents, _ := proto.Marshal(test.contentsProto)
+
+			gotMap, gotErr := getMap(contents, test.mimeType)
+			if gotErr != nil {
+				t.Errorf("getMap() returned unexpected error: %s", gotErr)
+			}
+			opts := cmp.Options{protocmp.Transform()}
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

cmd/registry/scoring/expression_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+func TestEvaluateScoreExpressionError(t *testing.T) {
+	tests := []struct {
+		desc          string
+		expression    string
+		mimeType      string
+		contentsProto proto.Message
+	}{
+		{
+			desc:          "error in structProto",
```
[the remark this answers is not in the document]
shrutiparabgoogle: Updated the description.

cmd/registry/scoring/score.go
shrutiparabgoogle: Done. I will do the error message refactor in a separate PR.  I think that can be done more broadly across the validation functions and it's tests.

cmd/registry/scoring/score.go
shrutiparabgoogle: This switch case block fails at typecasting `int32(v)`. I think this is happening because the type of `v` is known only at runtime and this code fails to build because at build time, the type of `v` is still `interface{}`. Even in the `switch` block, I will need a typecheck, which will bring it close to the current `if else` structure. Example: https://go.dev/play/p/zWNXMHsJuON

Updated the error message as suggested.

cmd/registry/scoring/score.go
shrutiparabgoogle: Done. Thanks for the suggestion!

cmd/registry/scoring/score.go
seaneganx: I suppose that's what I get for writing that out without running it. Type switches in Go automatically typecast the switch variable (if one is defined, `v` in this case) to the type in each switch case. My mistake was thinking it would know that `int32(v)` works in both cases.

Splitting it into separate cases should work as intended.
```
switch v := scoreValue.(type) {
case int64:
	value = int32(v)
case float64:
	value = int32(v)
default:
	return nil, fmt.Errorf("failed typecheck for output: expected either int64 or float64 got %T %s", v, v)
}
```

shrutiparabgoogle: Thanks for clarifying. It didn't strike me to try for different cases. 
Works as expected!

seaneganx (approved):
I still have some open comments about improving test descriptions, but I won't block the PR for that. LGTM overall.

cmd/registry/scoring/expression_test.go
```diff
@@ -0,0 +1,266 @@
...
+			expression:    "",
+			mimeType:      "application/octet-stream;type=google.cloud.apigeeregistry.v1.scoring.ScoreDefinition",
+			contentsProto: &rpc.ScoreDefinition{},
+		},
+		{
+			desc:       "error in expression",
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

cmd/registry/scoring/expression_test.go
```diff
@@ -0,0 +1,266 @@
...
+								Message: "lint-error",
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			desc:       "unsupported type",
+			expression: "files[0].problems", //this will return a list value
```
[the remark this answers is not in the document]
shrutiparabgoogle: Done.

shrutiparabgoogle:
> I still have some open comments about improving test descriptions, but I won't block the PR for that. LGTM overall.

The test descriptions were fixed in [this](https://github.com/apigee/registry/pull/552/commits/a3a7aa0640a48f59c8d3fde1e2791449e65604ef#diff-4d809861e18c54411199dc95b93bc7afdc8abf9b50c1c48939ec99f9fec05694) commit. I missed resolving the comments.

cmd/registry/cmd/compute/score.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/log"
+	"github.com/apigee/registry/rpc"
+	"github.com/spf13/cobra"
+)
+
+func scoreCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "score (experimental)",
+		Short: "Compute score from ScoreDefinition",
```
timburks, line 18: Could we make this "Compute scores for APIs and API specs" to match the style of other help strings in this section?

cmd/registry/cmd/compute/score.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/apigee/registry/cmd/registry/scoring"
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/log"
+	"github.com/apigee/registry/rpc"
+	"github.com/spf13/cobra"
+)
+
+func scoreCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "score (experimental)",
```
timburks, line 17: I don't think we should use "(experimental)" here since this is building out an agreed-upon design.

timburks (approved):
Just two minor comments about the help strings

cmd/registry/cmd/compute/score.go
shrutiparabgoogle: I was planning to take it out once the tool was ready. Basically something to represent that this is WIP. I had seen the use of term `experimental` before and thought of following the same. I can change it to something else.

cmd/registry/cmd/compute/score.go
shrutiparabgoogle: Done.

cmd/registry/cmd/compute/score.go
shrutiparabgoogle: Updated this to `(wip)` instead.

PR #560, opened 2022-04-29
Use contexts for database operations during API requests

seaneganx:


timburks (approved):
LGTM - tangibly, what do we get for this? cancellation?

seaneganx:
>LGTM - tangibly, what do we get for this? cancellation?

As far as I can tell it just adds cancellation/timeout support, which we don't use much at all. 

The reason this came up is that I ran a linter for unused variables in function signatures and noticed every `ctx` in the storage layer was unused. Instead of leaving them unused or removing them, I decided to use it.

PR #564, opened 2022-05-03
Registry tool: Restructure and simplify artifact YAML support code.

timburks:


cmd/registry/cmd/upload/artifact.go
```diff
@@ -83,23 +83,23 @@ func buildArtifact(ctx context.Context, parent string, filename string) (*rpc.Ar
 	jsonBytes, _ := yaml.YAMLToJSON(yamlBytes) // to use protojson.Unmarshal()
 	var artifact *rpc.Artifact
 	switch header.Kind {
-	case "DisplaySettings", patch.DisplaySettingsMimeType:
```
timburks, line 86: Including the mime types in the Kind field was (IMO) a mistake - I think if we include anything other than the message name, it should be the fully-qualified message name to resolve any ambiguities (which don't exist now because currently all message names are unique)

cmd/registry/patch/apply.go
```diff
@@ -56,9 +55,7 @@ func applyFile(ctx context.Context, client connection.Client, fileName, parent s
 	switch header.Kind {
 	case "API":
 		return applyApiPatch(ctx, client, bytes, parent)
-	case "DisplaySettings", "Lifecycle", "Manifest", "ReferenceList", "TaxonomyList":
+	default: // for everything else, try an artifact type
```
timburks, line 58: Future cases here might include "Deployment", "ApiVersion", and "ApiSpec".

cmd/registry/cmd/apply/testdata/lifecycle.yaml
```diff
@@ -9,28 +9,35 @@ data:
     - id: concept
       displayName: Concept
       description: Description of the business case and user needs for why an API should exist
+      url: ""
```
timburks, line 12: All these fields are here because YAML export now includes default values for fields. The value of including empty strings is unclear, but zero-valued ints (displayOrder, below) seem important to include to allay (naive) concerns that they might be unset. The controls for this are not type-specific, so whatever we do for ints applies to strings, bools, etc.

cmd/registry/cmd/upload/artifact.go
```diff
@@ -124,7 +124,7 @@ func buildDisplaySettingsArtifact(ctx context.Context, jsonBytes []byte) (*rpc.A
 	}
 	return &rpc.Artifact{
 		Contents: artifactBytes,
-		MimeType: patch.DisplaySettingsMimeType,
+		MimeType: patch.MimeTypeForKind("DisplaySettings"),
```
timburks, line 127: This is admittedly a bit uglier, but is because further down, in the patch package, all the information about supported artifact types is consolidated into one place. I didn't (in this PR) try to simplify this file further, but I expect that it can be refactored similarly to patch/artifact.go to remove these type-specific build functions.

cmd/registry/patch/artifact.go
```diff
@@ -60,69 +59,98 @@ func ExportArtifact(ctx context.Context, client *gapic.RegistryClient, message *
...
+	if err = proto.Unmarshal(message.Contents, m); err != nil {
+		return nil, err
+	}
+	// Marshal the artifact content as JSON using the protobuf marshaller.
+	var s []byte
+	s, err = protojson.MarshalOptions{
+		UseEnumNumbers:  false,
+		EmitUnpopulated: true,
+		Indent:          "  ",
+		UseProtoNames:   false,
```
seaneganx, line 124: Are our protos still intended to be the source of truth for structure and documentation? If so, using the same field names seems like it would eliminate a lot of mental burden for users.

go.sum
```diff
@@ -315,7 +315,6 @@ github.com/ianlancetaylor/demangle v0.0.0-20200824232613-28f6c0f3b639/go.mod h1:
 github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NHg9XEKhtSvM=
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/influxdata/influxdb1-client v0.0.0-20191209144304-8bf82d3c094d/go.mod h1:qj24IKcXYK6Iy9ceXlo3Tc+vtHo9lIhSX5JddghvEPo=
-github.com/jackc/chunkreader v1.0.0 h1:4s39bBR8ByfqH+DKm8rQA3E1LHZWB9XWcrz8fqaZbe0=
```
seaneganx, line 318: Any idea why there were go.sum changes without any go.mod changes? If you haven't already, run `go mod tidy` to make sure they're both in sync.

cmd/registry/patch/artifact.go
```diff
@@ -159,3 +205,66 @@ func applyArtifactPatch(ctx context.Context, client connection.Client, content *
...
+	for k, v := range artifactMessageTypes {
+		if strings.HasSuffix(k, "."+kind) {
+			return v(), nil
+		}
+	}
+	return nil, fmt.Errorf("unsupported kind %s", kind)
+}
+
+// messageTypeForKind finds the full message type associated with a kind (typically a message name).
+func messageTypeForKind(kind string) string {
```
seaneganx, line 237: Consider merging this with `MimeTypeForKind` since that's the sole user of this function. Among other benefits, it becomes obvious that an empty type can be returned.

```
func MimeTypeForKind(kind string) string {
	for k := range artifactMessageTypes {
		if strings.HasSuffix(k, "."+kind) {
			return fmt.Sprintf("application/octet-stream;type=%s", k)
		}
	}
	return "application/octet-stream;type="
}
```

cmd/registry/patch/artifact.go
```diff
@@ -159,3 +205,66 @@ func applyArtifactPatch(ctx context.Context, client connection.Client, content *
...
+	return fmt.Sprintf("application/octet-stream;type=%s", messageTypeForKind(kind))
+}
+
+// messageFactory represents functions that construct message structs.
+type messageFactory func() proto.Message
+
+// artifactMessageTypes is the single source of truth for artifact types that can be represented in YAML.
+var artifactMessageTypes map[string]messageFactory
+
+func init() {
```
seaneganx, line 257: Is this `init` function necessary? Why can't we just declare the map in the var?

seaneganx (approved):
Some concerns, but overall LGTM

cmd/registry/patch/artifact.go
timburks: Done! I noticed that it wasn't possible to declare this as a const but didn't try the package-level var.

go.sum
timburks: No idea - I tried `go mod tidy` earlier and again just now, getting the same result.

cmd/registry/patch/artifact.go
timburks: Noted. I prefer this because I think there are consumption-side benefits to using protos (generated support code) and user-side benefits to following YAML conventions and that the camel-to-snake conversion isn't that hard. But I'm open to keeping an eye on this.

cmd/registry/patch/artifact.go
timburks: Done, but changing the fall-through return type to just "application/octet-stream"

Issue #565, opened 2022-05-03
Registry Tool: Failures in test setup are scoped incorrectly


seaneganx:
Tests that define setup functions for each test case are passing the parent test's `t` argument into helpers that are used to set up child subtests. Instead, each subtest should pass its `t` into the setup helper to scope the errors to the subtest.

For example, if a failure occurs during setup we currently get a message like this:
```
--- FAIL: TestArtifacts (0.01s)
    controller_test.go:60: something went wrong
    --- FAIL: TestArtifacts/single_spec (0.01s)
        testing.go:1336: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestAggregateArtifacts (0.01s)
    controller_test.go:220: something went wrong
    --- FAIL: TestAggregateArtifacts/create_artifacts (0.01s)
        testing.go:1336: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestDerivedArtifacts (0.01s)
    controller_test.go:352: something went wrong
    --- FAIL: TestDerivedArtifacts/create_artifacts (0.01s)
        testing.go:1336: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestReceiptArtifacts (0.01s)
    controller_test.go:531: something went wrong
    --- FAIL: TestReceiptArtifacts/create_artifacts (0.01s)
        testing.go:1336: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestReceiptAggArtifacts (0.01s)
    controller_test.go:620: something went wrong
    --- FAIL: TestReceiptAggArtifacts/create_artifacts (0.01s)
        testing.go:1336: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestMultipleEntitiesArtifacts (0.02s)
    controller_test.go:730: something went wrong
    --- FAIL: TestMultipleEntitiesArtifacts/create_artifacts (0.02s)
        testing.go:1336: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

seaneganx:
[...]

PR #566, opened 2022-05-05
Add GAPIC client adapter for seeding non-server tests

### LINKED ISSUE #441: Testing: Use a shared seeder package for test setup in non-server tests
### Background
Server tests have a [seeder package](https://github.com/apigee/registry/tree/4b43762e57f64242cad80b8c7d2e3add157fde2d/server/registry/internal/test/seeder) making it convenient to initialize the server with resources for a given test. The seeder is declarative, making it easy to concisely represent any desired state. The seeder creates default resources in cases where a parent resource is required but not declared, e.g. when seeding a spec without specifying the parent api or version.

Non-server tests, e.g. registry tool [command tests](https://github.com/apigee/registry/blob/4b43762e57f64242cad80b8c7d2e3add157fde2d/cmd/registry/cmd/upload/csv_test.go), do not have access to a similar package. Currently server setup is handled independently by each set of tests ([1](https://github.com/apigee/registry/blob/4b43762e57f64242cad80b8c7d2e3add157fde2d/cmd/registry/cmd/upload/csv_test.go#L130), [2](https://github.com/apigee/registry/blob/4b43762e57f64242cad80b8c7d2e3add157fde2d/cmd/registry/cmd/annotate/annotate_test.go#L53), [3](https://github.com/apigee/registry/blob/4b43762e57f64242cad80b8c7d2e3add157fde2d/cmd/registry/controller/test_helpers.go)) resulting in code duplication and inconsistency between tests. Additionally, these tend to use procedural seeding which increases verbosity and decreases readability compared to declarative seeding.

### Ideas

Importantly, non-server tests interact with the server using two clients: `connection.AdminClient` for project resources, and `connection.Client` for all other resources. This differs from the server tests which use `registry.RegistryServer` for every resource.

The non-server tests also run against a single shared instance of the server and database, so resource seeding must work correctly in the presence of pre-existing resources from prior tests. Most tests handle this by using different projects to avoid collisions, and by deleting that project before seeding to "reset" the server state. The existing server seeder doesn't take this into consideration.

1. Create a separate seeding package for non-server tests with a similar interface to the existing seeder.
**Pro**: The package can be designed specifically for the non-server environment.
**Con**: Similar code will exist in both packages.

2. Create a composite client type for non-server tests so they can reuse the existing seeding package.
**Pro**: Tests can reuse the existing seeder package with little extra code.
**Con**: All functionality offered by the package will need to be environment independent.

We can also use a hybrid of those ideas. Specifically, we can create a separate seeding package for non-server tests including an internal/unexported composite client. The new package can handle functionality specific to non-server tests like cleaning up pre-existing resources, then reuse the existing package with the composite client.

seaneganx:
Some changes to the existing seeder were necessary to make it work for non-server tests.

* Export the seeder to give non-server tests access to the package.
* Remove "safety features" that prevent multiple updates to the same resource.

When this package was originally created I didn't expect us to have tests that depended on the update times of resources. The controller tests depend on this very frequently, so it makes more sense for the seeder to allow multiple updates to the same resource. I included one example of how the seeder can be used in tests.

Fixes #441. If fully adopted, it will also make #565 obsolete.


shrutiparabgoogle (approved):
This looks great! It will helpful in `scoring` tests as well.

timburks (approved):

PR #569, opened 2022-05-17
Add support for ScoreCard generation

shrutiparabgoogle:
In this PR:
Support for generating Scorecards

TODOs:
- `dry_run` flag across both the commands
- seeder test pattern for both commands


cmd/registry/cmd/compute/scorecard.go
```diff
@@ -0,0 +1,89 @@
...
+	defArtifact *rpc.Artifact
+	resource    patterns.ResourceInstance
+}
+
+func (task *computeScoreCardTask) String() string {
+	return "compute score " + task.resource.ResourceName().String()
+}
+
+func (task *computeScoreCardTask) Run(ctx context.Context) error {
+	return scoring.CalculateScoreCard(task.ctx, task.client, task.defArtifact, task.resource)
```
seaneganx, line 88: This should use the context passed into the `Run` method.

timburks (approved):
LGTM - not blocking, but it would be useful to see some examples that are outside of the boilerplate of our test infrastructure and closer to the the actual user experience (YAML to YAML instead of test structs to test structs)

seaneganx (approved):
Approved, but please address my comment about the context.

cmd/registry/cmd/compute/scorecard.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/apigee/registry/cmd/registry/scoring"
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/log"
+	"github.com/apigee/registry/rpc"
+	"github.com/spf13/cobra"
+)
+
+func scoreCardCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "scorecard (wip)",
```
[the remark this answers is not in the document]
shrutiparabgoogle: Thanks for catching this. I took this out from `compute score` command but missed taking it out from `scorecard`.

Done.

cmd/registry/cmd/compute/scorecard.go
shrutiparabgoogle: Done.

shrutiparabgoogle:
> LGTM - not blocking, but it would be useful to see some examples that are outside of the boilerplate of our test infrastructure and closer to the the actual user experience (YAML to YAML instead of test structs to test structs)

Yes, currently the examples that we have are through the unit and functional tests. I can create another e2e test which provide examples and also test the whole flow.

PR #571, opened 2022-05-18
Rename the registry tool to "regctl".

timburks:
See #572 for explanation and any followup discussion


cmd/regctl/cmd/apply/testdata/manifest.yaml
```diff
@@ -12,7 +12,7 @@ data:
       dependencies:
         - pattern: $resource.spec
           filter: mime_type.contains('openapi')
-      action: registry compute lint $resource.spec --linter spectral
```
timburks, line 15: @shrutiparabgoogle renaming `registry` forces us to update the manifest actions, which might be hard to change. What would you think about making the controller resilient to this by having it automatically replace "registry" with "regctl" (when it is at the start of an action string)?

timburks: Right - I agree that we should fix the manifests also.

timburks: Addressed here: https://github.com/apigee/registry/pull/571/commits/b52caed59a1d8a34268683bebdd2e8f71a8581c9

cmd/regctl/controller/local-tasks.go
```diff
@@ -57,13 +57,19 @@ func (task *ExecCommandTask) Run(ctx context.Context) error {
 		"taskID": fmt.Sprintf("{%s}", task.TaskID),
 	})
 
-	if strings.HasPrefix(task.Action.Command, "registry resolve") {
-		logger.Debug("Failed Execution: 'registry resolve' not allowed in action")
-		return errors.New("'registry resolve' not allowed in action")
+	// Temporarily replace "registry " invocations with "regctl " ones
+	// to allow existing manifests to work with renamed tools.
+	if strings.HasPrefix(task.Action.Command, "registry ") {
+		task.Action.Command = "regctl " + strings.TrimPrefix(task.Action.Command, "registry ")
```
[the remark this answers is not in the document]
timburks: Agreed, will do.

timburks:
Closing because we don't want to collide with https://github.com/regclient/regclient/blob/main/docs/regctl.md

PR #573, opened 2022-05-19
Use seeder package for `controller` and `scoring` tests

shrutiparabgoogle:


cmd/registry/controller/controller_test.go
```diff
@@ -225,37 +250,39 @@ func TestArtifacts(t *testing.T) {
...
-				createVersion(ctx, client, t, "projects/controller-test/locations/global/apis/test-api-2", "1.0.1")
-				createSpec(ctx, client, t, "projects/controller-test/locations/global/apis/test-api-2/versions/1.0.1", "openapi.yaml", gzipOpenAPIv3)
-				// Version 1.1.0
-				createVersion(ctx, client, t, "projects/controller-test/locations/global/apis/test-api-2", "1.1.0")
-				createSpec(ctx, client, t, "projects/controller-test/locations/global/apis/test-api-2/versions/1.1.0", "openapi.yaml", gzipOpenAPIv3)
+			seed: []seeder.RegistryResource{
+				// test api 1
+				&rpc.ApiSpec{
+					Name:     "projects/controller-test/locations/global/apis/test-api-1/versions/1.0.0/specs/openapi.yaml",
+					MimeType: gzipOpenAPIv3,
```
seaneganx, line 263: Do we need all these MIME types specified, since we're creating specs that have empty contents? Aside from having a test that covers the filtering behavior I don't expect this field to be very useful in most tests.

seaneganx (approved):
Looks great! My only comment isn't necessary to address in this PR. Just a general question about which information is really necessary for each test.

cmd/registry/controller/controller_test.go
shrutiparabgoogle: Thanks for pointing this out. I guess the MimeType usage got copied over as I used the same template from the initial controller tests. Removed the unnecessary usage.

PR #577, opened 2022-05-19
Simplify demo and walkthrough scripts, replace references to `apg` with `registry rpc` calls

timburks:
This fixes the demos and walkthrough scripts to use `registry rpc` instead of `apg`. Demo scripts are reduced to API uploading and information management (`compute` and related calls are removed).


demos/setup-project.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
giteshk, line 1: Can we callout this file in the README.md file for the demos directory?

timburks: Yes, will do

giteshk (approved):

timburks:
@seaneganx FYI

PR #580, opened 2022-05-20
Tweak setup-project.sh to minimize confusion.

timburks:
This recently-added demo script uses `registry rpc admin` subcommands to delete a project (if it exists) and create a new one.

Errors in the delete command will occur if the project doesn't already exist, so to reduce confusion we suppress those by redirecting stdout/stderr to /dev/null. Errors other than NOT_FOUND will likely be reported by the subsequent create command, which gets a `--json` flag to make its output more readable.

### DIFF
--- a/demos/setup-project.sh
+++ b/demos/setup-project.sh
@@ -22,8 +22,9 @@ PROJECT=sample
 # Self-hosted (open source) installations require project creation.
 
 # Delete and re-create the project to get a fresh start.
-registry rpc admin delete-project --name projects/$PROJECT --force
+registry rpc admin delete-project --name projects/$PROJECT --force >& /dev/null
 registry rpc admin create-project --project_id $PROJECT \
 	--project.display_name $PROJECT \
-	--project.description "A registry project"
+	--project.description "A registry project" \
+	--json
 


timburks:
@giteshk @seaneganx I'll submit on first approval, but feel free to comment post-submit if you have any concerns.

demos/setup-project.sh
```diff
@@ -22,8 +22,9 @@ PROJECT=sample
 # Self-hosted (open source) installations require project creation.
 
 # Delete and re-create the project to get a fresh start.
-registry rpc admin delete-project --name projects/$PROJECT --force
+registry rpc admin delete-project --name projects/$PROJECT --force >& /dev/null
```
giteshk, line 25: could we use  `set -e` at the beginning of the script to allow the script to continue on error.

And use
registry rpc admin delete-project --name projects/$PROJECT --force || true

seaneganx: Alternatively, are we avoiding `registry delete` for some reason? That might be an easier option.

timburks: That won't suppress the error message that's printed by the `delete-project` handler -- I think we would have to modify the generated code to do that (thus the redirect instead).

timburks: This also alerts us that the generated handler doesn't return an error for grpc errors. It seems that it succeeds because the grpc call succeeds, regardless of response code.

timburks: @seaneganx It's not really necessary to use `registry delete` and since there's no corresponding `registry create`, I think the symmetrical commands are easier to follow.

timburks: @seaneganx Also, `registry delete` currently does not delete projects.
```
% registry delete projects/sample
 FATAL[0000] Failed to match or handle command error=unsupported resource name: see the 'apg registry delete-' subcommands for alternatives uid=31e7858b
```

seaneganx (approved):

PR #581, opened 2022-05-20
Fix error message that referred to the apg tool.

timburks:


timburks:
Meant to fix this in #579 but apparently didn't add the file to my commit.

seaneganx (approved):

PR #582, opened 2022-05-20
Recognize accept-encoding headers and return gzip-encoded spec contents.

### LINKED ISSUE #200: Registry API: Improve handling of compressed API specs
Because API specs can occasionally be large enough to exceed gRPC message size limits but also compress extremely well, we encourage users to compress specs before uploading them. But gzip encoding can include timestamps and other variations that make compressions non-unique, so the **SHA hashes that we compute for specs should be based on uncompressed specs** - allowing them to be unique, so that we can use them to detect meaningful changes to specs. **Spec content size should probably also be the uncompressed spec size**. 

Currently, for user convenience, the /contents singleton returns uncompressed specs, but this breaks downloads for specs greater than the gRPC maximum message size (4MB). To address this, **the GetSpecContents method should provide some mechanism for user control of the compression of returned values** - having /contents return the uncompressed specs makes it easier to import specs into third-party tools, so this should at least be available as an option, if not the default.

timburks:
With this change, if clients send "accept-encoding" headers that include "gzip", then gzipped specs will be returned *if the specs were uploaded and stored gzipped and have +gzip in their mime types*. "gzip" does not cause specs that were stored uncompressed to be returned compressed, because it might not be appropriate to compress some types (e.g. "+zip") and we don't plan to maintain an internal list of such uncompressible mime types.

This change also modifies the `registry` tool to send `accept-encoding=[gzip]` from the `registry get --contents` command handler.

This would resolve #200


server/registry/actions_specs_test.go
```diff
@@ -477,6 +480,121 @@ func TestGetApiSpecContents(t *testing.T) {
...
+		return nil, err
+	}
+	if err := zw.Close(); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func TestGetApiSpecContentsData(t *testing.T) {
+	specContentsCompressed, _ := gZippedBytes(specContents)
```
seaneganx, line 497: Let's check this error.

server/registry/actions_specs_test.go
```diff
@@ -477,6 +480,121 @@ func TestGetApiSpecContents(t *testing.T) {
 	}
 }
 
+func gZippedBytes(input []byte) ([]byte, error) {
+	var buf bytes.Buffer
+	zw, _ := gzip.NewWriterLevel(&buf, gzip.BestCompression)
```
seaneganx, line 485: This should never return an error, but I think it's worth checking and returning anyway.

seaneganx (approved):
LGTM but I think it's worth adding some guards for our assumptions about the test helper always working without error.

server/registry/actions_specs_test.go
timburks: ok, done

server/registry/actions_specs_test.go
timburks: agreed, done

PR #583, opened 2022-05-23
Dry run support in compute command

shrutiparabgoogle:
I started this PR with adding `dry-run` support for score commands. 

However, it would be better to add that support across compute commands since each compute sub-command is uploading some artifact to the registry, users should have the option of not uploading the artifact.

### DIFF
--- a/cmd/registry/cmd/compute/complexity.go
+++ b/cmd/registry/cmd/compute/complexity.go
@@ -43,6 +43,10 @@ func complexityCommand() *cobra.Command {
 			if err != nil {
 				log.FromContext(ctx).WithError(err).Fatal("Failed to get filter from flags")
 			}
+			dryRun, err := cmd.Flags().GetBool("dry-run")
+			if err != nil {
+				log.FromContext(ctx).WithError(err).Fatal("Failed to get dry-run from flags")
+			}
 
 			client, err := connection.NewClient(ctx)
 			if err != nil {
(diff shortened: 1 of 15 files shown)


cmd/registry/cmd/compute/compute.go
```diff
@@ -34,5 +34,6 @@ func Command() *cobra.Command {
 	cmd.AddCommand(vocabularyCommand())
 
 	cmd.PersistentFlags().String("filter", "", "Filter selected resources")
+	cmd.PersistentFlags().Bool("dry-run", false, "if set, computation results will only be printed and not uploaded to registry")
```
timburks, line 37: could we have a minor rewording? "if set, computation results will only be printed and not stored in the registry"

shrutiparabgoogle: Done.

cmd/registry/cmd/compute/complexity.go
```diff
@@ -122,16 +128,22 @@ func (task *computeComplexityTask) Run(ctx context.Context) error {
...
-		MimeType: core.MimeTypeForMessageType("gnostic.metrics.Complexity"),
-		Contents: messageData,
-	}
-	err = core.SetArtifact(ctx, task.client, artifact)
-	if err != nil {
-		return err
+
+	if task.dryRun {
+		core.PrintMessage(complexity)
+	} else {
```
seaneganx, line 134: Nit: Prefer stopping the function early over indenting the non-dry-run code one level deeper.

```
if task.dryRun {
	core.PrintMessage(complexity)
	return nil
}

// non-dry-run code
```

seaneganx (approved):

Issue #585, opened 2022-05-24
Registry Scores: Add support for `sum` aggregate function


shrutiparabgoogle:
**Proposal**
It is possible to add extension functions to CEL which behave as built in CEL functions. The proposal in this issue is to add support for a sum function `sum(list)`. The sum function can be initially supported for only lists and then be extended to others types as maps once we find a justified use-case.

**Why do we need this**
In order to create examples for scores, from conformance reports, we need a single number derived from conformance reports which we can assign to a score. Currently, the conformance report stores information in nested lists, we can get a list of the number of errors under each guideline, something like `[2,3,4]`. In order to sum up these errors into a single number, we need a `sum(list)` functionality. 

**Future scope**
This is a good function to have given that we can later extend it to other types as well like maps (if required).
Similar to the `sum` function, `avg` is also a good candidate.

PR #588, opened 2022-06-01
Extend CEL library with sum functions

shrutiparabgoogle:
Add support for `sum` functions in CEL
This is required because the conformance report stores errors and warning in lists.
While it is possible to derive errors for each guideLineReport in the form of lists, example:
```
[
   {
       guideLine: "sample guideline 1",
       errors: []errors,
   },
  {
       guideLine: "sample guideline 2",
       errors: []errors,
   }
]
```

We need an aggregate function which can sum all the error counts together to give us a final number.
See https://github.com/apigee/registry/issues/585 for more details

Future TODO:
Support `sum` for `double` lists.


cmd/registry/scoring/extensions/extensions.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/google/cel-go/checker/decls"
+	"github.com/google/cel-go/interpreter/functions"
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+func Extensions() cel.EnvOption {
+	return cel.Lib(ExtensionLib{})
+}
+
+type ExtensionLib struct{}
```
seaneganx, line 14: Does this need to be exported?

seaneganx (approved):

cmd/registry/scoring/extensions/extensions.go
shrutiparabgoogle: Not required removed.

PR #595, opened 2022-06-09
Add support to upload ApiSpecExtensionList

shrutiparabgoogle:


timburks (commented):
This all LGTM, but I don't think it includes support for `registry export yaml <artifact>` and `registry apply -f <artifact.yaml> `. We can address that in a followup.

giteshk (approved):

giteshk:
@shrutiparabgoogle Thank you for adding support for this.

Issue #594, opened 2022-06-09
Registry tool: Fix `registry apply` for artifacts


shrutiparabgoogle:
In a recent instance of experimentation, I realized that the `registry apply` tool is not setting some of the artifacts correctly. When we upload a styleguide using `registry apply`. The styleguide ID is not set in the artifact's contents proto. This causes issues while computing conformance reports.

Fix this bug and also verify that this is working as expected for other artifacts like `manifest` and `score(card) definitions`.

PR #597, opened 2022-06-10
Fix for missing Id in artifacts through `apply` command

shrutiparabgoogle:
Populate `Id` and `Kind` fields in artifact contents while reading from serialized yamls

Fixes https://github.com/apigee/registry/issues/594

### DIFF
--- a/cmd/registry/patch/artifact.go
+++ b/cmd/registry/patch/artifact.go
@@ -212,7 +212,7 @@ func applyArtifactPatch(ctx context.Context, client connection.Client, content *
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


cmd/registry/patch/artifact.go
```diff
@@ -206,6 +212,24 @@ func applyArtifactPatch(ctx context.Context, client connection.Client, content *
 	return err
 }
 
+// populateIdAndKind inserts the "id" and "kind" in the supplied json bytes
```
timburks, line 215: Nit: I think there's a missing word "fields" between "kind" and "in". Also, comments that are sentences should end with periods.

timburks (approved):

Issue #603, opened 2022-06-13
Evaluate Go Report Card


theganyo:
See: https://goreportcard.com/report/github.com/apigee/registry

Create appropriate issues for more investigation or work as needed.

theganyo:
The `"availaible" is a misspelling of "available"` issue is fixed in this PR: https://github.com/googleapis/gapic-generator-go/pull/1031 and client code generation will need to be rerun once merged and released.

theganyo:
All reported issues aside from gocyclo have been addressed. Gocyclo refactoring can be done on an ad hoc basis.

PR #605, opened 2022-06-13
fix formatting and misspellings

theganyo:
partially address #603

### DIFF
--- a/cmd/registry/diff/differ.go
+++ b/cmd/registry/diff/differ.go
@@ -24,7 +24,7 @@ import (
 	"github.com/tufin/oasdiff/diff"
 )
 
-// change repersents one change in the diff.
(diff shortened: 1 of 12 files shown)


timburks (approved):
The gapic files are generated -- does this change mean we'll want to run `gofmt` or `goimports` on them whenever they are regenerated? I notice that the generated files in `rpc` also change when I run `goimports` on them. Did you explicitly exclude those from the report card?

theganyo:
Oops. You're right, we should avoid the generated files here. I'll take a look.

theganyo:
There's no way to configure go report to ignore these files. Yes, we'd have to run `gofmt -s` after generating the files to clear this issue.

timburks (approved):

PR #607, opened 2022-06-13
Implement the behavior of "refresh" field in manifest definition

shrutiparabgoogle:


cmd/registry/controller/controller_test.go
```diff
@@ -959,3 +961,161 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
...
+			client := seeder.Client{
+				RegistryClient: registryClient,
+				AdminClient:    adminClient,
+			}
+
+			if err := seeder.SeedRegistry(ctx, client, test.seed...); err != nil {
+				t.Fatalf("Setup: failed to seed registry: %s", err)
+			}
+
+			time.Sleep(test.wait * time.Second)
```
seaneganx, line 1097: I think we should consider refactoring the controller so we can easily inject a clock for testing. The quick and simple way to do this would be to add a clock argument to ProcessManifest, but a struct with a field for the "Now" function would be more typical. 

These tests would seed the registry and call the controller with a clock that's artificially pushed ahead (or kept behind) the refresh moment. I don't see a test to make sure the controller doesn't generate actions when it hasn't waited long enough with a non-zero refresh delay, but in that case an injected clock will eliminate assumptions about how quickly the code will execute.

cmd/registry/controller/controller_test.go
```diff
@@ -959,3 +961,161 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
...
+				},
+				&rpc.Artifact{
+					Name: "projects/controller-test/locations/global/apis/petstore/versions/1.0.1/specs/openapi.yaml/artifacts/score-receipt",
+				},
+				&rpc.Artifact{
+					Name: "projects/controller-test/locations/global/apis/petstore/versions/1.1.0/specs/openapi.yaml/artifacts/score-receipt",
+				},
+			},
+			want: nil,
+			wait: 0,
```
seaneganx, line 1018: Nit: These zero waits look like they can be removed, since it should default to zero anyway.

cmd/registry/cmd/apply/testdata/sample/artifacts/manifest.yaml
```diff
@@ -13,6 +13,7 @@ data:
         - pattern: $resource.spec
           filter: mime_type.contains('openapi')
       action: registry compute lint $resource.spec --linter spectral
+      refresh: null
```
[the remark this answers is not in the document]
shrutiparabgoogle: This is needed because [apply_test.go](https://github.com/apigee/registry/blob/main/cmd/registry/cmd/apply/apply_test.go#L104) uploads an artifact from this yaml and then exports it to check if the exported artifact is the same. 

While exporting, the export artifact logic is using this [option](https://github.com/apigee/registry/blob/main/cmd/registry/patch/artifact.go#L123) which exports `refresh:null` fields in the yaml. If the exported yaml doesn't match the one defined in the file, the test fails. Hence, I added these to make sure `apply_test.go` is passing.

@timburks do you think there can be a better workaround to handle this test?

cmd/registry/cmd/upload/artifact_test.go
```diff
@@ -221,7 +221,7 @@ func TestScoreDefinitionArtifactUpload(t *testing.T) {
 						Artifact: &rpc.ResourcePattern{
 							Pattern: "$resource.spec/artifacts/conformance-apihub-styleguide",
 						},
-						ScoreExpression: "sum(guidelineReportGroups[2].guidelineReports.map(r, has(r.ruleReportGroups[1].ruleReports) ? size(r.ruleReportGroups[1].ruleReports) : 0))",
+						ScoreExpression: "has(guidelineReportGroups[2].guidelineReports) ? sum(guidelineReportGroups[2].guidelineReports.map(r, has(r.ruleReportGroups[1].ruleReports) ? size(r.ruleReportGroups[1].ruleReports) : 0)) : 0",
```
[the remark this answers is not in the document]
shrutiparabgoogle: Nope, there were no test failures before, I updated this because of an update in score-definition yaml: `cmd/registry/cmd/upload/testdata/score-definition.yaml`.

The reason for the definition update is, when I was generating scores for googleapis using the previous definition `sum(guidelineReportGroups[2].guidelineReports.map(r, has(r.ruleReportGroups[1].ruleReports) ? size(r.ruleReportGroups[1].ruleReports) : 0))` some of those calculations were failing with `guidelineReports not found`, hence I modified the CEL expression to this: `has(guidelineReportGroups[2].guidelineReports) ? sum(guidelineReportGroups[2].guidelineReports.map(r, has(r.ruleReportGroups[1].ruleReports) ? size(r.ruleReportGroups[1].ruleReports) : 0)) : 0`

This change has nothing to do with the codebase or the tests. It's just an update to the CEL expression to make sure it handles all edge cases while calculating scores. This will be formally defined in an end to end example for scores.

cmd/registry/controller/controller_test.go
shrutiparabgoogle: Yes, I agree. I was uncomfortable adding a sleep in the tests, but couldn't figure out a workaround to test this otherwise. 

`I don't see a test to make sure the controller doesn't generate actions when it hasn't waited long enough with a non-zero refresh delay` -> The test case "existing valid artifacts" on line 1053 of this file captures that case. Where there is no wait in the controller and the artifacts are valid because their refresh delay of 2 seconds is not finished yet. I can update the description if it is not clear.

I can create a tracking bug for the controller refactoring.

cmd/registry/controller/controller_test.go
shrutiparabgoogle: Done.

cmd/registry/cmd/apply/testdata/sample/artifacts/manifest.yaml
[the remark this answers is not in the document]
timburks: Setting "EmitUnpopulated" addressed a problem where zero-valued integers (display order in lifecycle and taxonomies) weren't being written. I don't think we can get different behavior for different fields without adding per-resource custom code (which we've minimized). I agree that explicitly including this here is verbose and unnecessary for reimport, and it does raise the question "what is this?" but that seems fine (maybe beneficial even) in an example.

tl;dr I'm ok with keeping this default field in this sample manifest.

timburks (approved):

cmd/registry/cmd/apply/testdata/sample/artifacts/manifest.yaml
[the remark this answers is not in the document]
seaneganx: Sounds like EmitUnpopulated is needed. The test verification might have room for improvement though. I'm not sure we need to be failing a test if the export YAML isn't identical to the input, as long as we make sure they're semantically equivalent.

I'm fine with this until (if ever) we change the test verification.

seaneganx (approved):

cmd/registry/controller/controller_test.go
shrutiparabgoogle: https://github.com/apigee/registry/issues/613

Issue #189, opened 2021-06-16
Testing: Create library for producing isolated test servers


seaneganx:
Tests should be able to start and automatically stop a server that can be accessed using the registry client that the code under test depends on. If these servers can point to separate databases, then tests will be able to execute in isolation.

Currently our tests run against a shared locally running server. Each test has to be careful to avoid interfering with other tests or depending on state that can be affected by other tests. Developers need to make sure a server is running before executing their tests, and also need to be aware of when they last built and ran the server to ensure their tests are covering the expected code.

timburks:
Just noting: an effectively-required best practice is for every test to run against a unique project.

seaneganx:
Right. That helps avoid collisions between packages. Each package still has potentially dozens of tests that need to isolate from each other. 

Even with isolated projects, I've encountered issues listing `projects/-/apis/-/versions/-/specs/-` to verify the specs created by my tests, and didn't notice the isolation bug until CI failed. Tests can be written in isolation but need to work alongside the entire suite, which makes bugs like that possible.

timburks:
I think tests should also be runnable against a hosted service; please keep this in mind in any restructuring.

seaneganx:
All tests or some tests? I agree we should have integration testing to verify our system works in both local and hosted environments. Having a smaller suite of integration tests for that purpose with a larger set that runs in a faster, deterministic, and hermetic environment seems like the best approach IMO.

I think that hybrid approach works best because tests need to be written according to the "lowest common denominator" of environments they will run against. Every test that runs against a shared hosted service needs to consider how it will interact with the other N-1 tests it may be running with.

Subjecting only a small subset of our tests (e.g. the demo tests) to the restrictions of non-hermetic testing gives us a better experience developing and maintaining the majority of our tests that are just meant to verify functionality.

What do you think about this approach?

timburks:
If it's possible to write tests that run against all environments, I think that is useful and more future-proof than overcommitting a set of tests to a particular implementation. The current test architecture can be run against hosted versions and instances that use other storage backends (Postgres, possibly other gorm DBs). I don't like sacrificing that to achieve something that we could achieve with a test discipline like requiring each test to run in a unique project.

PR #610, opened 2022-06-15
Internal-test-server

theganyo:
Enable use of an internal test server for unit testing.

### DIFF
--- a/cmd/registry/cmd/upload/upload_test.go
+++ b/cmd/registry/cmd/upload/upload_test.go
@@ -1,0 +1,29 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package upload
+
+import (
+	"testing"
+
+	"github.com/apigee/registry/server/registry"
+	"github.com/apigee/registry/server/registry/test/grpctest"
+)
+
+// TestMain will set up a local RegistryServer and grpc.Server for all
+// tests in this package if APG_REGISTRY_ADDRESS env var is not set
+// for the client.
+func TestMain(m *testing.M) {
+	grpctest.TestMain(m, registry.Config{})
+}
--- a/server/registry/test/grpctest/server.go
+++ b/server/registry/test/grpctest/server.go
@@ -1,0 +1,134 @@
+package grpctest
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net"
+	"os"
+	"testing"
+
+	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/registry"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+// NewIfNoAddress will create a RegistryServer served by a
+// basic grpc.Server if env var APG_REGISTRY_ADDRESS is
+// not set, otherwise will simply return a dummy Server
+// assuming the tests will connect to a the remote address.
+func NewIfNoAddress(rc registry.Config) (*Server, error) {
+	addr := os.Getenv("APG_REGISTRY_ADDRESS")
+	if addr != "" {
+		return &Server{}, nil
+	}
+	return NewServer(registry.Config{})
+}
+
+// TestMain can delegate here in packages that wish to
+// use TestMain for tests using a RegistryServer. This
+// uses NewIfNoAddress() semantics.
+func TestMain(m *testing.M, rc registry.Config) {
+	l, err := NewIfNoAddress(rc)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer l.Close()
+	os.Exit(m.Run())
+}
+
+// NewServer creates a RegistryServer served by a basic grpc.Server.
+// If rc.Database and rc.DBConfig are blank, a RegistryServer using
+// sqlite3 on a tmpDir is automatically created.
+// APG_REGISTRY_ADDRESS, APG_REGISTRY_AUDIENCES, and APG_REGISTRY_INSECURE
+// env vars are set for the client to connect to the created grpc service.
+// Call Close() when done to close server and clean up tmpDir as needed.
+// Example:
+// func TestXXX(t *testing.T) {
+// 	 l, err := grpctest.NewServer(registry.Config{})
+// 	 if err != nil {
+// 		t.Fatal(err)
+// 	 }
+// 	 defer l.Close()
+//   ... run test here ...
+// }
+func NewServer(rc registry.Config) (*Server, error) {
+	gl := &Server{}
+	var err error
+	if rc.Database == "" {
+		rc.Database = "sqlite3"
+	}
+	if rc.Database == "sqlite3" && rc.DBConfig == "" {
+		f, err := ioutil.TempFile("", "registry.db.*")
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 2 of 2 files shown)


theganyo:
Opened this as a draft example of how we could update all of the unit tests to run without required an outside server to be started.

cmd/registry/cmd/upload/upload_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+
+	"github.com/apigee/registry/server/registry"
+	"github.com/apigee/registry/server/registry/test/grpctest"
+)
+
+// TestMain will set up a local RegistryServer and grpc.Server for all
+// tests in this package if APG_REGISTRY_ADDRESS env var is not set
+// for the client.
+func TestMain(m *testing.M) {
```
timburks, line 27: I'm guessing that this would need to be done for every package under `cmd`, right?

timburks (commented):
I don't know, this seems like a lot of magic to autostart a server that we might instead do in `make test` by just running `registry-server` right before running [`go test ./...`](https://github.com/apigee/registry/blob/257aefd033e819d318499092b9ef9d12148542ef/Makefile#L27) -- that call will fail if a server is already running (and we could catch and ignore the error there). This wouldn't require any code changes to the tests.

theganyo:
Other than common expectation and ease of use, there are additional advantages to unit tests being self-contained. This way, each test is isolated from the others, so no potential side-effects, and testing can also be safely parallelized.

cmd/registry/cmd/upload/upload_test.go
timburks: That seems like a high line-count impact. Could this initialization be made implicit or moved up a level?

server/registry/test/grpctest/server.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		gl.Close()
+		return nil, err
+	}
+
+	gl.Server = grpc.NewServer()
+	reflection.Register(gl.Server)
+	rpc.RegisterRegistryServer(gl.Server, rs)
+	rpc.RegisterAdminServer(gl.Server, rs)
+
```
timburks, line 86: This seems like a lot of code duplication.

timburks:
@theganyo When we test the `registry-server` with the PostgreSQL backend, we just set up a PostgreSQL instance and run against that. I'm ok with that compromise (since we don't have a mock postgres to instantiate in our test) and I prefer the testing power of flexibility in this case as well, but I also am prepared to be outvoted (https://github.com/apigee/registry/issues/189) :).

server/registry/test/grpctest/server.go
theganyo: Hmm. Do you mean with this? https://github.com/apigee/registry/blob/main/cmd/registry-server/main.go#L144. We could do a little refactoring once we decide this is a good direction.

cmd/registry/cmd/upload/upload_test.go
theganyo: I believe [TestMain()](https://pkg.go.dev/testing#hdr-Main) is the best / only official Go test mechanism to do this and it's package-specific.

timburks:
https://github.com/apigee/registry/issues/611 describes another testing-related concern that's possibly competing with this one.

theganyo:
> #611 describes another testing-related concern that's possibly competing with this one.

I'm not sure that competes, I believe unit tests should be a completely different concern than conformance testing.

server/registry/test/grpctest/server.go
timburks: That's right. That's not in `server/registry` because I wanted to be able to mix in other servers (like a search server) from other future locations (at one time there was a search subdirectory that is now in registry-experimental). But a helper somewhere that does common work seems appropriate if you do this.

cmd/registry/cmd/upload/upload_test.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
```
seaneganx, line 1: 2022

server/registry/test/grpctest/server.go
```diff
@@ -0,0 +1,134 @@
+package grpctest
```
seaneganx, line 1: This file should probably have a license.

server/registry/test/grpctest/server.go
```diff
@@ -0,0 +1,134 @@
...
+// APG_REGISTRY_ADDRESS, APG_REGISTRY_AUDIENCES, and APG_REGISTRY_INSECURE
+// env vars are set for the client to connect to the created grpc service.
+// Call Close() when done to close server and clean up tmpDir as needed.
+// Example:
+// func TestXXX(t *testing.T) {
+// 	 l, err := grpctest.NewServer(registry.Config{})
+// 	 if err != nil {
+// 		t.Fatal(err)
+// 	 }
+// 	 defer l.Close()
```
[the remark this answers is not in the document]
theganyo: testing.T is unavailable here. This defer won't happen until the test is finished.

server/registry/test/grpctest/server.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/apigee/registry/server/registry"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+// NewIfNoAddress will create a RegistryServer served by a
+// basic grpc.Server if env var APG_REGISTRY_ADDRESS is
+// not set, otherwise will simply return a dummy Server
+// assuming the tests will connect to a the remote address.
+func NewIfNoAddress(rc registry.Config) (*Server, error) {
```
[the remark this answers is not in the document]
theganyo: This function is exported for test functions that want to set up an instance per iteration.

theganyo:
This change will isolate tests by giving them each their own DB. However, it's worth noting that because this preserves the existing client semantics that rely on global env vars for connection information, the tests cannot currently be parallelized whilst maintaining this isolation.

theganyo:
This is ready for final review.

connection/grpctest/server.go
```diff
@@ -0,0 +1,153 @@
...
+// 	 defer l.Close()
+//   ... run test here ...
+// }
+func NewIfNoAddress(rc registry.Config) (*Server, error) {
+	addr := os.Getenv("APG_REGISTRY_ADDRESS")
+	if addr != "" {
+		log.Printf("Client will use remote database at: %s", addr)
+		return nil, nil
+	}
+	log.Println("Client will use embedded SQLite3 database")
```
timburks, line 48: Is there a way (or is there no way) to configure the automatically-generated server to use a Postgres database?

timburks: I want to be able to detect (and rule-out) database-specific problems that might appear in the cmd tests.

Most database-specific problems should be found by the core server tests (in server/registry), and these can be run locally with postgres, which is great. If we assume that the core tests are complete, then we wouldn't need to run the cmd tests against a postgres database. But to me, that assumption takes an unnecessary risk for the arguable convenience of not having to start a test server.

theganyo: Aside: I'm just going to drop this link here in case we decide to run unit tests with an embedded postgres: https://github.com/fergusstrange/embedded-postgres.

timburks: I just created #624 to evaluate embedded postgres for unit and CI testing.

Summarizing our face-to-face discussion, we agreed to move forward with this PR with a new near-term goal to replace the use of environment variables to configure the registry client with configuration that can be read from a file. This file can be managed by the registry tool (allowing commands like `registry auth login` and `registry config set location`) and could point to arbitrary servers with arbitrary database backends. We think this is useful for allowing `registry config set project` and `registry config set location`, which could allow shorter relative resource names to be provided to other `registry` subcommands like `registry list`.

Most relevant to this PR, by removing the dependence on global values (environment variables) to configure tests, that reduces the uncertainty about what is happening that is expressed in all the comments like this one:
```
// TestMain will set up a local RegistryServer and grpc.Server for all
// tests in this package if APG_REGISTRY_ADDRESS env var is not set
// for the client.
```

tl;dr for now we're moving to a slightly more complex state that fixes a short-term problem and will make future improvements (issue TBD) to improve overall simplicity and clarity.

timburks (approved):

PR #615, opened 2022-06-17
Add new low-level db.create() method that properly returns "AlreadyExists"

timburks:
This adds a new `create()` method to the `storage` package that recognizes error codes that are returned when attempts are made to create entries that already exist, i.e. they have the same primary key value as an existing entry. Unfortunately this requires checking database-specific response codes, as gorm does not currently include this in its list of [cross-database error codes](https://github.com/go-gorm/gorm/blob/master/errors.go).

With this change, we can rewrite the create actions to remove the pre-save "get" checks and we can rewrite update actions to speculatively create entities (when "allow missing" is set) and continue with updates when these speculative creates fail because entities already exist. This removes multiple race conditions and makes it possible to remove the mutexes in the update methods. _Update: additional commits in this PR make these changes for projects, apis, and versions._


server/registry/actions_apis.go
```diff
@@ -202,29 +185,29 @@ func (s *RegistryServer) UpdateApi(ctx context.Context, req *rpc.UpdateApiReques
...
+		}
+	}
+
+	db, err := s.getStorageClient(ctx)
+	if err != nil {
+		return nil, status.Error(codes.Unavailable, err.Error())
 	}
+	defer db.Close()
 
+	// TODO: the Get/Update/Save calls below should be a transaction.
```
theganyo, line 203: Is there an issue filed to address this?

server/registry/internal/storage/logger.go
```diff
@@ -58,7 +58,7 @@ func (l gormLogger) Trace(ctx context.Context, begin time.Time, fc func() (sql s
 		"duration": time.Since(begin),
 	})
 
-	if err != nil && err != gorm.ErrRecordNotFound {
+	if err != nil && err != gorm.ErrRecordNotFound && !alreadyExists(err) {
```
theganyo, line 61: A larger question than this PR, but should we consider using [errors.Is](https://pkg.go.dev/errors#Is)?

seaneganx (approved):

server/registry/actions_projects_test.go
```diff
@@ -203,7 +203,7 @@ func TestCreateProjectDuplicates(t *testing.T) {
 				ProjectId: "My-Project",
 				Project:   &rpc.Project{},
 			},
-			want: codes.AlreadyExists,
+			want: codes.InvalidArgument, // the upper-case characters are invalid; this is discovered before the save
```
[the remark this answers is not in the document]
timburks: Done - the tables are removed but I kept the test structs for clarity.

server/registry/storage_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"testing"
+
+	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/registry/internal/storage/models"
+	"github.com/apigee/registry/server/registry/names"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func TestDuplicateCreation(t *testing.T) {
```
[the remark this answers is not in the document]
timburks: I added this to assure myself that `AlreadyExists` was being returned when I expected - so in one sense, it's development scaffolding, but in another I think it's useful to verify an underlying assumption. I see this as a test that particularly defends against changes in the gorm dependency. If the code returned by gorm changes, it would be easier to detect and diagnose here than by stepping through the API method tests. 

I had hoped to put this inside the `storage` package, but the database configuration that it needed was in the `registry` (server) package. We might rearrange that in the future if we decide to add specific storage package tests.

server/registry/internal/storage/logger.go
timburks: Sure, I think that would be fine. In this situation the `alreadyExists` function requires looking at status codes attached to the errors so I think it might not apply.

server/registry/actions_apis.go
timburks: #623

server/registry/storage_test.go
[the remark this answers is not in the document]
timburks: Mentioned [here](https://github.com/apigee/registry/pull/640/files#r915198299)

PR #620, opened 2022-06-22
Registry server: remove client mutex

### LINKED ISSUE #619: Registry server: remove client mutex
The [disableMutex](https://github.com/theganyo/registry/blob/main/server/registry/internal/storage/client.go#L51) variable is not thread-safe as it is globally referenced but not protected by a mutex.

That said, rather than fix this issue, I suggest we remove the client mutex entirely. Evidence this is safe:

1. SQLite3 is [thread-safe by default](https://www.sqlite.org/threadsafe.html).
2. The SQLite3 library we use [ensures the DB was compiled for thread-safety](https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L1070).
3. All tests pass still pass after removing the mutex.
4. Running `registry upload bulk discovery --jobs 100` is successful after removing the mutex.

theganyo:
Fixes #619

### DIFF
--- a/server/registry/internal/storage/client.go
+++ b/server/registry/internal/storage/client.go
@@ -17,7 +17,6 @@ package storage
 import (
 	"context"
 	"fmt"
-	"sync"
 
 	_ "github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/postgres"
 	"github.com/apigee/registry/server/registry/internal/storage/models"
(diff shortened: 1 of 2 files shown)


server/registry/internal/storage/client.go
```diff
@@ -77,13 +60,11 @@ func NewClient(ctx context.Context, driver, dsn string) (*Client, error) {
...
 			c := &Client{db: db}
 			c.close()
-			unlock()
 			return nil, err
 		}
-		unlock()
 		// empirically, it does not seem safe to disable the mutex for sqlite3,
 		// which might make sense since sqlite database access is in-process.
-		disableMutex = false
+		// disableMutex = false
```
timburks, line 67: Let's remove these comments

PR #630, opened 2022-06-24
WIP: Exporting Apigee deployments into a API Registry compatible format

seaneganx:
Run the program using `go run ./cmd/registry sync organizations/my-org $(gcloud auth print-access-token)`, replacing the Apigee organization with one you have access to via `gcloud apigee organizations list`. Make sure you're logged into `gcloud` so the access token can be found.

An organization with one proxy deployed to an environment accessible through one hostname will produce output similar to the following.

```
{
	"name":  "projects/my-project/locations/global/apis/hello-world/deployments/egansean-experimental-eval-apigee-net",
	"displayName":  "hello-world",
	"endpointUri":  "egansean-experimental-eval.apigee.net",
	"labels":  {
		"apigeex-envgroup":  "organizations/egansean-experimental/envgroups/eval-group",
		"apigeex-environment":  "organizations/egansean-experimental/environments/eval",
		"apigeex-proxy":  "organizations/egansean-experimental/apis/hello-world/revisions/3"
	}
}
```

This code is very experimental, and I don't intend on merging it into this repo at all. My intention is to start developing this in the registry-experimental repo and potentially move it into a more permanent non-experimental location in the future.

Notes:
* This program is unlikely to be part of the registry tool, and unlikely to be called `sync`.
* I don't intend on users providing an Apigee API access token as a program argument. 
* Almost everything in this draft is likely to change.

### DIFF
--- a/cmd/registry/cmd/sync/sync.go
+++ b/cmd/registry/cmd/sync/sync.go
@@ -1,0 +1,241 @@
+// Copyright 2022 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sync
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/apigee/registry/log"
+	"github.com/apigee/registry/rpc"
+	"github.com/spf13/cobra"
+	"google.golang.org/api/apigee/v1"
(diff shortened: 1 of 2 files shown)


cmd/registry/cmd/sync/sync.go
```diff
@@ -0,0 +1,241 @@
...
+				m.hostnames[attachment.Environment] = append(m.hostnames[attachment.Environment], hostname)
+				m.envgroup[hostname] = envgroup
+			}
+		}
+	}
+
+	return m, nil
+}
+
+func bundle(ctx context.Context, revision, token string) ([]byte, error) {
```
seaneganx, line 221: This isn't used anywhere, but I kept it just in case we decide to download proxy bundles for some reason.

timburks:
This worked for me on a test org. I see that what we're getting here is a list of deployments that would be associated with an API in the registry (just restating the obvious). I wonder if there's some metadata that we could attach to the proxies that might identify the "parent" API. `apigeecli` shows metdata attached to proxies, but haven't yet seen how to set it:
```
$ apigeecli apis get -t $(gcloud auth print-access-token) -o timburks-testproject --name hello-world
{
        "metaData": {
                "createdAt": "1655160964752",
                "lastModifiedAt": "1655160964752",
                "subType": "Proxy"
        },
        "name": "hello-world",
        "revision": [
                "1"
        ],
        "apiProxyType": "PROGRAMMABLE"
}
```

seaneganx:
Obsolete: See https://github.com/apigee/registry-experimental/pull/76

Issue #593, opened 2022-06-09
Review conformance to Go standard project layout


timburks:
https://github.com/golang-standards/project-layout

timburks:
We could start by creating (and using) a top-level `pkg` directory. There's one proposed entry in #631. What else might we include? I think the [names package](https://github.com/apigee/registry/tree/main/server/registry/names) is a good candidate and also the top-level `yaml` package that's proposed in #629. Others?

theganyo:
We'll just want to be very careful about package structure, names, and contents as we'd be implicitly declaring that this shouldn't break - at least between major versions.

PR #631, opened 2022-06-25
Minimal initial benchmarks

timburks:
This pares down @giteshk's benchmarks to a single set of CRUD tests on API resources that is hopefully easily extensible to other resources and scenarios. I'm not sure that we need to do this for every resource since the underlying operations don't vary much per resource. This alone seems enough to do some interesting comparisons of various hosting alternatives.


.github/workflows/go.yml
```diff
@@ -82,6 +82,18 @@ jobs:
...
 
+    - name: Verify that benchmarks run
+      env:
+        APG_REGISTRY_ADDRESS: localhost:8080 
+        APG_REGISTRY_AUDIENCES: http://localhost:8080
+        APG_REGISTRY_INSECURE: 1
+        PROJECTID: bench
+        ITERATIONS: 1
+      run: |
+        registry-server &
```
timburks, line 93: This isn't necessary because a server is started on line 65. In fact, it exists with an error because the port is in use by that server, but this doesn't block or affect the outcome of the test. I would take it out but I think @theganyo is likely to remove the server start on line 65 soon.

pkg/wipeout/wipeout.go
```diff
@@ -0,0 +1,201 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
```
timburks, line 1: This is the first file in `pkg`, which is a nod to #593

.github/workflows/go.yml
```diff
@@ -82,6 +82,17 @@ jobs:
...
 
+    - name: Verify that benchmarks run
+      env:
+        APG_REGISTRY_ADDRESS: localhost:8080 
+        APG_REGISTRY_AUDIENCES: http://localhost:8080
+        APG_REGISTRY_INSECURE: 1
+        PROJECTID: bench
+        ITERATIONS: 1
+      run: |
+        registry rpc admin create-project --project_id=$PROJECTID --json
```
timburks, line 93: This works because a server is started above - I believe the one on line 65 is still running. I think @theganyo might remove that line soon, and this will break as a surprise bonus which we can either fix by starting a server here or starting an embedded server using new tricks.

theganyo: It wasn't my intention to remove that as I consider this to be part of the integration test suite.

pkg/wipeout/wipeout.go
theganyo: Perhaps we should soon do at least the `pkg` part of that refactor to avoid divergence.

.github/workflows/go.yml
timburks: @theganyo OK, leaving that in is fine with me (as is taking it out). I'm not that interested in regularly benchmarking against SQLite since all our deployments will use Postgres. I added this to CI only to ensure that checkins don't break the benchmarking code.

pkg/wipeout/wipeout.go
timburks: +1 let's discuss further in #593

timburks:
@theganyo it works for me... I was originally running on my Chromebook but just verified that it works on a MacBook (with a small change to line 21 of the script). How did you run this?

theganyo:
`./tests/benchmark/BENCH.sh` needs +x perm set

theganyo:
I've found the issue. It's no coincidence that the errors are for ID 10: It's the [default page size](https://github.com/apigee/registry/blob/main/cmd/registry/cmd/rpc/generated/list-apis.go#L28) for the ListXXXCmd actions which are being used in wipeout.

timburks:
Well, I thought the problem was with wipeout but it seems instead that the function passed to Run() gets called multiple times without cleaning up after itself. Fixing...

theganyo:
On another note re: schema. The wipeout code seems to make it clear we're not using foreign key constraints in the database? Manually maintaining DB integrity can be quite error prone... Shall I create an issue to address this?

timburks:
@theganyo Agreed - after starting a `wipeout_test.go`, I'd prefer to add that in a separate PR because I expect a lot of discussion about the test structure.

theganyo (approved):

pkg/wipeout/wipeout.go
```diff
@@ -0,0 +1,204 @@
...
+// Set this global variable in calling code to run verbosely.
+var Verbose bool
+
+// This collects one (arbitrary) error that occurs during wipeout.
+// It's a global and gets assigned from concurrent goroutines,
+// but if it fails, it fails.
+var _err error
+
+// Wipeout deletes all resources in a project using the specified number of parallel worker jobs.
+func Wipeout(ctx context.Context, client connection.Client, projectid string, jobs int) error {
```
seaneganx, line 35: projectID

pkg/wipeout/wipeout.go
```diff
@@ -0,0 +1,204 @@
...
+	"context"
+
+	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/log"
+	"github.com/apigee/registry/rpc"
+)
+
+// Set this global variable in calling code to run verbosely.
+var Verbose bool
```
seaneganx, line 27: Have we considered using flags or some other method for this config? I think this is the first use of config by setting a global var.

Also worth considering whether we need this at all, especially in the "minimal" initial benchmarks. We already have multiple logging levels that give control over the verbosity of output logs. If we want wipeout to be silent by default and optionally turned on, maybe we should just log at debug level instead of adding another dial for controlling log output.

pkg/wipeout/wipeout.go
```diff
@@ -0,0 +1,204 @@
...
+
+const (
+	API ResourceKind = iota
+	Version
+	Spec
+	Deployment
+	Artifact
+)
+
+type deleteResourceTask struct {
```
seaneganx, line 101: These seems like a duplicate of `deleteTask` from the `registry delete` command. In #445 I suggested creating some reusable types for common tasks to avoid reinventing the wheel each time we want to do simple operations like creating/updating/deleting resources. Maybe now's the time?

The tasks can also be simpler by making them type-specific, which eliminates the "this should never happen" situation where we encounter an "unknown resource type" in these generic-esque tasks.

timburks:
Done (added pkg/README.md)

tests/benchmark/benchmark_apis_test.go
```diff
@@ -0,0 +1,170 @@
...
+func apiId(i int) string {
+	return fmt.Sprintf("test-%d", i)
+}
+
+func apiName(apiId string) string {
+	return fmt.Sprintf("%s/apis/%s", root(), apiId)
+}
+
+func getApi(b *testing.B, ctx context.Context, client connection.Client, apiId string) error {
+	b.Helper()
```
seaneganx, line 37: Since this function doesn't actually fail/error/log using `b`, I don't think we gain anything from marking it as a helper (and can drop it entirely from the function).

tests/benchmark/benchmark_helpers.go
```diff
@@ -0,0 +1,64 @@
...
+	return fmt.Sprintf("projects/%s/locations/global", projectID)
+}
+
+func setup(b *testing.B) (context.Context, connection.Client) {
+	b.Helper()
+	ctx := context.Background()
+	client, err := connection.NewClient(ctx)
+	if err != nil {
+		b.Logf("Unable to connect to registry server. Is it running?")
+		b.FailNow()
```
seaneganx, line 49: b.Fatalf?

tests/benchmark/benchmark_helpers.go
```diff
@@ -0,0 +1,64 @@
...
+
+func init() {
+	flag.StringVar(&projectID, "project_id", "bench", "registry project ID")
+}
+
+func root() string {
+	return fmt.Sprintf("projects/%s/locations/global", projectID)
+}
+
+func setup(b *testing.B) (context.Context, connection.Client) {
```
seaneganx, line 43: Why do we return a context here? This seems like something we should accept from callers.

tests/benchmark/benchmark_helpers.go
```diff
@@ -0,0 +1,64 @@
...
+		b.FailNow()
+	}
+	if err := wipeout.Wipeout(ctx, client, projectID, jobs); err != nil {
+		b.Logf("Pre-test wipeout failed")
+		b.FailNow()
+	}
+	return ctx, client
+}
+
+func teardown(b *testing.B, ctx context.Context, client connection.Client) {
```
seaneganx, line 58: We use `ctx context.Context, t *testing.T` elsewhere, so maybe we should swap `b` and `ctx` here. In other Google projects we prefer that ordering too.

pkg/wipeout/wipeout.go
timburks: Thanks, I'll use a followup PR to change this to logging at Debug. How would we set that for a specific run of the test? (directly in code?)

pkg/wipeout/wipeout.go
timburks: https://github.com/apigee/registry/pull/639/commits/e67bea23963b3353f0ec1cc3e0f672b8d3f3fc51

pkg/wipeout/wipeout.go
timburks: Done in https://github.com/apigee/registry/pull/639/commits/19f552bb3e9abe726a6151dc3243eb9e7d50ef4a

tests/benchmark/benchmark_helpers.go
timburks: Updated in https://github.com/apigee/registry/pull/639/commits/3c841fc7482bb9babb142060f2fd1b29a5a72e9f

tests/benchmark/benchmark_helpers.go
timburks: Done in https://github.com/apigee/registry/pull/639/commits/8a8cd103e117462e94b1e39645af90099ffd2351

tests/benchmark/benchmark_helpers.go
timburks: This reduced the boilerplate at the start of tests to a single call to setup(). Is there a style guide preference to not return a context from a function?

pkg/wipeout/wipeout.go
timburks: partly addressed in #639

PR #638, opened 2022-06-30
Registry server: reuse PubSub client

### LINKED ISSUE #635: Registry server: reuse PubSub client
Similar to our database connections, we should only create the PubSub client once. (The [canonical example](https://pkg.go.dev/cloud.google.com/go/pubsub#section-readme) for PubSub in the documentation states "create a pubsub.Client to use throughout your application".) Also, we need unit tests.

theganyo:
Also adds the ability to check logging in tests

Closes #635

### DIFF
--- a/server/registry/server.go
+++ b/server/registry/server.go
@@ -16,11 +16,14 @@ package registry
 
 import (
 	"context"
+	"errors"
 	"log"
 	"net"
 
+	"cloud.google.com/go/pubsub"
 	"github.com/apigee/registry/rpc"
 	"github.com/apigee/registry/server/registry/internal/storage"
+	"google.golang.org/api/option"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/reflection"
@@ -29,12 +32,13 @@ import (
 
 // Config configures the registry server.
 type Config struct {
-	Database  string
-	DBConfig  string
-	LogLevel  string
-	LogFormat string
-	Notify    bool
-	ProjectID string
+	Database         string
+	DBConfig         string
+	LogLevel         string
+	LogFormat        string
+	Notify           bool
+	ProjectID        string
+	APIClientOptions []option.ClientOption
 }
 
 // RegistryServer implements a Registry server.
@@ -43,7 +47,8 @@ type RegistryServer struct {
 	dbConfig      string
 	notifyEnabled bool
 	projectID     string
-	client        *storage.Client
+	storageClient *storage.Client
+	pubSubClient  *pubsub.Client
 
 	rpc.UnimplementedRegistryServer
 	rpc.UnimplementedAdminServer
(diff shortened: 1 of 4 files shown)


theganyo:
There's one other change I'd like to make. This line can be problematic in that it will wait for the notification to be sent:

https://github.com/apigee/registry/blob/main/server/registry/notifications.go#L73

Do we really need this? And if we feel we do need to log this, we should probably do that asynchronously.

server/registry/server.go
```diff
@@ -29,12 +32,13 @@ import (
...
-	LogFormat string
-	Notify    bool
-	ProjectID string
+	Database         string
+	DBConfig         string
+	LogLevel         string
+	LogFormat        string
+	Notify           bool
+	ProjectID        string
+	APIClientOptions []option.ClientOption
```
timburks, line 41: What does this look like in YAML? There's a default config file in the `config` directory, could we modify that with something that illustrates this? (it could be commented out)

timburks, line 41: Is this specifically for pubsub? If so, maybe the name should reflect that? As it is it's kind of inscrutable (e.g. "why would I have an API client in the server?" "are APIClientOptions extra options that clients need to pass to the server?").

timburks (commented):
No sigificant concerns, but I'll leave it to @seaneganx to approve the tests (and anything else of concern)

timburks:
@theganyo I'm ok with removing that line that you mentioned - this was originally just copied out of pubsub sample code. I think it makes sense to not wait for confirmation.

server/registry/server.go
theganyo: Not exactly, ClientOptions are set of options for any Google API clients (see: https://pkg.go.dev/google.golang.org/api/option#ClientOption), but yes, PubSub is the only API we use here.

theganyo: At this point, this wouldn't be exposing this as it's only used by the tests to set up a local connection to the mock server. So, if wanted to add any such setting to the main ServerConfig class, we'd just add it there something like this:

```
pubsub:
   endpoint: http://not.the.normal.google.com
// or global
google:
   endpoint: http://not.the.normal.google.com
```

That said, maybe we should punt on all that. There's another way to set this up that just uses an env var, I can just do that instead and remove the APIClientOptions. See: https://pkg.go.dev/cloud.google.com/go/pubsub#hdr-Emulator

theganyo: Done

theganyo:
@timburks I've made all the changes I suggested above.

> @theganyo I'm ok with removing that line that you mentioned - this was originally just copied out of pubsub sample code. I think it makes sense to not wait for confirmation.

Perhaps it is an issue with the mock server, but the test fails if the Get() call is removed - so I added it back in for now.

timburks (approved):
Thanks @theganyo! I think we're fine to merge this - @seaneganx if you see concerns, please follow up but everything here LGTM.

seaneganx (commented):
I'm suspicious about the log recorder and using it to make assertions about specific details of what's being logged by the notify function. It feels to me like we're testing the wrong behavior (did it log vs. did it notify), and doing it in a way that's more brittle (logs must say exactly X) than inspecting how the notify function interacts with a fake pubsub client, like you did in the other test. 

If I'm misunderstanding the goal of the tests, and we really do care about the logging, can you help me understand why?

Issue #611, opened 2022-06-16
Registry API: no API conformance tests; server tests only verify locally-available code.


timburks:
We currently lack a way to verify that a running server properly conforms to the Registry API specification as described by protos and the unit tests in [/server/registry](https://github.com/apigee/registry/tree/main/server/registry). This makes it difficult to verify hosted instances and alternate implementations that include internal mocks of the Registry API. 

How can we avoid redundant test-writing but also address this need?

timburks:
Resolved by #664.

PR #640, opened 2022-07-02
Add a proxy to allow tests written to the generated RegistryServer interface to call remote servers

timburks:
Everything prior to [this commit](https://github.com/apigee/registry/pull/640/commits/4122b98bee2dec3d40c89434eac3f37493eaa101) is  from #639.

This addresses #611 by adding a proxy that makes it possible to run the server tests against any remote server that includes the admin service. This allows testing of arbitrary configurations and builds that include archived containers. The proxy uses the default client configuration (currently `APG_REGISTRY_ADDRESS`, etc), and is used when the `-remote` flag is passed to `go test ./servers/registry`, e.g.
```
go test ./servers/registry -remote
```


pkg/remote/proxy.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/rpc"
+	"google.golang.org/api/iterator"
+	"google.golang.org/genproto/googleapis/api/httpbody"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+// Proxy implements a local proxy for a remote Registry server.
+// This allows tests written to the generated "RegistryServer" and "AdminServer" interfaces to be run against remote servers.
```
theganyo, line 29: Is this just for testing? It almost seems like it could  be general-purpose (aside from the Open() function behavior). If this _is_ just intended for supporting tests, we should probably place it in a package under the "tests" folder or some other similar "test" oriented folder.

pkg/remote/proxy.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (p *Proxy) Open(ctx context.Context) error {
+	var err error
+	p.adminClient, err = connection.NewAdminClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Delete everything in the remote server to be tested.
+	it := p.adminClient.ListProjects(ctx, &rpc.ListProjectsRequest{})
```
theganyo, line 46: Deleting everything in the remote server would be an astonishing side-effect of an Open() function (at least to me). We should make the delete an explicit call or perhaps change the name of the function to indicate this behavior as well as add a comment to that effect. (All public functions should have comments anyway.)

pkg/wipeout/tasks.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"context"
+
+	"github.com/apigee/registry/connection"
+	"github.com/apigee/registry/log"
+	"github.com/apigee/registry/rpc"
+)
+
+// DeleteApiTask deletes a specified API.
+type DeleteApiTask struct {
```
theganyo, line 26: Should these Tasks be just wipeout-related? Or would such Tasks (and processing of Tasks) be an interesting abstraction to make available in general?

timburks: @seaneganx suggested general tasks [in this comment](https://github.com/apigee/registry/pull/631#discussion_r910381339) and a preceeding issue. I'm open to that, which might mean moving these elsewhere.

pkg/remote/proxy.go
timburks: For sure, yes, that's putting it politely. In addition to warning people about this, I wonder if we could limit its blast radius by adopting some testing conventions -- if all projects used in testing matched a certain pattern (e.g. `".*test.*"`), then we could only clear those projects and not others, though this might still break a few tests that enumerate all the projects in a running server.

pkg/wipeout/wipeout_test.go
```diff
@@ -34,24 +35,59 @@ func TestWipeout(t *testing.T) {
...
 	}); err != nil && status.Code(err) != codes.NotFound {
 		t.Fatalf("Setup: failed to delete test project: %s", err)
 	}
 
-	if _, err := adminClient.CreateProject(ctx, &rpc.CreateProjectRequest{
-		ProjectId: project.ProjectID,
-		Project:   &rpc.Project{},
-	}); err != nil {
-		t.Fatalf("Setup: Failed to create test project: %s", err)
+	seed := []*rpc.Artifact{
```
timburks, line 51: I'm leaning towards writing some loops to make this more compact.

pkg/remote/proxy.go
timburks: I just moved the remote proxy into server/registry/test/remote

pkg/wipeout/wipeout_test.go
timburks: Done. I could go either way with this.

timburks:
#664 was merged and includes everything here.

Issue #641, opened 2022-07-06
Registry controller: Parallel execution of actions leads to lost updates


shrutiparabgoogle:
This issue was recently discovered in the way the controller executes actions.

An example of the dependency relationship between resources is as follows:
```
- pattern: apis/-/versions/-/specs/-/artifacts/conformance-receipt
   receipt: true
   dependencies:
     - pattern: $resource.spec
     - pattern: artifacts/apihub-styleguide
   action: registry compute conformance $resource.spec
- pattern: apis/-/versions/-/specs/-/artifacts/score-receipt
   dependencies:
      - pattern: $resource.spec
      - pattern: artifacts/conformance-apihub-styleguide
   action: registry compute score $resource.spec
```

In the above definition, the controller reads these dependencies, fetches all the resources in the form of a snapshot of the state of the registry. 

Consider the case where:
- the `conformance-report` was computed for the first time in the last run.
- there has been a spec update since the last run.

In the current run:
- The controller will notice that there is a spec update and hence it will generate the action for compute conformance report.
- Since the score artifact doesn't exist yet, it will also generate an action to compute the score.
- Now, both the above actions will be submitted to a task queue to be run in parallel.
- Depending on when the task queue delegates this action to the workers for execution, there is a scenario where:
   * the score command fetches the artifact to be processed for score computation. After this fetch happens, the conformance command gets executed in parallel and updates the conformance report.
   * Then the score gets computed using the old copy and uploaded to the registry. 
   * The next time the controller runs and compares the timestamps of the conformance report and the score, it will see that the score was generated after the conformance report and hence the controller will conclude that the score is already up-to-date.
- In this whole scenario we lost an update to the conformance report and the score will remain outdated unless there is one more update to the conformance report in the future.

**Solution:**
- Modify the [timestamp condition](https://github.com/apigee/registry/blob/main/cmd/registry/controller/controller.go#L351) used by the controller to determine if an update is needed or not.
- The current condition simply checks if the resource was generated after it's dependencies. This condition can be modified to include a threshold:
```
resources should be at least N seconds newer than their dependencies, where N is the maximum time to compute the resource.
```

shrutiparabgoogle:
There are few options to what N should be in the updated condition:
```
resources should be at least N seconds newer than their dependencies, where N is the maximum time to compute the resource.
```

[...]

PR #653, opened 2022-07-08
Update timestamp condition to figure out resource regeneration

shrutiparabgoogle:
Fixes https://github.com/apigee/registry/issues/641


cmd/registry/controller/controller.go
```diff
@@ -34,7 +33,7 @@ type Action struct {
 
 func ProcessManifest(
 	ctx context.Context,
-	client connection.RegistryClient,
+	listerClient Lister,
```
timburks, line 36: to reduce confusion, could "listerClient" be simply "lister"?

cmd/registry/controller/controller.go
```diff
@@ -68,7 +67,7 @@ func ProcessManifest(
 
 func processManifestResource(
 	ctx context.Context,
-	client connection.RegistryClient,
+	client Lister,
```
timburks, line 70: Could "client" be "lister" instead? I think that would be more clear.

cmd/registry/controller/list.go
```diff
@@ -0,0 +1,134 @@
...
+	"context"
+
+	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/cmd/registry/patterns"
+	"github.com/apigee/registry/pkg/connection"
+	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/registry/names"
+)
+
+type Lister interface {
```
timburks, line 13: 1. Why does Lister exist? (for testability?)
2. Why is Lister defined in the controller and not at some more general level (`core` or the top-level `pkg` directory)?

shrutiparabgoogle: The lister is an interface which can be faked while testing time sensitive controller features, specially the updated timestamp condition. It is currently defined in the controller because the interface includes only the functions which are used by the controller.

I explored the option of defining a generic registry client interface, but that complicates the  process where faking it means implementing almost all of the registry's functions. 

This is a much lighter implementation where I only need to care about the functions that the controller is using.

cmd/registry/patterns/resources.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/apigee/registry/server/registry/names"
 )
 
+const ResourceUpdateThresholdSeconds = 2
```
seaneganx, line 27: Maybe update this to a duration: `var ResourceUpdateThreshold = 2 * time.Second`. Then instead of converting to seconds before comparing, you can use the `time` package `Before` and `After` methods.

cmd/registry/controller/controller.go
shrutiparabgoogle: My intention of naming it client is to indicate that the functionality is still like a client to talk to the registry, but a specific implementation of the client: `Lister`, which focuses on `List` calls.

However, I have used this as `listerClient` in some parts of the code. For consistency, I am going to name everything as client.

cmd/registry/patterns/resources.go
shrutiparabgoogle: Done.

cmd/registry/controller/controller.go
shrutiparabgoogle: Renamed to client.

cmd/registry/controller/controller.go
```diff
@@ -348,8 +346,11 @@ func needsUpdate(
 			return false, nil
 		}
 
-		if maxUpdateTime.After(targetResourceTime) {
-			return true, nil // Take action if atleast one dependency timestamp is later than resource timestamp
+		// Take action if atleast one dependency timestamp is later than resource timestamp
+		// Or if the target resource is less than 2 seconds newer compared to the dependencies.
```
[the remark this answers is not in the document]

cmd/registry/controller/controller.go
```diff
@@ -348,8 +346,11 @@ func needsUpdate(
...
+		// The second condition is required to avoid this scenario: https://github.com/apigee/registry/issues/641
+		if maxUpdateTime.After(targetResourceTime) || targetResourceTime.Sub(maxUpdateTime).Seconds() < patterns.ResourceUpdateThresholdSeconds {
```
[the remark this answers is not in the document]

cmd/registry/controller/list.go
seaneganx: Maybe `RegistryLister` belongs in `resolve.go` then. The controller defines the `Lister` interface to tell `resolve.go` and `controller-timestamp_test.go` what it needs to fetch resources. Then `resolve.go` and `controller-timestamp_test.go` have their own types that satisfy that interface.

As I said before though, if the types become useful in other places, we can move them to a more general place like `core`, `pkg`, etc. as Tim suggested.

Thoughts on this approach?

seaneganx (approved):
This looks good to me. Just one open question about where the `RegistryLister` type should be defined.

cmd/registry/controller/list.go
shrutiparabgoogle: I think what you are suggesting makes sense to me. Just to mention here, there is going to be a similar refactor for the scoring package where we will define an interface to make registry calls and update the timestamp condition. The `lister` interface and the one I define for scores might have a potential of overlap. 

So I am proposing that we keep things as is here for now. Once I define the interface for scoring package, we can do either of the following two things:
* merge the two interfaces and move the definition of RegistryLister (might be renamed) to a common package like `core`.
* Or if we decide that there is not enough overlap, move the `RegistryLister` definition to `resolve.go`.

cmd/registry/controller/controller.go
timburks: Could we put `Client` in the type name of `Lister`? As it is, these terms are so conceptually distant that they introduce confusion to new readers of the code. `ListingClient` might resolve this.

shrutiparabgoogle: I agree.

timburks (approved):

seaneganx (approved):

PR #655, opened 2022-07-08
Add simple tests for the MigrateDatabase admin rpc.

timburks:
This increases test coverage of `actions_migrate_database.go` from zero to 73% (all but a few returns from error conditions).

### DIFF
--- a/server/registry/actions_migrate_database_test.go
+++ b/server/registry/actions_migrate_database_test.go
@@ -1,0 +1,109 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package registry
+
+import (
+	"context"
+	"testing"
+
+	"github.com/apigee/registry/rpc"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/genproto/googleapis/longrunning"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/testing/protocmp"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+func TestMigrateDatabase(t *testing.T) {
+	ctx := context.Background()
+	server := defaultTestServer(t)
+
+	req := &rpc.MigrateDatabaseRequest{}
+	metadata, err := anypb.New(&rpc.MigrateDatabaseMetadata{})
+	if err != nil {
+		t.Fatalf("MigrateDatabase(%+v) test failed to build expected metadata: %s", req, err)
+	}
+	response, err := anypb.New(&rpc.MigrateDatabaseResponse{
+		Message: "OK",
+	})
+	if err != nil {
+		t.Fatalf("MigrateDatabase(%+v) test failed to build expected response: %s", req, err)
+	}
+	want := &longrunning.Operation{
+		Name:     "migrate",
+		Done:     true,
+		Metadata: metadata,
+		Result:   &longrunning.Operation_Response{Response: response},
+	}
+
+	got, err := server.MigrateDatabase(ctx, req)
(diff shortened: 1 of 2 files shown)


server/registry/actions_migrate_database_test.go
```diff
@@ -0,0 +1,109 @@
...
+			req:  &rpc.MigrateDatabaseRequest{Kind: "unsupported"},
+			want: codes.InvalidArgument,
+		},
+		{
+			req:  &rpc.MigrateDatabaseRequest{Kind: "invalid"},
+			want: codes.InvalidArgument,
+		},
+	}
+
+	for _, test := range tests {
```
theganyo, line 95: Normally, there would be a `t.Run(test.desc)` here. Alternatively, remove the `desc string` var from the test struct.

server/registry/actions_migrate_database_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
```
seaneganx, line 1: 2022

server/registry/actions_migrate_database_test.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+func TestMigrateDatabase(t *testing.T) {
+	ctx := context.Background()
+	server := defaultTestServer(t)
+
+	req := &rpc.MigrateDatabaseRequest{}
+	metadata, err := anypb.New(&rpc.MigrateDatabaseMetadata{})
+	if err != nil {
+		t.Fatalf("MigrateDatabase(%+v) test failed to build expected metadata: %s", req, err)
```
seaneganx, line 37: This error message doesn't seem to match the cause of the error. We're not calling `MigrateDatabase` yet, and we haven't used the `req` variable yet either. Same for the next error.

Maybe `Setup: Failed to build response message: %s`?

server/registry/actions_migrate_database_test.go
seaneganx: Agreed. I just don't want us to use `Fatalf` without subtests because it will prevent unrelated test cases from executing.

server/registry/actions_migrate_database_test.go
```diff
@@ -0,0 +1,109 @@
...
+	for _, test := range tests {
+		_, err := server.MigrateDatabase(ctx, test.req)
+		if test.want == codes.OK {
+			if err != nil {
+				t.Fatalf("MigrateDatabase(%+v) returned error: %s (expected %s)", test.req, err, test.want)
+			}
+		} else {
+			if err == nil {
+				t.Fatalf("MigrateDatabase(%+v) succeeded (expected %s)", test.req, test.want)
+			} else if status.Code(err) != codes.InvalidArgument {
```
seaneganx, line 104: Why is `codes.InvalidArgument` hard-coded into this check? Are we assuming `test.want` is always either `codes.OK` or `codes.InvalidArgument`? Should this be `test.want` instead?

server/registry/actions_migrate_database_test.go
```diff
@@ -0,0 +1,109 @@
...
+				t.Fatalf("MigrateDatabase(%+v) returned error: %s (expected %s)", test.req, err, test.want)
+			}
+		}
```
seaneganx, line 107: Can we simplify this? In other cases where we test status codes we use code like the following. 

```
if _, err := server.GetArtifact(ctx, test.req); status.Code(err) != test.want {
	t.Errorf("GetArtifact(%+v) returned status code %q, want %q: %v", test.req, status.Code(err), test.want, err)
}
```

server/registry/actions_migrate_database_test.go
timburks: thanks, fixed

server/registry/actions_migrate_database_test.go
timburks: These are building the *expected* response messages, not the actual ones. I'll update these to "failed to build expected response (metadata|message)".

server/registry/actions_migrate_database_test.go
timburks: right, thanks - this should be test.want, changed

server/registry/actions_migrate_database_test.go
timburks: OK, updated to use the test descriptions inside t.Run()

server/registry/actions_migrate_database_test.go
timburks: Thanks, that feels a bit dense but good to have as a common pattern. Updated.

timburks:
@theganyo @seaneganx I'll merge this now since I think I've addressed everything and the scope is small. Feel free to follow up if you see any problems. Thanks!

Issue #649, opened 2022-07-07
Add CLI commands to manage config file


theganyo:
See #625

PR #659, opened 2022-07-13
Enable config file usage and flags

### LINKED ISSUE #647: Load and use file-based configuration
See #625

theganyo:
Enables config file usage and flags as follows:

Flags enabled for registry CLI:

```
  -c, --config string             Name of a settings profile or path to config file
      --registry.address string   the server and port of the registry api (eg. localhost:8080)
      --registry.insecure         set to true if client must connect via http (not https)
      --registry.token string     the token to use for authorization to registry
```

Notes:

* This is 100% backward compatible
* `go test` will use env vars, not flags
* Settings are honored in prioritized order from: 1. flags, 2. env vars, 3. config file
* Config files are assumed to be stored in $HOME/.config/registry
* However, `--config` flag can specify an explicit path to a settings file directly

Config file management will be handled in #649 

Fixes #647

### DIFF
--- a/pkg/connection/settings.go
+++ b/pkg/connection/settings.go
@@ -1,0 +1,188 @@
+// Copyright 2022 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connection
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+// Flags defines Flags that may be bound to the Settings. Use like:
+// `cmd.PersistentFlags().AddFlagSet(connection.Flags)`
+var Flags *pflag.FlagSet
+
+var configPath string
+var activeConfigPointerFilename = "active_config"
+
+func init() {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		panic(err)
+	}
+	configPath = filepath.Join(home, ".config/registry")
+
+	Flags = pflag.NewFlagSet("registry", pflag.ExitOnError)
+	Flags.StringP("config", "c", "", "Name of a settings profile or path to config file")
+	Flags.String("registry.address", "", "the server and port of the registry api (eg. localhost:8080)")
+	Flags.Bool("registry.insecure", false, "set to true if client must connect via http (not https)")
+	Flags.String("registry.token", "", "the token to use for authorization to registry")
+}
+
+// Settings configure the client.
+type Settings struct {
+	Address  string `mapstructure:"address"`  // service address
+	Insecure bool   `mapstructure:"insecure"` // if true, connect over HTTP
+	Token    string `mapstructure:"token"`    // bearer token
+}
+
+func (s Settings) Validate() error {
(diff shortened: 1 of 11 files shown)


timburks:
Let's watch out for possible collisions between the `--registry.*` flags and options in autogenerated commands.
```
$ registry rpc update-api --help
UpdateApi can be used to modify a specified API.

Usage:
  registry rpc update-api [flags]

Flags:
      --allow_missing                       If set to true, and the api is not found, a new...
      --api.annotations stringArray         key=value pairs. Annotations attach non-identifying metadata to...
      --api.availability string             A user-definable description of the availability...
      --api.description string              A detailed description.
      --api.display_name string             Human-meaningful name.
      --api.labels stringArray              key=value pairs. Labels attach identifying metadata to resources....
      --api.name string                     Resource name.
      --api.recommended_deployment string   The recommended deployment of the API.  Format:...
      --api.recommended_version string      The recommended version of the API.  Format:...
      --from_file string                    Absolute path to JSON file containing request payload
  -h, --help                                help for update-api
      --update_mask.paths strings           The set of field mask paths.

Global Flags:
  -j, --json      Print JSON output
  -v, --verbose   Print verbose output
```

I think we're ok now because none of our API methods are taking a `registry` message as a request argument.

theganyo:
Hmm. I don't know how we could ensure that we don't collide with the current autogen scheme since it does't have it's own namespace...

timburks:
I think the current option names are good. Just wanted to point out the possibility of future surprises.

pkg/connection/settings.go
```diff
@@ -0,0 +1,190 @@
...
+	home, err := os.UserHomeDir()
+	if err != nil {
+		panic(err)
+	}
+	configPath = filepath.Join(home, ".config/registry")
+
+	Flags = pflag.NewFlagSet("registry", pflag.ExitOnError)
+	Flags.StringP("config", "c", "", "Name of a settings profile or path to config file")
+	Flags.String("registry.address", "", "the server and port of the registry api (eg. localhost:8080)")
+	Flags.Bool("registry.insecure", false, "set to true if client must connect via http (not https)")
```
timburks, line 46: could we clarify this, possibly with "if specified, client connects via http (not https)"

timburks (commented):
This is great!

We're going to need to add some more special handling for the generated RPC handlers. These two client creations are in generated code that will not use the new configuration.  I think it might be best to just patch these two functions, replacing their entire contents with new code that generates AdminClient and RegistryClient using `pkg/connection`. 

https://github.com/apigee/registry/blob/7bc06ba752571bcb29b327c4351e0a810cc9e283/cmd/registry/cmd/rpc/generated/admin_service.go#L76

https://github.com/apigee/registry/blob/7bc06ba752571bcb29b327c4351e0a810cc9e283/cmd/registry/cmd/rpc/generated/registry_service.go#L103

Other patching is done [here](https://github.com/apigee/registry/blob/main/tools/GENERATE-CLI.sh).

pkg/connection/settings.go
theganyo: yup

theganyo:
The commit I just pushed fixes the generated rpc cmds to use the same files and flags as the main registry cmd. However, it leaves quite a bit of dead code in the generate clients that we need to work on removing. The following is a diff example of what we have yet to do: https://gist.github.com/theganyo/99150cbc129557d991205c87528aa972

timburks (approved):
Looks great!

I'm ok with the patch code as-is. It will be easier to simplify it after #660 is merged - then I think all the patches could be done with the same set of (parameterized) steps.

pkg/connection/settings.go
```diff
@@ -0,0 +1,190 @@
...
+// `cmd.PersistentFlags().AddFlagSet(connection.Flags)`
+var Flags *pflag.FlagSet
+
+var configPath string
+var activeConfigPointerFilename = "active_config"
+
+func init() {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		panic(err)
```
timburks, line 39: nit: I think there's something in the style guide about panics, but it seems pretty hard to get an error here: https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/os/file.go;l=500

PR #665, opened 2022-07-18
replace NewClient with NewRegistryClient

### LINKED ISSUE #660: pkg/connection: let's deprecate NewClient and NewClientWithSettings and add NewRegistryClient and NewRegistryClientWithSettings
https://github.com/apigee/registry/blob/7bc06ba752571bcb29b327c4351e0a810cc9e283/pkg/connection/client.go#L74

theganyo:
Fixes #660

### DIFF
--- a/gapic/doc.go
+++ b/gapic/doc.go
@@ -59,7 +59,7 @@
 //
 // Use of Context
 //
-// The ctx passed to NewClient is used for authentication requests and
+// The ctx passed to NewRegistryClient is used for authentication requests and
 // for creating the underlying connection, but is not used for subsequent calls.
 // Individual methods on the client use the ctx given to them.
 //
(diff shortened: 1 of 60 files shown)


timburks (approved):

gapic/doc.go
```diff
@@ -59,7 +59,7 @@
 //
 // Use of Context
 //
-// The ctx passed to NewClient is used for authentication requests and
+// The ctx passed to NewRegistryClient is used for authentication requests and
```
timburks, line 62: This is generated code, and this change gets clobbered when we rerun ./tools/GENERATE-GAPIC.sh. I think it's an upstream problem in the generator.

theganyo: Should we merge and file a new issue on this?

theganyo: rolled back the doc change

timburks:
I filed [this issue](https://github.com/googleapis/gapic-generator-go/issues/1077) about the `doc.go` problem.

PR #670, opened 2022-07-20
Minor cleanup to CI tests

timburks:
- replace `make test` with `go test ./...`
- use built-in server for coverage testing

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -83,7 +83,7 @@ jobs:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       run: |
         go clean -testcache
-        go test -race -coverprofile=coverage.txt -covermode=atomic ./...
+        go test -race -coverprofile=coverage.txt -covermode=atomic -coverpkg=./server/... ./...
         curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import # One-time step
         curl -Os https://uploader.codecov.io/latest/linux/codecov
         curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -83,7 +83,7 @@ jobs:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       run: |
         go clean -testcache
-        go test -race -coverprofile=coverage.txt -covermode=atomic -coverpkg=all ./...
+        go test -race -coverprofile=coverage.txt -covermode=atomic -coverpkg=./server/... ./...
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -78,6 +78,21 @@ jobs:
     - name: Test registry server with PostgreSQL
       run: go test ./server/registry -postgresql
 
+    - name: Compute code coverage
```
timburks, line 81: I moved this earlier to clarify that it's *not* using the standalone server (notice that the `APG_REGISTRY` environment variable settings are also removed)

.github/workflows/go.yml
```diff
@@ -78,6 +78,21 @@ jobs:
     - name: Test registry server with PostgreSQL
       run: go test ./server/registry -postgresql
```
timburks, line 79: Should we be clearing the test cache here?

theganyo: It won't do anything as `go test` hasn't been run before.

timburks: go test is being run on line 65 and again on line 79. Should we clear the cache before the run on line 79?

.github/workflows/go.yml
theganyo: Is there any reason to not to combine this with the `Test everything with SQLite` test?

.github/workflows/go.yml
timburks: From the runtime of the test, it appears to be doing something but there are no code changes. I wonder if changing the test flags is enough to clear the cache. *off to read docs*

timburks: "To disable test caching, use any test flag or argument other than the cacheable flags" https://pkg.go.dev/cmd/go/internal/test

theganyo: JFYI, a quick trick to bypass the cache is to add a count flag, ie. `go test ./... -count=1`.

.github/workflows/go.yml
theganyo: Hmm. If we have postgres-specific code (where is that?) and that is our targeted production platform, then I think we should emit the coverage with the postgres test.

theganyo (approved):

.github/workflows/go.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       run: |
         go clean -testcache
-        go test -race -coverprofile=coverage.txt -covermode=atomic ./...
+        go test -race -coverprofile=coverage.txt -covermode=atomic -coverpkg=./server/... ./...
```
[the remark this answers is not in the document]
timburks: weirdly, when I changed to `coverpkg=./...`, the coverage dropped back to 50%.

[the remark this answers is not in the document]
theganyo: That's what I had seen as well.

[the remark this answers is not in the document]
timburks: I'm reverting any attempt to tweak `coverpkg` for this PR. Feel free to pick it up again.

PR #671, opened 2022-07-25
Add CLI commands to manage config file 

### LINKED ISSUE #649: Add CLI commands to manage config file

theganyo:
Commands added:
```
registry config [command]

Available Commands:
  configurations Maintain configuration profiles
  get            Print the value of a property.
  list           List properties for the currently active configuration.
  set            Set the value of a property.
  unset          Unset the value of a property.


registry config configurations [command]

Available Commands:
  activate    Activates an existing named configuration.
  create      Creates a new named configuration.
  delete      Deletes a named configuration.
  describe    Describes a named configuration by listing its properties.
  list        Lists existing named configs.
```

Closes #649


pkg/connection/client.go
```diff
@@ -50,18 +50,18 @@ func clientOptions(settings Settings) ([]option.ClientOption, error) {
 // RegistryClient is a client of the Registry API
 type RegistryClient = *gapic.RegistryClient
 
-// NewRegistryClient creates a new GAPIC client using environment variable settings.
+// NewRegistryClient creates a new client using the active Config.
 func NewRegistryClient(ctx context.Context) (RegistryClient, error) {
-	settings, err := ActiveSettings()
+	config, err := ActiveConfig()
```
seaneganx, line 55: Have we considered removing this "on demand" loading of config values? In most (all?) cases we create one client at the start of the program and pass it along to anywhere it's needed. By loading the config once and setting up the client accordingly, we might be able to eliminate quite a bit of code.

cmd/registry/cmd/config/set.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func setCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set PROPERTY VALUE",
```
seaneganx, line 25: Do we expect users to want to set multiple properties with this command? This seems like it can be generalized to handle multiple properties without much difficulty.

cmd/registry/cmd/config/configurations/activate.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/apigee/registry/pkg/connection"
+	"github.com/spf13/cobra"
+)
+
+func activateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "activate CONFIGURATION_NAME",
+		Short: "Activates an existing named configuration.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
seaneganx, line 29: Most of our other commands use `log.Fatalf` instead of returning errors from `RunE`. I'm open to changing how we do it, but I think we should be consistent.

cmd/registry/cmd/config/configurations/activate.go
```diff
@@ -0,0 +1,51 @@
...
+			if err != nil {
+				return fmt.Errorf("Cannot read config %q: %v", name, err)
+			}
+
+			err = connection.ActivateConfig(name)
+			if err != nil {
+				return fmt.Errorf("Cannot activate config %q: %v", name, err)
+			}
+
+			fmt.Printf("Activated %q.\n", name)
```
seaneganx, line 46: Similar to error handling, we usually use `log.Infof` to print information for users.

cmd/registry/cmd/config/configurations/list.go
```diff
@@ -0,0 +1,61 @@
...
+			}
+
+			names := make([]string, 0, len(configs))
+			for n := range configs {
+				names = append(names, n)
+			}
+			sort.Strings(names)
+
+			w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+			fmt.Fprintln(w, "NAME\tIS_ACTIVE\tADDRESS\tINSECURE")
```
seaneganx, line 51: Does this depend on the user's terminal tabwidth at all? Using tabs for formatting can make output really hard to understand for users with different terminal settings if we aren't careful.

cmd/registry/cmd/config/configurations/list.go
```diff
@@ -0,0 +1,61 @@
...
+			for _, name := range names {
+				config := configs[name]
+				fmt.Fprintf(w, "%s\t%t\t%s\t%t\n", name, name == activeName, config.Address, config.Insecure)
+			}
+			w.Flush()
```
seaneganx, line 56: Consider calling `defer w.Flush()` immediately after creating the tabwriter instead. It keeps the relevant code co-located and guarantees we'll flush before the function ends even if a new exit point appears in the future.

cmd/registry/cmd/config/set.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			fmt.Printf("Updated property %q.\n", name)
+			return nil
+		},
+	}
+	return cmd
+}
+
+func contains[T comparable](arr []T, x T) bool {
```
seaneganx, line 58: This only seems to be used with string slices and a string target. Can we drop the generics until we need it to be generic?

pkg/connection/config.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+type ValidationError struct {
+	Field      string
+	Validation string
+}
+
+func (e ValidationError) Error() string {
+	return fmt.Sprintf("%s: %s", e.Field, e.Validation)
+}
```
seaneganx, line 306: This only seems to be used in one place, as `ValidationError{"registry.address", "required"}`. Can we replace it with `var AddressRequiredError = errors.New("registry.address: required")`?

pkg/connection/config_test.go
```diff
@@ -0,0 +1,341 @@
...
+	origConfigPath := ConfigPath
+	ConfigPath = tmpDir
+	return func() {
+		ConfigPath = origConfigPath
+		os.RemoveAll(tmpDir)
+	}
+}
+
+func TestActiveSettings(t *testing.T) {
+	defer cleanConfigDir(t)()
```
seaneganx, line 39: Usually in tests `t.Cleanup` is preferred to `defer` because it works even if the test becomes nested inside another. I'm not sure about the behavior of `defer` when `t.Fatal` is called, but it definitely works as expected with `t.Cleanup`.

seaneganx (commented):
Lots of questions and some suggestions.

pkg/connection/config.go
theganyo: It is today, but as this is library code (in `pkg`) the point of this is to have a well-structured format for clients.

pkg/connection/grpctest/server.go
```diff
@@ -41,13 +41,13 @@ import (
 //   ... run test here ...
 // }
 func NewIfNoAddress(rc registry.Config) (*Server, error) {
-	settings, _ := connection.ReadSettings("")
-	if settings.Address != "" && settings.Validate() == nil {
-		log.Printf("Client will use remote registry at: %s", settings.Address)
+	config, _ := connection.ReadConfig("")
```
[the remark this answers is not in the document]
theganyo: It can return an error, I just don't care. I added a comment.

pkg/connection/config_test.go
theganyo: ok

cmd/registry/cmd/config/configurations/activate.go
theganyo: discussion above

cmd/registry/cmd/test/test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// Capture captures stdin / stdout for a command. Input can provided
```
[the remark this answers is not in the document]
theganyo: Good point. I think this can be redone a bit.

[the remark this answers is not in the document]
theganyo: done

timburks:
@theganyo for a high-level summary, could you edit the PR description to list the commands that are added?

theganyo:
> @theganyo for a high-level summary, could you edit the PR description to list the commands that are added?

done

timburks:
This may be out of scope for this initial PR but...

What do we want the startup experience to be?
```
$ registry config list
insecure = false

Your active configuration is: "".
```
I think here we might want to say "you have no active configuration; see `registry config configurations` to see how to create and manage configurations`.

```
$ registry config configurations list
Error: Cannot read configs: open /home/timburks/.config/registry: no such file or directory
Cannot read configs: open /home/timburks/.config/registry: no such file or directory
```

Maybe: "No configuration directory exists. We expect configurations to be stored in ..."

```
$ registry config configurations create default
Error: Cannot create config "default": open /home/timburks/.config/registry/default: no such file or directory
Cannot create config "default": open /home/timburks/.config/registry/default: no such file or directory
```

Should we automatically create the needed directories?

theganyo:
Yes, this PR hadn't been intended to address installation... but it's a very good question that we need to answer. I think we should think about automatically creating or having the option to create the same configurations that we put into `auth/`. (Also: We should definitely at least create the directory if it's missing, that was an oversight I will fix now.)

theganyo:
Also: Instead of telling the user to go do another thing, how about we address it more directly like, "You have no configurations. Please enter a configuration name to create: "

timburks (approved):

theganyo:
On second thought, I think I like your idea better: Directing the user to another command may not be the most convenient, but putting the client into a new mode may be jarring and could cause scripts to hang.

cmd/registry/cmd/test/test.go
```diff
@@ -0,0 +1,60 @@
...
+		return "", err
+	}
+	_ = w.Close()
+	got, err := ioutil.ReadAll(r)
+	_ = r.Close()
+	if err != nil {
+		return "", err
+	}
+	return string(got), err
+}
```
seaneganx, line 60: Is this doing something different than the following?

```
if input != "" {
	cmd.SetIn(strings.NewReader(input))
}

var got bytes.Buffer
cmd.SetOut(&got)

if err := cmd.Execute(); err != nil {
	return "", err
}

return string(got), nil
```

Note: I didn't actually compile this code, but I think something similar should work.

If not, is this something we want to abstract away in a test utility package? I think we can wire up the i/o more clearly in the tests.

theganyo: Good simplification, thanks.

pkg/connection/client.go
theganyo: I'm unsure where we'd save a lot of code, but I generally like the idea of separating the concerns. We should maybe look at that in another issue, though?

pkg/connection/config.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"go.uber.org/multierr"
+)
+
+const ActiveConfigPointerFilename = "active_config"
+
+// Flags defines Flags that may be bound to a Configuration. Use like:
+// `cmd.PersistentFlags().AddFlagSet(connection.Flags)`
+var Flags *pflag.FlagSet = createFlagSet()
```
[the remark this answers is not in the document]
theganyo: This actually seems pretty tricky... Instead of trying to rework it right now, we could potentially change access to the FlagSet via a public accessor for now instead of the public variable... That would at least give us a point of control for future work. Otherwise, I think this will be a significant refactor that should probably also incorporate the separating the config and client concerns mentioned in the other comment - and maybe in a follow up PR?

cmd/registry/cmd/config/config_test.go
```diff
@@ -100,108 +100,116 @@ func TestConfig(t *testing.T) {
 
 	cmd := setCommand()
 	cmd.SetArgs([]string{"test", "test"})
-	_, err = capture(cmd, "")
-	if err.Error() != `Config has no property "test".` {
+	want := `Config has no property "test".`
+	if err := cmd.Execute(); err.Error() != want {
```
seaneganx, line 103: Can we compare error values instead of hardcoding a string? In other words, can we make sure this test will pass even if we change the error message? Maybe that means defining `type MissingPropertyError struct{field string}` and checking if `err != MissingPropertyError{"test"}`.

seaneganx (approved):
Please remove the dead code before submitting (see the linter annotation).

cmd/registry/cmd/config/config_test.go
theganyo: sure

theganyo: sure

PR #673, opened 2022-07-25
Bump protoc version to 3.21.3

seaneganx:


tools/PROTOS.sh
```diff
@@ -51,7 +51,7 @@ function clone_common_protos {
 # Require a specific version of protoc for generating files.
 # This stabilizes the generated file output, which includes the protoc version.
 . tools/PROTOC-VERSION.sh
-if [ "$(protoc --version)" != "libprotoc $PROTOC_VERSION" ]; then
```
theganyo, line 54: This is odd. Why is the `3.` separately hard-coded from the rest of the version now?

seaneganx: Because https://github.com/protocolbuffers/protobuf/releases doesn't include the major version in the asset names anymore, but the output from `protoc --version` still includes it.

We use the `$PROTOC_VERSION` variable to build the asset URL.

https://github.com/apigee/registry/blob/9bf1213bc7bf2ca5740bd12dd19a8fc50f7eeee0/tools/FETCH-PROTOC.sh#L32-L38

Hardcoding the `3.` in our version checks is easier than trimming it from the variable before building that URL.

theganyo: It's a confusing strategy to me... It sounds like they are still doing major version releases on a per-language basis: https://developers.google.com/protocol-buffers/docs/news/2022-05-06. And, if you look at the tags, they are still tagging (if not creating "releases") that match the full major versioning scheme: https://github.com/protocolbuffers/protobuf/tags.

theganyo (approved):

tools/PROTOS.sh
```diff
@@ -51,7 +51,7 @@ function clone_common_protos {
...
-    echo "Please update your protoc to version $PROTOC_VERSION, the current required version as specified in tools/PROTOC_VERSION.sh"
+if [ "$(protoc --version)" != "libprotoc 3.$PROTOC_VERSION" ]; then
+    echo "Please update your protoc to version 3.$PROTOC_VERSION, the current required version as specified in tools/PROTOC_VERSION.sh"
```
timburks, line 55: So now it seems that "3" is the $GO_PROTOC_MAJOR_VERSION. Someday we might want to make that explicit instead of leaving the mysterious "3." around, which someone in the future might think is a typo.

PR #674, opened 2022-07-25
Support ordering in list operations

### LINKED ISSUE #54: Registry API: Support ordering in list operations
List methods currently do not support custom ordering. There are many cases where this could be useful, such as listing APIs in order by their creation time. Ordering should be supported as described in [AIP-132](https://google.aip.dev/132#ordering).

seaneganx:
Fixes #54.


theganyo:
Is there a way this be based on the branch for PR #673 instead of including that commit in this PR as well?

seaneganx:
>Is there a way this be based on the branch for PR https://github.com/apigee/registry/pull/673 instead of including that commit in this PR as well?

My intention is to merge that PR quickly, eliminating the duplication from this one.

I don't know of any way to specify that one PR depends on another. The closest way I know would be to open a PR onto that other branch, but then we would need to merge this PR onto the intermediate branch before merging the combined PR into main.

theganyo:
> > Is there a way this be based on the branch for PR #673 instead of including that commit in this PR as well?
> 
> My intention is to merge that PR quickly, eliminating the duplication from this one.
> 
> I don't know of any way to specify that one PR depends on another. The closest way I know would be to open a PR onto that other branch, but then we would need to merge this PR onto the intermediate branch before merging the combined PR into main.

It looks like this "stacked PR" idea is discussed in depth here: https://blog.logrocket.com/using-stacked-pull-requests-in-github/.

But I'm wondering, how are these PRs dependent?

seaneganx:
>But I'm wondering, how are these PRs dependent?

Theoretically they aren't dependent at all, but I didn't have easy access to the version of protoc we had pinned. Instead of making this "support ordering in list operations and bump protoc version," I thought it would be good to do the protoc version bump in a separate PR.

theganyo (approved):

server/registry/internal/storage/filtering/filtering.go
```diff
@@ -60,22 +55,22 @@ func (f *Filter) Matches(model map[string]interface{}) (bool, error) {
 	return match, nil
 }
 
-func NewFilter(filter string, fields []Field) (Filter, error) {
+func NewFilter(filter string, fields map[string]FieldType) (Filter, error) {
```
timburks, line 58: Is this change related to ordering support? (I'm not seeing a relationship on initial reading)

seaneganx: Good catch, I made this change at an intermediate step during development because I wanted to avoid referencing types from the `filtering` package. That didn't work out in the end: `func gormOrdering(ordering string, fields map[string]filtering.FieldType)` was the best I could do instead of `func gormOrdering(ordering string, fields []filtering.Field)`.

That said, this change isn't really important to the PR (although I think it's still worth making - this type really doesn't add value). I can take some time to revert this and open a new PR with the small improvement, but IMO it's insignificant enough to keep it here and save the effort of reorganizing. If it wasn't lumped into a larger commit I could easily separate it, but that's not the case here.

Let me know if you want me to separate it or just drop it entirely, otherwise I'll leave it here.

timburks: Thanks, to keep our commits easy to review I'd like to do this in two PRs - a first one that makes the relatively low impact / high line count change to the filters and then we can merge that into this one which will hopefully then be a high impact / low line count change that is easy to refer back to.

timburks: Regarding the filtering change specifically: I agree that the exported type isn't adding any value. The iteration order becomes unstable when we use a map, but that shouldn't be affecting anything (considering the way this is used).

seaneganx: #678

server/registry/internal/storage/list.go
```diff
@@ -109,6 +110,51 @@ var artifactFields = map[string]filtering.FieldType{
 	"size_bytes":  filtering.Int,
 }
 
+func gormOrdering(ordering string, fields map[string]filtering.FieldType) (string, error) {
```
timburks, line 113: Could we add  a brief comment here explaining what this is doing?

something like: "parses specified order_by values, generating a string that gets added to the gorm queries that implements the specified ordering"

timburks (approved):

server/registry/internal/storage/list.go
seaneganx: Done.

PR #681, opened 2022-08-10
Registry tool: add auth handling

### LINKED ISSUE #675: Registry tool: add auth handling
From https://github.com/apigee/registry/issues/625#issuecomment-1172540159:
> We also need to handle authentication. `gcloud auth login` is a pattern that we might follow but involves an OAuth transaction that I would rather leave to gcloud. Among auth subcommands, we might have `registry auth gcloud` to configure the tool to always get its credentials from running `$(gcloud print-access-token)`, `registry auth application-default` might configure the tool to use application default credentials, and `registry auth token` might set a specific token. (Should these auth subcommands be under config? They aren't for gcloud but we might argue that different registries require different authentication when they are hosted in different ways. However, because this is unclear, I would default to following the gcloud precedent of keeping them separate.)

theganyo:
Set and view registry CLI authentication.

Added auth command:
```
$ registry auth
Manage registry CLI authentication.

Usage:
  registry auth [command]

Available Commands:
  print-token Invoke token source and print token.
```

To set or get `token-source`, use commands:
```
$ registry config set token-source
$ registry config get token-source
```

Fixes #675


timburks:
Is there a strong reason to prefer "generator" over other command names? My initial preference on seeing it was to suggest "provider", but when I looked for other precedents I found "source" in [the Go oauth2 package](https://pkg.go.dev/golang.org/x/oauth2@v0.0.0-20220808172628-8227340efae7#TokenSource), which I think has a stronger case to be made for it.

"generator" feels awkward to me because we have so many situations where code is generated. Tokens are more short-lived.

pkg/config/configuration.go
```diff
@@ -0,0 +1,186 @@
...
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+	"gopkg.in/yaml.v3"
+)
+
+// File is an on-disk configuration of a Config.
+type Configuration struct {
+	Registry Registry `mapstructure:"registry"`
+	TokenCmd string   `mapstructure:"token_cmd" yaml:"token_cmd"` // run in shell to generate token
```
timburks, line 32: I think it would be helpful to name this consistently with the command used to set it.. "generator" or "source"

pkg/config/configuration.go
```diff
@@ -0,0 +1,186 @@
...
+	decoder, err := mapstructure.NewDecoder(config)
+	if err != nil {
+		return err
+	}
+
+	return decoder.Decode(m)
+}
+
+// Resolve derived values (eg. Registry.Token from Registry.TokenCmd)
+func (c *Configuration) Resolve() error {
```
timburks, line 119: How frequently is this run? 
1. before every token usage? 
2. once for each run of the registry tool?
3. when tokens expire?

cmd/registry/cmd/auth/token.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+
+	"github.com/apigee/registry/pkg/config"
+	"github.com/spf13/cobra"
+)
+
+func tokenCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "token",
+		Short: "Invoke the generator to print an auth token. See: `registry auth generator`.",
```
timburks, line 27: This command is to allow other tools to get tokens, right? If it's analogous to `gcloud auth print-*-token` could we name it `registry auth print-token` (for consistency)?

pkg/config/configuration.go
timburks: Seeking consistency with `gcloud`, how about `registry auth set-token-source` and `registry auth print-token`? 

Or could setting this be handed with `registry config set token-source`?

theganyo: Agreed. Once we decide on that.

pkg/config/configuration.go
theganyo: Currently, this would only happen on load - ie. #2.

cmd/registry/cmd/auth/token.go
theganyo: Sure.

theganyo:
> Is there a strong reason to prefer "generator" over other command names? My initial preference on seeing it was to suggest "provider", but when I looked for other precedents I found "source" in [the Go oauth2 package](https://pkg.go.dev/golang.org/x/oauth2@v0.0.0-20220808172628-8227340efae7#TokenSource), which I think has a stronger case to be made for it.

Nope. I think "token-source" would be fine.

> Could setting this be handed with `registry config set token-source`?

I was trying to keep with your idea of putting this auth stuff on "auth" instead of config. Is that not desirable?

timburks:
> I was trying to keep with your idea of putting this auth stuff on "auth" instead of config. Is that not desirable?

I guess I'm undecided about this - on one hand, it's good to have `auth token-source` because it's in the `auth` hierarchy. But I got the impression that the token source was just one more piece of configuration that would be stored alongside all the other values that are set with `config set`.

theganyo:
> I guess I'm undecided about this - on one hand, it's good to have `auth token-source` because it's in the `auth` hierarchy. But I got the impression that the token source was just one more piece of configuration that would be stored alongside all the other values that are set with `config set`.

Yeah, I don't have a strong opinion either way. We could do `auth config set token-source xxxx`... it's a bit more verbose, fwiw, but I'm really uncertain which is more intuitive.

timburks (approved):

Issue #666, opened 2022-07-20
Registry server: sqlite "database is locked" errors are reported as Internal


timburks:
While running some parallel wipeout tests with the sqlite backend, I get errors like this:
```
 FATAL[0021] Task failed: delete projects/hosted-ci-test/locations/global/apis/my-api/versions/v037 error=rpc error: code = Internal desc = database is locked
```
That's described in more detail [here](https://github.com/mattn/go-sqlite3/issues/274), and I think it would be appropriate to return a retriable error code in this case, such as [Unavailable](https://github.com/grpc/grpc-go/blob/64174955202ffb5ea4122e25d1aaece49cc5a3ed/codes/codes.go#L180).

timburks:
I think we should be checking [here](https://github.com/apigee/registry/blob/2c94ac7bf013279f5c9747c2b6bc37587181a189/server/registry/internal/storage/delete.go#L43) among other places in the same file.

seaneganx:
Some potentially relevant documentation.

http://sqlite.com/matrix/rescode.html#busy

>The SQLITE_BUSY result code differs from [SQLITE_LOCKED](http://sqlite.com/matrix/rescode.html#locked) in that SQLITE_BUSY indicates a conflict with a separate [database connection](http://sqlite.com/matrix/c3ref/sqlite3.html), probably in a separate process, whereas [SQLITE_LOCKED](http://sqlite.com/matrix/rescode.html#locked) indicates a conflict within the same [database connection](http://sqlite.com/matrix/c3ref/sqlite3.html) (or sometimes a database connection with a [shared cache](http://sqlite.com/matrix/sharedcache.html)).

With the wipeout command we make lots of concurrent requests to the server, but the server only uses one database client per request. That said, we would expect a conflict with a separate database connection (SQLITE_BUSY) instead of one within the same connection (SQLITE_LOCKED). https://github.com/mattn/go-sqlite3/issues/274 seems to be focused on the SQLITE_BUSY error too.

It's possible we're running into the following variant of this error.

http://sqlite.com/matrix/rescode.html#busy_snapshot

>The following scenario illustrates how an SQLITE_BUSY_SNAPSHOT error might arise:
>
>1. Process A starts a read transaction on the database and does one or more SELECT statement. Process A keeps the transaction open.
>2. Process B updates the database, changing values previous read by process A.
[...]

PR #682, opened 2022-08-12
Expand handling of database error codes

timburks:
Addresses #666 by adding a translation function that converts gorm and database-specific errors to errors with gRPC status codes; all functions in the `server/registry/internal/storage` package should now return gRPC-ready errors with appropriate status codes that allow retriable errors to be recognized and retried by clients.

This was verified with various DiRT-style tests run on a local `registry-server` with local databases (SQLite and PostgreSQL) using `registry upload bulk openapi` to load thousands of APIs concurrently (`--jobs 5000`) and using `registry count versions` running over all APIs with concurrency hard-coded to 1000 (this command currently does not accept the `--jobs` flag). This extremely high concurrency caused timeouts during bulk uploads, so this manual testing was done with a regenerated GAPIC that was configured to have 600-second timeouts.

This commit does not add CI testing of these error conditions.

Database errors that aren't explicitly handled are logged and treated as `Internal` errors (as before).


server/registry/internal/storage/errors.go
```diff
@@ -15,15 +15,78 @@
...
+		}
+	}
+	return false
+}
+
 // grpcErrorForDBError converts recognized database error codes to grpc error codes.
-func grpcErrorForDBError(err error) error {
-	if alreadyExists(err) {
-		return status.Error(codes.AlreadyExists, err.Error())
+func grpcErrorForDBError(ctx context.Context, err error) error {
```
theganyo, line 43: Would there be any benefit in logging the original error here?

theganyo: On further inspection, I think that only the `unique_violation` err would probably have any additional useful information. Is detail getting logged somewhere?

theganyo (approved):

server/registry/internal/storage/errors.go
```diff
@@ -15,15 +15,78 @@
...
+	}
+	// handle sqlite3 errors separately so their support can be conditionally compiled.
+	if err2 := grpcErrorForSQLite3Error(ctx, err); err2 != nil {
+		return err2
 	}
+	// handle all other known error types.
+	switch v := err.(type) {
+	case *pq.Error:
+		if v.Code.Name() == "unique_violation" {
+			return status.Error(codes.AlreadyExists, err.Error())
```
timburks, line 56: @theganyo This is copying the text of the original error message into the gRPC-style "Status" error that is being created (with `codes.AlreadyExists`). I think that adds the additional information that you were suggesting.

PR #683, opened 2022-08-16
add project, location config and allow short names

### LINKED ISSUE #677: Registry tool: add project and location config properties
From https://github.com/apigee/registry/issues/625#issue-1282776780:
> In many situations it would be useful for the registry tool to have a default project and location. For example:
registry list apis is much preferable to registry list projects/PROJECTID/locations/global/apis and the --parent argument to registry apply seem needlessly tedious. We might set these with new environment variables APG_REGISTRY_PROJECT and APG_REGISTRY_LOCATION, but now that the generated apg client is merged into the registry tool, it's easier for us to explore other options like using a configuration file to store this information.

theganyo:
This isn't as elegant as I'd like, but I didn't see a good integration point that could fix it as a cross-cutting concern. Reworking the naming package to require configuration didn't seem reasonable. One potential alternative to this rather brute-force approach would be to rework the configured clients to include naming lookups... however, right now they are direct pointers to the generated clients... so we'd have to wrap them in our own client structs instead and change the references. I can investigate that (or other ideas) if the solution in this PR isn't desireable.

Fixes #677

### DIFF
--- a/cmd/registry/cmd/util/util.go
+++ b/cmd/registry/cmd/util/util.go
@@ -16,11 +16,8 @@ package util
 
 import (
 	"fmt"
-	"path"
-	"strings"
 
 	"github.com/apigee/registry/pkg/config"
-	"github.com/apigee/registry/pkg/connection"
 )
 
 var NoActiveConfigurationError = fmt.Errorf(`No active configuration. Use 'registry config configurations' to manage.`)
@@ -32,18 +29,4 @@ func TargetConfiguration() (name string, c config.Configuration, err error) {
 	}
 
 	return
-}
-
-// FQName ensures the project and location, if available,
-// are properly included in the resource name.
-func FQName(c connection.Config, name string) string {
-	name = strings.TrimPrefix(name, "/")
-	if !strings.HasPrefix(name, "projects") && c.Project != "" {
-		if strings.HasPrefix(name, "locations") {
-			name = path.Join("projects", c.Project, name)
-		} else if c.Location != "" {
-			name = path.Join("projects", c.Project, "locations", c.Location, name)
-		}
-	}
-	return name
 }
--- a/cmd/registry/cmd/util/util.go
+++ b/cmd/registry/cmd/util/util.go
@@ -16,8 +16,11 @@ package util
 
 import (
 	"fmt"
+	"path"
+	"strings"
 
 	"github.com/apigee/registry/pkg/config"
+	"github.com/apigee/registry/pkg/connection"
 )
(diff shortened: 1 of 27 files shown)


cmd/registry/cmd/util/util.go
```diff
@@ -30,3 +33,17 @@ func TargetConfiguration() (name string, c config.Configuration, err error) {
 
 	return
 }
+
+// FQName ensures the project and location, if available,
+// are properly included in the resource name.
+func FQName(c connection.Config, name string) string {
```
timburks, line 39: Could we make this a member function of connection.Config? I think the `util` package might attract some [negative attention](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_avoid_package_names_like_base_common_or_util) and there's not very much in it -- I think `NoActiveConfigurationError` also makes sense as `connection.NoActiveConfigurationError`

cmd/registry/cmd/util/util.go
```diff
@@ -30,3 +33,17 @@ func TargetConfiguration() (name string, c config.Configuration, err error) {
 
 	return
 }
+
+// FQName ensures the project and location, if available,
+// are properly included in the resource name.
```
timburks, line 38: I think it would be good to spell out "fully qualified" somewhere in the comment.

timburks (approved):
Just a couple of minor nonblocking comments. Thanks!

timburks:
I think the `registry rpc` commands are notably low-level, so I think we can live with requiring full resource names in them. The improvements that this makes to `registry get` and `registry list` are really nice, and might motivate us to add/improve other handwritten commands (like `registry delete`) that make the `rpc` commands less necessary.

cmd/registry/cmd/util/util.go
theganyo: What's funny is that I did have this as a member of connnection.Config in an earlier iteration and moved it out. I don't actually remember why exactly, other than it seems like maybe I thought it was more closely tied to cmd stuff. Whatever. I'll move it back. :)

cmd/registry/cmd/util/util.go
theganyo: Will do

cmd/registry/cmd/util/util.go
theganyo: Oh, as for moving NoActiveConfigurationError error (and, I assume, the TargetConfiguration function?), that seems out of scope. I'll add an issue.

PR #694, opened 2022-08-23
improve create entity validation and efficiency

### LINKED ISSUE #691: Registry server: improve create entity validation and efficiency
[This code](https://github.com/apigee/registry/blob/main/server/registry/actions_artifacts.go#L69) does record validation after a query to see if the record exists - it would be better to reverse them and avoid the query in the error case. Better yet, I believe we can avoid the duplicate check entirely by using a DB "create" instead of "save" [here](https://github.com/apigee/registry/blob/main/server/registry/actions_artifacts.go#L107).

Check other entities as this may also apply to entities other than artifacts.

theganyo:
Fixes #691

### DIFF
--- a/server/registry/actions_artifacts_test.go
+++ b/server/registry/actions_artifacts_test.go
@@ -373,7 +373,7 @@ func TestCreateArtifactDuplicates(t *testing.T) {
 				ArtifactId: "My-Artifact",
 				Artifact:   &rpc.Artifact{},
 			},
-			want: codes.AlreadyExists,
+			want: codes.InvalidArgument,
 		},
 	}
 
--- a/server/registry/actions_artifacts_test.go
+++ b/server/registry/actions_artifacts_test.go
@@ -365,16 +365,6 @@ func TestCreateArtifactDuplicates(t *testing.T) {
 			},
 			want: codes.AlreadyExists,
 		},
-		{
-			desc: "case insensitive",
-			seed: &rpc.Artifact{Name: "projects/my-project/locations/global/artifacts/my-artifact"},
-			req: &rpc.CreateArtifactRequest{
-				Parent:     "projects/my-project/locations/global",
-				ArtifactId: "My-Artifact",
-				Artifact:   &rpc.Artifact{},
-			},
-			want: codes.InvalidArgument,
-		},
 	}
 
 	for _, test := range tests {
--- a/server/registry/actions_deployments.go
+++ b/server/registry/actions_deployments.go
@@ -49,14 +49,15 @@ func (s *RegistryServer) createDeployment(ctx context.Context, name names.Deploy
 		return nil, status.Error(codes.Unavailable, err.Error())
 	}
 
+	if err := name.Validate(); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	// SDG: can we get the create to do this check?
 	if _, err := db.GetDeployment(ctx, name); err == nil {
 		return nil, status.Errorf(codes.AlreadyExists, "API deployment %q already exists", name)
 	} else if !isNotFound(err) {
 		return nil, err
-	}
-
-	if err := name.Validate(); err != nil {
-		return nil, status.Error(codes.InvalidArgument, err.Error())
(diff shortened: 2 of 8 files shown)


server/registry/internal/storage/save.go
```diff
@@ -74,20 +74,9 @@ func (c *Client) SaveArtifactContents(ctx context.Context, artifact *models.Arti
...
-				return err
-			}
-		}
-
-		return nil
-	})
+	// update if exists or create as needed
+	err := c.db.WithContext(ctx).Clauses(clause.OnConflict{
+		UpdateAll: true,
+	}).Create(v).Error
```
timburks, line 80: This looks like a small-but-significant change because it seems to be introducing a new way of performing this operation in gorm -- and it's removing code that admittedly seems weird. Could we link to gorm docs that explain this, preferably in a distinct PR?

theganyo: That'll teach me to do cleanup as I go! LOL. Sure, I'll do that.

server/registry/actions_specs_test.go
```diff
@@ -606,7 +606,7 @@ func TestGetApiSpecContentsData(t *testing.T) {
 			if status.Code(err) != test.wantCode {
 				t.Errorf("GetApiSpecContents(%+v) returned status code %q, want %q: %v", test.req, status.Code(err), test.wantCode, err)
 			}
-			if bytes.Compare(resp.Data, test.wantData) != 0 {
+			if !bytes.Equal(resp.Data, test.wantData) {
```
timburks, line 609: This would be easier to discuss as a standalone PR - here I think it adds noise to a PR that already takes some effort to understand, and the questions about this ("is this a styleguide-driven change, a personal preference, or fixing a functional problem?") seem better to discuss separately.

server/registry/internal/storage/save.go
timburks: Thanks! Sorry to be picky - my stack overflows easily :)

server/registry/actions_specs_test.go
theganyo: This was a lint fix. Yeah, sorry, I'll just stop doing this stuff.

server/registry/actions_artifacts_test.go
```diff
@@ -373,7 +373,7 @@ func TestCreateArtifactDuplicates(t *testing.T) {
 				ArtifactId: "My-Artifact",
 				Artifact:   &rpc.Artifact{},
 			},
-			want: codes.AlreadyExists,
+			want: codes.InvalidArgument,
```
timburks, line 376: Here I guess that we're getting the InvalidArgument result now because the validation check is happening first. I had a similar thing pop up in a different PR (https://github.com/apigee/registry/pull/615#discussion_r902911965), and there @seaneganx asked me to just remove the test because validation was already being tested.

theganyo: sure

server/registry/actions_deployments.go
```diff
@@ -49,16 +49,17 @@ func (s *RegistryServer) createDeployment(ctx context.Context, name names.Deploy
 		return nil, status.Error(codes.Unavailable, err.Error())
 	}
 
+	if err := name.Validate(); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	// SDG: can we get the create to do this check?
```
timburks, line 56: nit: I don't think we have any other personal identifiers (initials) in the code - "git blame" makes attribution easy enough :) Also, this comment is in essentially the same place as the one that's added to line 60 of actions_specs.go. Could these be identical? Just trying to reduce the burden on future readers of these two files, which should have a high level of similarity. Maybe someday someone will merge them with generics :)

theganyo: Oops. That was just an oversight - I had meant to review that, not push it. (I had stopped using TODO because there were too many to easily track in the codebase.)

theganyo:
Ok, I think that's everything.

timburks (approved):
Thanks!

PR #695, opened 2022-08-24
 clone active w/ flags on config create

### LINKED ISSUE #692: Registry tool: Use global flags for config create
If a user specifies any of flags that set a stored value during `config create`, the created config should save them. Eg.

      --registry.address string    the server and port of the registry api (eg. localhost:8080)
      --registry.location string   the registry location
      --registry.project string    the registry project

theganyo:
Clones current active configuration using any flag overrides when creating a new configuration.

Fixes #692

### DIFF
--- a/cmd/registry/cmd/config/configurations/create.go
+++ b/cmd/registry/cmd/config/configurations/create.go
@@ -38,8 +38,15 @@ func createCommand() *cobra.Command {
 				return fmt.Errorf("Cannot create config %q, it already exists.", name)
 			}
 
-			s := config.Configuration{}
-			err := s.Write(name)
+			// attempt to clone the active config or at least the current flags
+			s, err := config.Active()
+			if err != nil {
+				if s, err = config.ReadValid(""); err != nil {
+					s = config.Configuration{}
+				}
+			}
+
+			err = s.Write(name)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/registry/cmd/config/configurations/create.go
```diff
@@ -38,8 +38,15 @@ func createCommand() *cobra.Command {
 				return fmt.Errorf("Cannot create config %q, it already exists.", name)
 			}
 
-			s := config.Configuration{}
-			err := s.Write(name)
+			// attempt to clone the active config or at least the current flags
+			s, err := config.Active()
```
timburks, line 42: I struggled to understand this line, maybe because "Active" is an adjective and not a noun. That's the active configuration, right?

timburks: Where are the flag values (like "registry.project") being saved?

theganyo: That is the active configuration, yes... making that `config.ActiveConfiguration()` would seem like a stutter to me. The flag values are automatically populated while reading a configuration using flag bindings. See: https://github.com/theganyo/registry/blob/main/pkg/config/config.go#L174

timburks (approved):

PR #696, opened 2022-08-24
Add documentation to config and auth commands

### LINKED ISSUE #693: Registry tool: `config` commands need documentation
For example:

`registry config` should explain the use of configurations.

`registry config set PROPERTY VALUE` should display the names of valid config values.

`registry auth` should explain how the token-source works and is set.

theganyo:
Fixes #693

### DIFF
--- a/cmd/registry/cmd/auth/auth.go
+++ b/cmd/registry/cmd/auth/auth.go
@@ -24,7 +24,14 @@ import (
 func Command() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "auth",
-		Short: "Manage registry CLI authentication.",
+		Short: "Manage authentication to the registry.",
+		Long: "Authentication to the registry is via tokens. These tokens can be provided " +
+			"directly by using '--registry.token' in any command, or can be generated as needed " +
+			"by setting the token-source property in a configuration. The token-source can be set " +
+			"to any executable command that prints a registry token by using the 'config set token-source'. " +
+			"Once set, the command will be executed before a registry command is run and the output used " +
+			"as the token passed to the registry. The output can be verified by 'auth print-token'.",
+		Example: "config set token-source 'gcloud auth print-access-token email@example.com'",
 	}
 
 	cmd.AddCommand(printTokenCommand())
--- a/cmd/registry/cmd/auth/print_token.go
+++ b/cmd/registry/cmd/auth/print_token.go
@@ -24,7 +24,7 @@ import (
 func printTokenCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "print-token",
-		Short: "Invoke token source and print token.",
+		Short: "Invoke token-source to print a token.",
 		Args:  cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
--- a/cmd/registry/cmd/config/set.go
+++ b/cmd/registry/cmd/config/set.go
@@ -24,8 +24,16 @@ import (
 func setCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "set PROPERTY VALUE",
-		Short: "Set the value of a property.",
-		Args:  cobra.ExactArgs(2),
+		Short: "Set a property value in the active configuration.",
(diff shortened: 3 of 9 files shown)


cmd/registry/cmd/auth/print_token.go
```diff
@@ -24,7 +24,7 @@ import (
 func printTokenCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "print-token",
-		Short: "Invoke token source and print token.",
+		Short: "Invoke token-source to print a token.",
```
timburks, line 27: What happens if no token source is set? Would/should it default to printing a manually-set registry.token?

timburks: (Is `registry.token` going away? I notice it's not displayed by `registry config list` I'm not sure that it's actually getting set when I run `registry config set registry.token ... ` the `get` command isn't returning a value)

timburks (approved):

cmd/registry/cmd/auth/print_token.go
theganyo: It's not possible to manually set one in a configuration currently. The user would simply use the --registry.token flag.

cmd/registry/cmd/auth/auth.go
```diff
@@ -24,7 +24,14 @@ import (
...
-		Short: "Manage registry CLI authentication.",
+		Short: "Manage authentication to the registry.",
+		Long: "Authentication to the registry is via tokens. These tokens can be provided " +
+			"directly by using '--registry.token' in any command, or can be generated as needed " +
+			"by setting the token-source property in a configuration. The token-source can be set " +
+			"to any executable command that prints a registry token by using the 'config set token-source'. " +
+			"Once set, the command will be executed before a registry command is run and the output used " +
+			"as the token passed to the registry. The output can be verified by 'auth print-token'.",
+		Example: "config set token-source 'gcloud auth print-access-token email@example.com'",
 	}
```
giteshk, line 35: Just a suggestion :  Should we be looking at the expiration of a token before re-generating it? We are adding an additional step before every command is executed.

cmd/registry/cmd/config/set.go
```diff
@@ -24,8 +24,16 @@ import (
...
+		Long: "Set a property value in the active configuration. The following are valid properties:\n" +
+			"  - registry.address\n" +
+			"  - registry.insecure\n" +
+			"  - registry.location\n" +
+			"  - registry.project\n" +
+			"  - token-source",
+		Example: "registry config set registry.address localhost:8080\n" +
+			"registry config set token-source 'gcloud auth print-access-token email@example.com'",
+		Args: cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
```
giteshk, line 37: Does setting the registry.location and registry.project affect any commands?

Does it make `registry list project/project1/locations/global/apis` command shorter to `registry list apis` ?

theganyo: Yes, that's exactly what it does.

giteshk: I am going to give this a try.. Would be very useful. Does the `registry config configurations list` print these settings?

cmd/registry/cmd/auth/auth.go
theganyo: A token is currently considered opaque and isn't stored on disk. Unsure if it's a net gain to break from that?

cmd/registry/cmd/config/set.go
timburks: Yes, when location and project are set, the "project/PROJECT_ID/location/LOCATION_ID" prefix gets added when needed.

giteshk (approved):

cmd/registry/cmd/auth/auth.go
timburks: We're not inspecting the token now and probably wouldn't be able to parse all tokens that we might get. But we could consider doing this when it's possible.

cmd/registry/cmd/config/set.go
theganyo: No, `registry config configurations list` is meant to be a simple table. You can see all properties using `registry config list` (for the active configuration) or `registry config configurations describe NAME` (for a specific configuration).

PR #698, opened 2022-08-24
disallow delete of only SpecRevision

### LINKED ISSUE #257: Registry API: DeleteApiSpecRevision can delete the only remaining revision of a spec

theganyo:
Fixes #257

### DIFF
--- a/server/registry/actions_spec_revisions.go
+++ b/server/registry/actions_spec_revisions.go
@@ -93,10 +93,7 @@ func (s *RegistryServer) DeleteApiSpecRevision(ctx context.Context, req *rpc.Del
 	}
 
 	if err := db.DeleteSpecRevision(ctx, name); err != nil {
-		if _, ok := status.FromError(err); ok {
-			return nil, err
-		}
-		return nil, status.Error(codes.Internal, err.Error())
+		return nil, err
 	}
 
 	s.notify(ctx, rpc.Notification_DELETED, name.String())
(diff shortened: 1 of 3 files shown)


server/registry/actions_spec_revisions.go
```diff
@@ -93,6 +93,9 @@ func (s *RegistryServer) DeleteApiSpecRevision(ctx context.Context, req *rpc.Del
 	}
 
 	if err := db.DeleteSpecRevision(ctx, name); err != nil {
+		if _, ok := status.FromError(err); ok {
```
timburks, line 96: Could we replace this (and line 99) with a call to grpcErrorFromDBError() ?

theganyo: No, that's a private function to the storage package. But good point: that has already been done by storage, so this doesn't need to translate the error anyway.

timburks (approved):

PR #699, opened 2022-08-25
Registry tool: Use `registry config` instead of env vars for scripts

### LINKED ISSUE #676: Registry tool: Use `registry config` instead of env vars for scripts
Per https://github.com/apigee/registry/issues/625#issue-1282776780:

> This would probably replace the scripts in the auth directory with direct registry subcommands. auth/HOSTED.sh might become registry config gcloud and auth/LOCAL.sh might be registry config local (with defaults and flags for pointing to a locally-running server). 

Update client-side, customer-facing scripts and/or docs to use `registry config` instead of env vars.

theganyo:
Fixes #676


auth/HOSTED.sh
```diff
@@ -16,22 +16,33 @@
...
+CLIENT_EMAIL="$(gcloud config get account)"
+APG_REGISTRY_TOKEN_SOURCE="gcloud auth print-access-token ${APG_REGISTRY_CLIENT_EMAIL}"
 
-# The auth token is generated for the gcloud logged-in user.
-export APG_REGISTRY_CLIENT_EMAIL=$(gcloud config list account --format "value(core.account)")
-export APG_REGISTRY_TOKEN=$(gcloud auth print-access-token ${APG_REGISTRY_CLIENT_EMAIL})
+registry config configurations create hosted \
+  --registry.insecure="${APG_REGISTRY_INSECURE}" \
+  --registry.address="${APG_REGISTRY_ADDRESS}" \
+  --registry.project="${APG_REGISTRY_PROJECT}" \
```
timburks, line 45: :+1:

auth/LOCAL.sh
```diff
@@ -37,10 +37,17 @@ fi
...
+APG_REGISTRY_INSECURE="true"
 
-# Local calls don't need authentication.
-unset APG_REGISTRY_TOKEN
+APG_REGISTRY_PROJECT="${REGISTRY_PROJECT_IDENTIFIER}"
+APG_REGISTRY_LOCATION="global"
+CLIENT_EMAIL=$(gcloud config list account --format "value(core.account)")
+
+registry config configurations create local \
+  --registry.insecure="${APG_REGISTRY_INSECURE}" \
```
timburks, line 50: does this have to be "true" or "false" now?

theganyo: No. Can still be 0 or 1 if you like.

auth/README.md
```diff
@@ -1,19 +1,31 @@
...
+- [LOCAL.sh](LOCAL.sh) creates a configuration ("local") that configures the registry
+  client to work with a locally-running server.
+- [HOSTED.sh](HOSTED.sh) creates a configuration ("hosted") that configures the registry
+  client to work with a Google Cloud hosted server.
+
+Each script should only be run once to create and activate a configuration. Once
+created, the configurations can be selected be using the registry commands. For
+example:
+
+$ registry config configurations activate local
```
timburks, line 16: So... we're adding an extra step. That means that just using the scripts as-before won't work. WDYT about adding an activate step at the end of each script?

auth/LOCAL.sh
theganyo: Nope. Can still use 0 and 1 if you prefer.

auth/README.md
theganyo: No, they will work. When you create a new profile, it activates it.

timburks (approved):
This is great - we've also discussed deleting the other scripts and leaving HOSTED.bat for @giteshk to help us fix, all fine in this PR.

auth/README.md
timburks: ok I see now "create and activate" - LGTM

timburks (approved):

PR #700, opened 2022-08-25
Registry server: wrap mutating actions in transactions

timburks:
I've been testing with remote databases and very high parallelism, logging and handling errors that arose from that, and saw some issues related to our lack of transaction support.

This PR wraps all mutating operations in action handlers in transactions (stage 1 of #706). The transaction wrapper also translates errors into gRPC errors, which with other proposed changes would allow us remove the majority of `grpcErrorForDBError` calls in the `storage` package.

Since delete actions are wrapped in transactions, transactions are removed from the deletion support code in the `storage` package (stage 2 of #706).

Stages 3-5 of #706 are left for future PRs.


log/interceptor/interceptor.go
```diff
@@ -82,14 +82,38 @@ func CallLogger(opts ...log.Option) grpc.UnaryServerInterceptor {
 		switch status.Code(err) {
 		case codes.OK:
 			logger.Info("Success.")
-		case codes.Internal:
-			logger.Error("Internal error.")
+		case codes.Canceled:
```
timburks, line 85: This adds support for all gRPC status codes so that logging won't report "User error" for the unrecognized ones. It's in this commit because I used it for testing but I'll make a separate PR for just this change to hopefully commit before this one.

timburks:
@seaneganx if we merge this before you're back, PTAL anyway, comment on anything, and we'll address it in issues/followup PRs.

server/registry/actions_artifacts.go
```diff
@@ -51,88 +51,83 @@ func parseArtifactParent(name string) (artifactParent, error) {
...
-		if _, err := db.GetProject(ctx, parent); err != nil {
-			return nil, err
-		}
-	case names.Api:
-		if _, err := db.GetApi(ctx, parent); err != nil {
-			return nil, err
+	var response *rpc.Artifact
+	if err := s.runWithTransaction(ctx, func(ctx context.Context, db *storage.Client) error {
+		// Creation should only succeed when the parent exists.
+		switch parent := parent.(type) {
```
theganyo, line 71: nit: Probably subjective, but might this be easier to read like this?

		var err error
		switch parent := parent.(type) {
		case names.Project:
			_, err = db.GetProject(ctx, parent)
		case names.Api:
			_, err = db.GetApi(ctx, parent)
		case names.Version:
			_, err = db.GetVersion(ctx, parent)
		case names.Spec:
			_, err = db.GetSpec(ctx, parent)
		case names.Deployment:
			_, err = db.GetDeployment(ctx, parent)
		}
		if err != nil {
			return err
		}

server/registry/server.go
```diff
@@ -95,6 +95,14 @@ func (s *RegistryServer) getStorageClient(ctx context.Context) (*storage.Client,
 	return s.storageClient, nil
 }
 
+func (s *RegistryServer) runWithTransaction(ctx context.Context, fn func(ctx context.Context, db *storage.Client) error) error {
```
theganyo, line 98: nit: Personally, I would prefer `runInTransaction()` or `asTransaction()`.

theganyo (approved):

server/registry/server.go
timburks: Thanks, I'll change this to "runInTransaction"

server/registry/actions_artifacts.go
timburks: Agreed, thanks.

server/registry/server.go
timburks: Done

server/registry/actions_artifacts.go
timburks: Done

theganyo:
That last commit (https://github.com/apigee/registry/pull/700/commits/a1ece2a127f50e5077890578759a38983c0e0cc5) adds code with supporting tests. Is it reasonable to test?

timburks:
@theganyo let's discuss that -- I was just looking into retry configuration before adding a comment about that change. We can trigger the concurrency problem by running the bulk uploaders, e.g. `registry upload bulk discovery`, but I think this is too big to put in our CI flow. It might be better to directly construct a test by forcing a set of concurrent updates -- maybe starting several goroutines that all update the same resource at once? I think that's out of scope for this PR and would rather 1) leave this change in and file an issue for testing or 2) pull the last commit and address it in a different PR.

The change in that commit is observable [here](https://github.com/apigee/registry/blob/e9c6f8cd0a99a6bb699bc6abe773a235e77cfd8e/cmd/registry/cmd/upload/bulk/discovery.go#L130) which explicitly ignores the (weird) `AlreadyExists` errors that concurrent updates can trigger. With the change, those errors are reported as `Aborted` and are retried by the client libraries (because we [configured this](https://github.com/apigee/registry/blob/e9c6f8cd0a99a6bb699bc6abe773a235e77cfd8e/gapic/grpc_service_config.json#L15)).

theganyo:
I've been worried that we don't have a way of reliably triggering the concurrency / transactional conditions that we've been running into - and thus we'd be unable to verify that we've eliminated the issues. That said, if we can reliably trigger our concurrency issues through the bulk uploader and we don't have the means to more specifically test at this point, perhaps that's good enough for now.

server/registry/actions_specs.go
```diff
@@ -254,71 +246,53 @@ func (s *RegistryServer) ListApiSpecs(ctx context.Context, req *rpc.ListApiSpecs
...
-		before := time.Now()
-		// Prevent a race condition that can occur when two updates are made
-		// to the same non-existent resource. The db.Get...() call returns
-		// NotFound for both updates, and after one creates the resource,
-		// the other creation fails. The lock() prevents this by serializing
-		// the get and create operations. Future updates could improve this
-		// with improvements closer to the database level.
-		updateSpecMutex.Lock()
-		defer updateSpecMutex.Unlock()
-		log.Debugf(ctx, "Acquired lock after blocking for %v", time.Since(before))
```
timburks, line 287: Removing this here, but will address the underlying problem in https://github.com/apigee/registry/pull/715

PR #710, opened 2022-08-30
Support max-actions flag in resolve

shrutiparabgoogle:
The default value is `100` for now, it is subject to change once we do some testing while deploying it to the data plane.

### DIFF
--- a/cmd/registry/cmd/resolve/resolve.go
+++ b/cmd/registry/cmd/resolve/resolve.go
@@ -55,6 +55,7 @@ func fetchManifest(
 func Command() *cobra.Command {
 	var dryRun bool
 	var jobs int
+	var maxActions int
 	cmd := &cobra.Command{
 		Use:   "resolve MANIFEST_RESOURCE",
 		Short: "resolve the dependencies and update the registry state (experimental)",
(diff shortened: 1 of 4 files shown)


cmd/registry/cmd/resolve/resolve.go
```diff
@@ -120,5 +121,6 @@ func Command() *cobra.Command {
 
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "if set, actions will only be printed and not executed")
 	cmd.Flags().IntVarP(&jobs, "jobs", "j", 10, "Number of actions to execute simultaneously")
+	cmd.Flags().IntVarP(&maxActions, "max-actions", "a", 100, "Maximum number of actions to execute")
```
timburks, line 124: How does this differ from `--jobs`?

shrutiparabgoogle: `--jobs` limits the parallelism, the number of actions the controller can execute in parallel 
`--max-actions` is a cap on the total number of actions, as a whole the controller will execute only `max-actions` number of actions in one run. This to limit the controller from overloading the registry with large number of requests.

timburks (approved):

Issue #701, opened 2022-08-25
update auth/HOSTED.bat to create a configuration


theganyo:
The auth/HOSTED.bat file needs to be updated to create a registry configuration instead of env vars like auth/HOSTED.sh now does.

See also https://github.com/apigee/registry/pull/699#pullrequestreview-1086160695

PR #712, opened 2022-08-31
#701 Update windows auth file to use registry configurations

giteshk:
Ticket #701 This has been tested using https://github.com/giteshk-org/registry/releases/tag/v0.3.21


timburks (approved):

theganyo (commented):
Should we be using SETLOCAL to avoid leaking the environment variables to the terminal?

giteshk:
@theganyo good catch will add a PR for this change

PR #715, opened 2022-08-31
Registry server: report AlreadyExists errors in updates as Abort to cause them to be retried

timburks:
As discussed [here](https://github.com/apigee/registry/pull/700#issuecomment-1233376891), #700 removed a mutex with this associated comment:
```
// Prevent a race condition that can occur when two updates are made
// to the same non-existent resource. The db.Get...() call returns
// NotFound for both updates, and after one creates the resource,
// the other creation fails. The lock() prevents this by serializing
// the get and create operations. Future updates could improve this
// with improvements closer to the database level.
```

This commit addresses the underlying problem by detecting when it occurs and returning the gRPC `Aborted` status code, which client libraries are configured to retry. This client retry avoids slowing the common case while still allowing for correct handling.

The effect of this is observable in the bulk uploaders, e.g. [here](https://github.com/apigee/registry/blob/e9c6f8cd0a99a6bb699bc6abe773a235e77cfd8e/cmd/registry/cmd/upload/bulk/discovery.go#L130) which explicitly ignores the `AlreadyExists` errors that can be returned when two callers concurrently try to create the same API using `UpdateApi` calls with `allow missing` set to true.

With the change, those errors are reported as Aborted and are retried by the client libraries (because we [configured this](https://github.com/apigee/registry/blob/e9c6f8cd0a99a6bb699bc6abe773a235e77cfd8e/gapic/grpc_service_config.json#L15)). The modified bulk uploader code that explicitly ignores `AlreadyExists` is not reached.

How can we test this? It is observable by running `registry upload bulk discovery` with a high `--jobs` count (e.g. 50). However this has an external service dependency (it calls the API discovery service), and it seems overly-slow. A more direct test might be to run a large number of goroutines (maybe 10?) that concurrently try to create the same API (or version, spec, deployment, or artifact) using `Update` with `allow missing`. These should return `Aborted` or no error.

Let's consider making concurrency tests a separate group of tests that are run outside of the core server suite (perhaps like benchmarks). This would reduce the load of functional verification (as done by the core) and would increase focus on the kinds of concurrency situations that we are checking.


server/registry/concurrency_test.go
```diff
@@ -0,0 +1,69 @@
...
+				time.Sleep(time.Duration(rand.Intn(100)) * time.Microsecond)
+				_, err := server.UpdateApi(ctx, test.req)
+				code := status.Code(err)
+				switch code {
+				case codes.OK:
+				case codes.Unavailable:
+				default:
+					t.Errorf("UpdateApi(%+v) returned status code %q", test.req, status.Code(err))
+				}
+				wg.Done()
```
timburks, line 63: Here's a sample concurrency test that makes 10 UpdateApi calls concurrently. But instead of triggering the `AlreadyExists` or `Aborted` errors that we want, we get `Unavailable` because the database is locked.

theganyo:
A possibility for skipping long-running tests would be to simply use the `go test -short` flag. See:  https://pkg.go.dev/cmd/go#hdr-Testing_flags and https://pkg.go.dev/testing#hdr-Skipping.

server/registry/concurrency_test.go
theganyo: So... just to be clear: this test failed before this change was made?

theganyo:
The build is reporting that none of the new code is is covered by tests.

theganyo (approved):
Ok. It doesn't seem to be an issue with this PR directly, then, but I am concerned about the SQLite `Unavailable` failures as a separate issue.

timburks:
@theganyo This might allow us to block reads during Postgres transactions - https://gorm.io/docs/advanced_query.html#Locking-FOR-UPDATE - that would eliminate the race condition being addressed here. (Ideally this would be a row lock and not a table lock, right now I'm not sure which it is.)

theganyo:
The original issue of "when two updates are made to the same non-existent resource" seems like it would be a quite rare case. Is it worth trying to do more around that?

timburks:
> Is it worth trying to do more around that?

@theganyo Good question. Let's say no for now and discuss with @seaneganx later

PR #716, opened 2022-09-01
#711 ensure environment variables are scoped to the script

giteshk:


timburks (approved):

theganyo (commented):
endlocal isn't technically needed, but LGMT!

Issue #448, opened 2022-01-20
Registry API: Transactional database operations


seaneganx:
### Background

Most of our API methods perform several database operations. For example, the `UpdateApi` method reads the API first to make sure it exists and to load its current state, then applies any necessary changes in memory before writing back the API's updated state. Deletions have already been updated to use transactions in #434. 

Currently most of our API methods execute their necessary database operations using multiple transactions. This causes race conditions as discussed in #382, although using transactions may not solve other race conditions that might exist in the registry tool.

### Ideas

I think the ideal approach to this is to modify the interface exported by the `storage` package so each API method only interacts with the database once per request. Each interaction would be completed in one transaction.

Any other ideas I can think of seem like they would be messy and/or difficult to maintain.

timburks:
This is largely addressed by #706. However, because the transactions group write operations only, it is still possible to have races that occur between reads and updates. In #715, [we discussed the possibility of tightening the transaction locks](https://github.com/apigee/registry/pull/715#issuecomment-1234494283). We might also rewrite our update methods to only modify the fields specified in the request, eliminating the need for the initial get.

timburks:
See #720 for a stronger approach that does not rely on retry and that also prevents the read-modify-write races that could occur when updating existing resources. It uses full-table locks, which are probably only necessary for the `allow_missing` case (when no row initially exists), and so if performance seems to be a problem, we might refine this with single-row locks for rows that already exist.

theganyo:
I think this can probably be closed with https://github.com/apigee/registry/pull/760.

PR #720, opened 2022-09-01
Lock the table of affected entities during update operations.

timburks:
This is an alternate (and somewhat brute-force) approach to addressing race conditions mentioned in #448 and partially addressed in #715. It locks entire tables during update operations -- the full-table lock is needed because the update action first checks to see if a row exists, and if no row is present, there's no row to lock to prevent other processes or threads from trying to create it. The table locks are not supported by SQLite (it errors, so the locks are excluded for SQLite), and they are redundant because SQLite already locks tables during transactions.

Locks are automatically released when transactions end.

https://stackoverflow.com/a/72680220
https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-TABLES

### DIFF
--- a/server/registry/internal/storage/lock.go
+++ b/server/registry/internal/storage/lock.go
@@ -1,0 +1,53 @@
+// Copyright 2022 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import (
+	"context"
+	"fmt"
+
+	_ "github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/postgres"
+)
+
+func (c *Client) lockTable(ctx context.Context, name string) *Client {
+	if c.db.WithContext(ctx).Name() != "postgres" {
+		return c
+	}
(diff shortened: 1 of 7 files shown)


server/registry/internal/storage/lock.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+
+	_ "github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/postgres"
+)
+
+func (c *Client) lockTable(ctx context.Context, name string) *Client {
+	if c.db.WithContext(ctx).Name() != "postgres" {
+		return c
+	}
+	return &Client{db: c.db.Exec(fmt.Sprintf("LOCK TABLE %s IN ACCESS EXCLUSIVE MODE", name))}
```
timburks, line 28: At a lower lock level (`ROW EXCLUSIVE`), the concurrency tests fail.

timburks:
Runtimes before and after the change appear to be similar. I checked by running `registry upload bulk discovery --jobs 100 --project-id MYPROJECT` with a local server configured to use a remote cloudsql database. 

With the change, uploads completed in 1m 14s and 1m 20s, and prior to the change, uploads completed in 1m 20s and 1m22s. I don't think we can say this makes anything faster, but it doesn't seem to hurt.

timburks:
One more measurement `registry upload bulk openapi ~/Desktop/openapi-directory/APIs --project-id openapi --jobs 5000` completed in 1m 15s (with some tuning of the gapic retry parameters -- initial backoff of 5s, timeout at 600s). That was on the updated version, again running locally with a remote CloudSQL database. This seems good to go, performance-wise.

theganyo:
Single-threading updates will surely be painful at some point. I would expect the performance to drag with large databases, high utilization, and distributed operation. It's my belief that improved queries and use of relational behavior would be the proper solution - which is one of the reasons why I've been investigating those aspects. That said, it could be that the operation and scale at which this matters isn't within our target?

timburks:
@theganyo that's valid; all we can say so far is that this works for concurrent updates on the most strenuous example that I could run (bulk uploads). We might find that  more is needed when we look closely at controller operation; finding reproducible problems would help us guide and evaluate future steps.

theganyo (approved):
Ok. I don't have a strong argument against doing this for now.

PR #721, opened 2022-09-02
deprecation warning for client env var use

### LINKED ISSUE #702: Registry tool: emit deprecation warnings when env vars are used
Now that APG_REGISTRY_* env vars are unnecessary and configurations are preferred, we should emit warnings when the CLI picks them up.

theganyo:
Example:

```
$ registry get something
WARN: deprecated env: APG_REGISTRY_ADDRESS="localhost:8080"
WARN: deprecated env: APG_REGISTRY_INSECURE="1"
 DEBUG[0000] Unsupported entity [projects/theganyo-1/locations/global/something] uid=6965bb68
```

Note: Also fixes commands where we were unnecessarily loading config twice causing duplicate warnings.

Fixes #702

### DIFF
--- a/cmd/registry/cmd/annotate/annotate.go
+++ b/cmd/registry/cmd/annotate/annotate.go
@@ -47,7 +47,7 @@ func Command() *cobra.Command {
 			}
 			args[0] = c.FQName(args[0])
 
-			client, err := connection.NewRegistryClient(ctx)
+			client, err := connection.NewRegistryClientWithSettings(ctx, c)
 			if err != nil {
 				log.FromContext(ctx).WithError(err).Fatal("Failed to get client")
 			}
--- a/cmd/registry/cmd/compute/complexity.go
+++ b/cmd/registry/cmd/compute/complexity.go
@@ -54,7 +54,7 @@ func complexityCommand() *cobra.Command {
 				log.FromContext(ctx).WithError(err).Fatal("Failed to get dry-run from flags")
 			}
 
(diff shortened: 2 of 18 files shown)


cmd/registry/cmd/annotate/annotate.go
```diff
@@ -47,7 +47,7 @@ func Command() *cobra.Command {
 			}
 			args[0] = c.FQName(args[0])
 
-			client, err := connection.NewRegistryClient(ctx)
+			client, err := connection.NewRegistryClientWithSettings(ctx, c)
```
timburks, line 50: Just curious: since `c` is just `connection.ActiveConfig()` (essentially available to anything with access to the `connection` package), why did this function signature need to change? (couldn't `connection.NewRegistryClient()` just use the active config as its default?)

timburks (approved):

cmd/registry/cmd/annotate/annotate.go
theganyo: Yes, it can. The issue is that we're loading it twice. I could, instead, cache it in a global once loaded and simply retrieve it from that. What do you think?

timburks: That sounds good, but optional -- I don't think we need to hold this PR for that.

theganyo: Ok, thanks.

PR #725, opened 2022-09-03
Set limits on the number and lifetime of connections.

timburks:
This applies fairly tight restrictions to database connection usage. When all connections are in use, the server blocks until one becomes available. This can cause operations to take longer when concurrency is high, but by not consuming all available connections, this keeps the database available for other clients. As noted in the comments, I think we may want to make this configurable.

### DIFF
--- a/server/registry/internal/storage/client.go
+++ b/server/registry/internal/storage/client.go
@@ -17,6 +17,7 @@ package storage
 import (
 	"context"
 	"fmt"
+	"time"
 
 	_ "github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/postgres"
 	"github.com/apigee/registry/server/registry/internal/storage/models"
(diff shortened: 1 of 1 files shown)


server/registry/internal/storage/client.go
```diff
@@ -63,6 +64,11 @@ func NewClient(ctx context.Context, driver, dsn string) (*Client, error) {
 			c.close()
 			return nil, grpcErrorForDBError(ctx, err)
 		}
+		if err := applyConnectionLimits(db); err != nil {
+			c := &Client{db: db}
+			c.close()
```
theganyo, line 69: FWIW, I find this to be a strange pattern we follow: Create a Client in order to call close() on it, which simply calls close() on the db we just constructed it on, only to then throw the client away.

theganyo (approved):

server/registry/internal/storage/client.go
timburks: Good topic -- it looks like it's allowing us to keep the following code in one place:
```
	sqlDB, _ := c.db.DB()
	sqlDB.Close()
```
That could also be done with `Client{db:db}.close()`... or directly using the above code.

I don't have a strong preference. @seaneganx ?

PR #728, opened 2022-09-08
Upload dependencies of protos in uploaded APIs.

timburks:
This makes proto-based API specs fully self-contained. Before uploading protos, the uploader uses `protoc` to compile them and collect a list of referenced files, repeating until all transitive inclusions are listed, and then all of these files are included in the zip archive, making the uploaded spec "fully resolved". The resulting zip archives can be unzipped in a fresh directory and compiled with `protoc` for use in any other desired applications.

A new `--protoc-root` option allows the bulk uploader to upload individual protobuf specs. When `--protoc-root` is used to specify the proto compilation root, the `PATH` argument can point to a subdirectory of that root that includes a directory containing a single API. For example, the registry API can be individually uploaded with the following invocation:
```
registry upload bulk protos ~/googleapis/google/cloud/apigeeregistry/v1 --protoc-root ~/googleapis --project-id googleapis
```
An accompanying change updates the api-linter plugin to work correctly with these uploaded specs. It excludes common protos from linting and removes the prior (ugly) workarounds for incomplete specs (downloading [api-common-protos](https://github.com/googleapis/api-common-protos) and [googleapis](https://github.com/googleapis/googleapis) to resolve missing dependencies).

Here's a simple demonstration that uploads Google API protos and runs [api-linter](https://linter.aip.dev/)-based conformance checks:
```
# Create a sample project.
registry rpc admin create-project --project_id googleapis

# Load the protos for Google APIs.
# This assumes github.com/googleapis/googleapis is cloned to your home directory.
# Proto specs are uploaded as zip files that contain all the protos needed to compile the API spec.
registry upload bulk protos ~/googleapis --project-id googleapis

# Load the AIP-based style guide (get it from https://github.com/apigee/registry-experimental/pull/97).
registry apply -f apihub-styleguide.yaml --parent projects/googleapis/locations/global

# Compute conformance for all APIs. This uses the AIP style guide for all proto-based APIs.
registry compute conformance projects/googleapis/locations/global/apis/-/versions/-/specs/-

# Get one of the conformance reports.
registry get projects/googleapis/locations/global/apis/apigeeregistry/versions/v1/specs/google-cloud-apigeeregistry-v1.zip/artifacts/conformance-apihub-styleguide --print

# Count the conformance reports and specs. The numbers should match.
registry list projects/googleapis/locations/global/apis/-/versions/-/specs/-/artifacts/conformance-apihub-styleguide | wc -l
registry list projects/googleapis/locations/global/apis/-/versions/-/specs/- | wc -l
```

Known problems:
- No tests: #739

### DIFF
--- a/cmd/registry/cmd/upload/bulk/protos.go
+++ b/cmd/registry/cmd/upload/bulk/protos.go
@@ -159,6 +159,7 @@ type uploadProtoTask struct {
 	apiDescription string
 	versionID      string // computed at runtime
 	specID         string // computed at runtime
+	contents       []byte // computed at runtime
 }
 
 func (task *uploadProtoTask) String() string {
@@ -170,6 +171,11 @@ func (task *uploadProtoTask) Run(ctx context.Context) error {
 	task.populateFields()
 	log.Infof(ctx, "Uploading apis/%s/versions/%s/specs/%s", task.apiID, task.versionID, task.specID)
 
+	// Zip up the protos first; if that fails, skip the API.
+	var err error
+	if task.contents, err = task.zipContents(); err != nil {
+		return err
+	}
 	// If the API does not exist, create it.
 	if err := task.createAPI(ctx); err != nil {
 		return err
@@ -238,17 +244,12 @@ func (task *uploadProtoTask) createVersion(ctx context.Context) error {
 }
 
 func (task *uploadProtoTask) createOrUpdateSpec(ctx context.Context) error {
-	contents, err := task.zipContents()
-	if err != nil {
-		return err
-	}
-
 	// Use the spec size and hash to avoid unnecessary uploads.
 	spec, err := task.client.GetApiSpec(ctx, &rpc.GetApiSpecRequest{
 		Name: task.specName(),
 	})
 
-	if err == nil && int(spec.GetSizeBytes()) == len(contents) && spec.GetHash() == hashForBytes(contents) {
+	if err == nil && int(spec.GetSizeBytes()) == len(task.contents) && spec.GetHash() == hashForBytes(task.contents) {
 		log.Debugf(ctx, "Matched already uploaded spec %s", task.specName())
 		return nil
 	}
@@ -258,7 +259,7 @@ func (task *uploadProtoTask) createOrUpdateSpec(ctx context.Context) error {
 			Name:     task.specName(),
 			MimeType: core.ProtobufMimeType("+zip"),
 			Filename: task.fileName(),
-			Contents: contents,
+			Contents: task.contents,
 		},
 		AllowMissing: true,
 	}
@@ -268,7 +269,7 @@ func (task *uploadProtoTask) createOrUpdateSpec(ctx context.Context) error {
 
 	response, err := task.client.UpdateApiSpec(ctx, request)
 	if err != nil {
-		log.FromContext(ctx).WithError(err).Errorf("Error %s [contents-length: %d]", task.specName(), len(contents))
+		log.FromContext(ctx).WithError(err).Errorf("Error %s [contents-length: %d]", task.specName(), len(task.contents))
 	} else {
 		log.Debugf(ctx, "Updated %s", response.Name)
 	}
@@ -334,7 +335,6 @@ func (task *uploadProtoTask) zipContents() ([]byte, error) {
 
 	// Zip the listed files.
 	contents, err := core.ZipArchiveOfFiles(allfiles, prefix, true)
-
 	if err != nil {
 		return nil, err
 	}
@@ -342,8 +342,7 @@ func (task *uploadProtoTask) zipContents() ([]byte, error) {
 	return contents.Bytes(), nil
 }
 
-// collect the names of all metadata files in an source directory,
-// stripping the prefix
+// Collect the names of all metadata files in a source directory, stripping the prefix.
 func localMetadata(source, prefix string) ([]string, error) {
 	filenames := make([]string, 0)
 	err := filepath.WalkDir(source, func(p string, entry fs.DirEntry, err error) error {
@@ -359,8 +358,7 @@ func localMetadata(source, prefix string) ([]string, error) {
 	return filenames, err
 }
 
-// collect the names of all proto files in an source directory,
-// stripping the prefix
+// Collect the names of all proto files in a source directory, stripping the prefix.
 func localProtos(source, prefix string) ([]string, error) {
 	filenames := make([]string, 0)
 	err := filepath.WalkDir(source, func(p string, entry fs.DirEntry, err error) error {
@@ -376,6 +374,7 @@ func localProtos(source, prefix string) ([]string, error) {
 	return filenames, err
 }
 
+// Get all the protos that are referenced in the compilation of a list of protos.
 func referencedProtos(protos []string, root string) ([]string, error) {
 	tempDir, err := os.MkdirTemp("", "proto-import-")
 	if err != nil {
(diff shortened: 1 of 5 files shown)


shrutiparabgoogle:
Looks good to me. 

At some point, should we also provide some option to do this through the UI? If a user uploads a proto through the UI without resolving dependencies, the api linter will fail to lint such specs.

cmd/registry/cmd/upload/bulk/protos.go
```diff
@@ -289,10 +305,117 @@ func (task *uploadProtoTask) fileName() string {
...
+		} else if strings.HasSuffix(p, ".json") || strings.HasSuffix(p, ".yaml") {
+			filenames = append(filenames, strings.TrimPrefix(p, prefix))
+		}
+		return nil
+	})
+	return filenames, err
+}
+
+// Collect the names of all proto files in a source directory, stripping the prefix.
+func localProtos(source, prefix string) ([]string, error) {
```
theganyo, line 362: nit: localProtos() and localMetadata() are identical aside from suffix check. DRY it?

theganyo (commented):
The PR description mentions a "--root" flag, but the closest I see in the code is "--protoc-root" and that doesn't seem to match the description?

timburks:
@theganyo Thanks - the option that I meant is "--protoc-root" (I renamed it). I've updated the description to show that.

cmd/registry/cmd/upload/bulk/protos.go
```diff
@@ -289,10 +305,117 @@ func (task *uploadProtoTask) fileName() string {
...
+	for count > 0 {
+		protos, err = referencedProtos(protos, task.directory)
+		if err != nil {
+			return nil, err
+		}
+		if len(protos) == count {
+			break
+		}
+		count = len(protos)
+	}
```
seaneganx, line 326: I'm having a hard time understanding this loop. First, I think the following is equivalent? 

```
for count := len(protos); count > 0; count = len(protos) {
	protos, err = referencedProtos(protos, task.directory)
	if err != nil {
		return nil, err
	}
	if len(protos) == count {
		break
	}
}
```

The exit conditions are still confusing me though. `count` is the previous length of `protos`? Get more referenced protos as long as the previous one wasn't empty, and stop if the current fetch has the same size as the previous? Why?

cmd/registry/cmd/upload/bulk/protos.go
```diff
@@ -289,10 +305,117 @@ func (task *uploadProtoTask) fileName() string {
...
+	defer os.RemoveAll(tempDir)
+	args := []string{"-o", tempDir + "/proto.pb"}
+	args = append(args, protos...)
+	cmd := exec.Command("protoc", args...)
+	cmd.Dir = root
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("failed to compile protos with protoc: %s", err)
+	}
+	filenames, err := protosFromFileDescriptorSet(tempDir + "/proto.pb")
```
seaneganx, line 389: This can be just `return protosFromFileDescriptorSet(tempDir + "/proto.pb")`.

cmd/registry/cmd/upload/bulk/protos.go
timburks: OK! that wasn't too much to refactor. Updated, PTAL.

cmd/registry/cmd/upload/bulk/protos.go
```diff
@@ -289,10 +305,115 @@ func (task *uploadProtoTask) fileName() string {
...
+		count = len(protos)
+	}
+
+	// Get the metadata files in the main directory.
+	metadata, err := localMetadata(task.path, prefix)
+	if err != nil {
+		return nil, err
+	}
+
+	allfiles := append(metadata, protos...)
```
seaneganx, line 334: I think this would be fine to inline, but otherwise it should probably be `files` or `allFiles`.

cmd/registry/cmd/upload/bulk/protos.go
timburks: That's all true -- I wrote this loop because the call to protoc in referencedProtos wasn't returning the dependencies of dependencies.... so I found that I had to iteratively run it to be sure I had everything.

Later I found that the `--include_imports` option seemed to make this work in one pass, but I left this loop here to verify that this was true (and not trusting my own interpretation). But the loop can probably be removed. (?)

The zero length check is present because I found a directory that appeared to be an API but that had no protos, and this caused protoc to fail. This change allows uploading to proceed (we upload a zip file of metadata only) and leaves the question of whether or not we *should* do that for later.

cmd/registry/cmd/upload/bulk/protos.go
```diff
@@ -79,19 +84,29 @@ func protosCommand() *cobra.Command {
...
+				}
+
+				if err := scanDirectoryForProtos(client, projectID, baseURI, path, root, taskQueue); err != nil {
 					log.FromContext(ctx).WithError(err).Debug("Failed to walk directory")
 				}
 			}
 		},
 	}
 
+	cmd.Flags().StringVar(&root, "protoc-root", "", "Root directory to use for proto compilation, defaults to PATH")
```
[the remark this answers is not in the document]
timburks: I'm on the fence about this. Looking at the isolated option I agree with you, but when I look at the printed help for the command:
```
Usage:
  registry upload bulk protos PATH [flags]

Flags:
      --base-uri string      Prefix to use for the source_uri field of each proto upload
  -h, --help                 help for protos
      --protoc-root string   Root directory to use for proto compilation, defaults to PATH
```
the text as it is seems more explicit and concise. WDYT?

theganyo (approved):

cmd/registry/cmd/upload/bulk/protos.go
timburks: Inlined, thanks!

cmd/registry/cmd/upload/bulk/protos.go
timburks: Fixed, thanks!

theganyo:
Just curious: Would it ever possibly make sense to not gather and include all transitive dependencies or to limit them in any way?

timburks:
> Just curious: Would it ever possibly make sense to not gather and include all transitive dependencies or to limit them in any way?

Maybe? That was what I originally thought that we should do -- upload just what is in a directory and then when compilation is needed, trace all the dependencies and unbundle them together. This requires putting the common protos in the registry as "APIs" and building all the dependency tracing. Meanwhile, @earth2marsh has been preferring that we store "fully resolved specs" in the registry with no dependencies -- here we're registering a snapshot of an API as a user would consume it, and changes to the dependencies could change user experience... i.e. dependencies should be frozen, which is what "fully resolving" here does.

So this is an opinionated upload, but I think that's good for where we are right now. It simplifies downstream tooling and matches what @earth2marsh was wanting all along.

cmd/registry/cmd/upload/bulk/protos.go
timburks: I've replaced this with a non-iterated call to `referencedProtos` and verified it with the current `googleapis` protos.

timburks:
> At some point, should we also provide some option to do this through the UI? If a user uploads a proto through the UI without resolving dependencies, the api linter will fail to lint such specs.

@shrutiparabgoogle rather than push verification logic into the UI, I would rely on verification to ensure validity -- one thing that would verify that proto specs compile is [this](https://github.com/apigee/registry-experimental/pull/100) proposed `registry-experimental compute descriptor` command, which runs `protoc` and saves `FileDescriptorSets` into the registry.

Issue #636, opened 2022-06-28
Registry server: Declarative database associations


theganyo:
Manually maintaining DB integrity can be quite error prone and should generally be done via rules. For example, the current model doesn't use foreign key constraints and we have to manually code relationship integrity during deletion (eg. for [wipeout](https://github.com/apigee/registry/pull/631#issuecomment-1169199219). We should transition to declarative association modeling as much as possible.

GORM can automatically create and maintain the common associations:
* [Belongs To](https://gorm.io/docs/belongs_to.html)
* [Has One](https://gorm.io/docs/has_one.html)
* [Has Many](https://gorm.io/docs/has_many.html)
* [Many To Many](https://gorm.io/docs/many_to_many.html)

theganyo:
It's unclear to me why there is a cascade option in [storage/delete.go](https://github.com/apigee/registry/blob/main/server/registry/internal/storage/delete.go). Would `cascade=false` not leave the database in an inconsistent state?

PR #730, opened 2022-09-08
use foreign keys for delete

theganyo:
Reference #636


pkg/wipeout/wipeout.go
```diff
@@ -28,95 +28,44 @@ import (
...
 	project := "projects/" + projectID + "/locations/global"
-	// Wipeout resources in groups to ensure that children are deleted before parents.
-	{
-		log.Debugf(ctx, "Deleting artifacts")
-		taskQueue, wait := core.WorkerPool(ctx, jobs)
-		wipeoutArtifacts(ctx, client, taskQueue, project+"/apis/-/versions/-/specs/-")
-		wipeoutArtifacts(ctx, client, taskQueue, project+"/apis/-/versions/-")
-		wipeoutArtifacts(ctx, client, taskQueue, project+"/apis/-/deployments/-")
-		wipeoutArtifacts(ctx, client, taskQueue, project+"/apis/-")
-		wipeoutArtifacts(ctx, client, taskQueue, project)
```
timburks, line 39: This last line (39) is deleting top-level artifacts, which won't be deleted by deleting APIs in cascading mode (because top-level artifacts aren't under APIs).

server/registry/internal/storage/models/api.go
```diff
@@ -37,6 +37,8 @@ type Api struct {
 	RecommendedDeployment string    // Recommended API deployment.
 	Labels                []byte    // Serialized labels.
 	Annotations           []byte    // Serialized annotations.
+	ParentKey             string
+	Parent                *Project `gorm:"foreignKey:ParentKey;constraint:OnUpdate:CASCADE,OnDelete:CASCADE"`
```
timburks, line 41: Could we add these keys to an existing database with a database migration? I think that would require something more than the default gorm migration that we're using in the `AdminService` `migrateDatabase` RPC.

pkg/wipeout/wipeout.go
theganyo: Ah. I'll need to update the test. It didn't catch that.

pkg/wipeout/tasks.go
```diff
@@ -1,133 +0,0 @@
...
-	return "delete " + task.name
-}
-
-func (task *DeleteDeploymentTask) Run(ctx context.Context) error {
-	log.Debugf(ctx, "Deleting %s", task.name)
-	return task.client.DeleteApiDeployment(ctx, &rpc.DeleteApiDeploymentRequest{Name: task.name})
-}
-
-// DeleteArtifactTask deletes a specified artifact.
-type DeleteArtifactTask struct {
```
timburks, line 114: Since we also need to keep this task, maybe we should keep this file?

theganyo: Sure

server/registry/internal/storage/models/api.go
theganyo: Unsure. I'll have to figure out a test for that.

PR #734, opened 2022-09-12
log db errors with stack before converting

### LINKED ISSUE #733: Registry server: log errors with source
It would be useful to have line numbers with errors, especially at library boundaries. For example, we generally use this kind of wrapping of errors:

`return fmt.Errorf("failed to upload API spec: %s", err)`

Which yields the following log message:

 FATAL[0001] Task failed: upload spec testdata/cloudtasks/v2/openapi.yaml error=failed to upload API spec: rpc error: code = InvalidArgument desc = invalid version name "projects/csv-demo/apis/cloudtasks/versions/v2": must match "^projects/([A-Za-z0-9-.]+)/locations/global/apis/([A-Za-z0-9-.]+)/versions/([A-Za-z0-9-.]+)$"

But if we wrap with errors.Wrap(), apex automatically adds source and line numbers:

`return errors.Wrap(err, "failed to upload API spec")`

Which yields a log message with a `source` field:

 FATAL[0001] Task failed: upload spec testdata/cloudtasks/v2/openapi.yaml error=failed to upload API spec: rpc error: code = InvalidArgument desc = invalid version name "projects/csv-demo/apis/cloudtasks/versions/v2": must match "^projects/([A-Za-z0-9-.]+)/locations/global/apis/([A-Za-z0-9-.]+)/versions/([A-Za-z0-9-.]+)$" source=uploadSpecTask.Run: /Users/theganyo/dev/registry/cmd/registry/cmd/upload/csv.go:250

(See also info about the errors pkg: https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package)

theganyo:
This improves logging by using wrapping and adding context. It also enables stack traces for unhandled errors.

Fixes #733


theganyo:
Example from log:

```
  INFO[0002] DBError database is locked
create &models.Artifact{Key:"projects/controller-demo/locations/global/apis/petstore/versions/1.1.0/specs/openapi.yaml/artifacts/test-receipt-artifact", ProjectID:"controller-demo", ApiID:"petstore", VersionID:"1.1.0", SpecID:"openapi.yaml", DeploymentID:"", ArtifactID:"test-receipt-artifact", CreateTime:time.Date(2022, time.September, 12, 15, 10, 21, 488275000, time.Local), UpdateTime:time.Date(2022, time.September, 12, 15, 10, 21, 488275000, time.Local), MimeType:"application/octet-stream;type=google.cloud.apigeeregistry.v1.controller.Receipt", SizeInBytes:28, Hash:"f9b9fde4b133b45cd783cc372432f9259a665e2fd5dfc40d6aaa50534e63454a"}
github.com/apigee/registry/server/registry/internal/storage.(*Client).create
	/Users/theganyo/dev/registry/server/registry/internal/storage/create.go:56
github.com/apigee/registry/server/registry/internal/storage.(*Client).CreateArtifact
	/Users/theganyo/dev/registry/server/registry/internal/storage/create.go:41
github.com/apigee/registry/server/registry.(*RegistryServer).CreateArtifact.func1
	/Users/theganyo/dev/registry/server/registry/actions_artifacts.go:91
github.com/apigee/registry/server/registry/internal/storage.(*Client).Transaction.func1
	/Users/theganyo/dev/registry/server/registry/internal/storage/client.go:173
gorm.io/gorm.(*DB).Transaction
	/Users/theganyo/go/pkg/mod/gorm.io/gorm@v1.21.14/finisher_api.go:545
github.com/apigee/registry/server/registry/internal/storage.(*Client).Transaction
	/Users/theganyo/dev/registry/server/registry/internal/storage/client.go:172
github.com/apigee/registry/server/registry.(*RegistryServer).runInTransaction
	/Users/theganyo/dev/registry/server/registry/server.go:103
github.com/apigee/registry/server/registry.(*RegistryServer).CreateArtifact
	/Users/theganyo/dev/registry/server/registry/actions_artifacts.go:69
github.com/apigee/registry/rpc._Registry_CreateArtifact_Handler
	/Users/theganyo/dev/registry/rpc/registry_service_grpc.pb.go:1233
google.golang.org/grpc.(*Server).processUnaryRPC
	/Users/theganyo/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:1279
google.golang.org/grpc.(*Server).handleStream
	/Users/theganyo/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:1608
google.golang.org/grpc.(*Server).serveStreams.func1.2
	/Users/theganyo/go/pkg/mod/google.golang.org/grpc@v1.41.0/server.go:923
runtime.goexit
	/usr/local/go/src/runtime/asm_arm64.s:1172
```

server/registry/internal/storage/errors.go
```diff
@@ -47,76 +48,82 @@ func AlreadyExists(err error) bool {
 
 // grpcErrorForDBError converts recognized database error codes to grpc error codes.
 func grpcErrorForDBError(ctx context.Context, err error) error {
+	// unwrap err, if wrapped
+	cause := errors.Cause(err)
```
timburks, line 52: Had to look up [Cause](https://github.com/pkg/errors/blob/v0.9.1/errors.go#L275): "If the error does not implement Cause, the original error will be returned. If the error is nil, nil will be returned without further investigation."

server/registry/internal/storage/errors.go
```diff
@@ -47,76 +48,82 @@ func AlreadyExists(err error) bool {
...
 	// if this error already has a gRPC status code, just return it
-	if _, ok := status.FromError(err); ok {
-		return err
+	if _, ok := status.FromError(cause); ok {
+		return cause
 	}
-	if err == context.DeadlineExceeded {
-		return status.Error(codes.DeadlineExceeded, err.Error())
+
+	log.Infof(ctx, "DBError %+v", err) // log the stack trace
```
timburks, line 59: This ensures that errors from each call site are uniquely logged?

timburks (approved):
LGTM, just added comments to clarify my understanding

server/registry/internal/storage/errors.go
theganyo: Yes, this ensures we have a stack trace of the source error in the server log before we translate it for the client. This also potentially raises questions... Do we want to log this stack trace here? Do we want to log it at info level? Should we filter which errors are logged? But this strategy can be tweaked later based on how useful we find it in practice and this seems like a simple place to start.

timburks: I'm not concerned about oversharing there.  We might want to push the log level to debug (@seaneganx ?), but I'm fine to defer that to a future change.

timburks: BTW, only the first few levels of that call stack are informative.

theganyo: Yeah, I agree about the stack... but do you want us to write something to trim the stack?

seaneganx: This seems like something appropriate for debug level. There should already be some kind of error level log being produced, although with less information (just the error message itself I think).

If this is logged at debug level I don't think we need to worry about trimming the stack. Printing one extra debug log per failed request doesn't seem very noisy to me.

seaneganx (approved):

server/registry/internal/storage/errors.go
theganyo: 👍

PR #735, opened 2022-09-14
Registry tool: Deprecate `--contents` flag on `registry get`, replace it with `--print` and `--raw`

timburks:
Prior to this change, the `registry get` command did not offer a way to directly access the raw contents of artifacts - it printed known types as JSON and for unknown types, it printed a nearly-useless text dump of bytes for unknown types. This adds two new flags to `registry get`:

- `--raw`, which causes the command to write the raw bytes of the artifact to standard output.
- `--print`, which causes the command to print a formatted version of known artifact types to standard output. For unknown types, an error is printed with a message to use `--raw` to access the raw bytes.

To avoid disruption, the `--contents` flag is kept and associated with `--print` but a deprecation warning is printed when it is used.

Likely followup issues:
1. modify `--print` for specs to disallow printing of zip archives.
2. unit testing of `registry get`.
3. (eventual) deletion of the `--contents` flag.

### DIFF
--- a/cmd/registry/core/print.go
+++ b/cmd/registry/core/print.go
@@ -156,9 +156,6 @@ func PrintArtifactContents(artifact *rpc.Artifact) error {
 		return unmarshalAndPrint(artifact.GetContents(), &openapiv2.Document{})
 	case "gnostic.openapiv3.Document":
 		return unmarshalAndPrint(artifact.GetContents(), &openapiv3.Document{})
-	case "google.protobuf.FileDescriptorSet":
-		os.Stdout.Write(artifact.GetContents())
-		return nil
 	default:
 		fmt.Printf("Unsupported message type: %s, please use --contents to get raw contents", messageType)
 		return nil
--- a/cmd/registry/core/print.go
+++ b/cmd/registry/core/print.go
@@ -78,7 +78,7 @@ func PrintSpecDetail(message *rpc.ApiSpec) error {
 	return nil
 }
 
-func PrintSpecContents(message *rpc.ApiSpec) error {
+func WriteSpecContents(message *rpc.ApiSpec) error {
 	contents := message.GetContents()
 	if strings.Contains(message.GetMimeType(), "+gzip") {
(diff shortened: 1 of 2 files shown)


cmd/registry/core/print.go
```diff
@@ -152,7 +159,8 @@ func PrintArtifactContents(artifact *rpc.Artifact) error {
 	case "gnostic.openapiv3.Document":
 		return unmarshalAndPrint(artifact.GetContents(), &openapiv3.Document{})
 	default:
-		fmt.Printf("%+v", artifact.GetContents())
+		// To get a parent context here would require changing the ArtifactHandler type in handlers.go
+		log.FromContext(context.TODO()).Fatalf("Unsupported message type: %s, please use --raw to get raw contents", messageType)
```
seaneganx, line 163: Can this be returned as an error that gets logged by the caller?

timburks: Thanks, I missed that - I'm returning an error now and changed the caller to log it at Error instead of Debug.

cmd/registry/core/print.go
```diff
@@ -97,6 +99,11 @@ func PrintArtifactDetail(artifact *rpc.Artifact) error {
 	return nil
 }
 
+func WriteArtifactContents(artifact *rpc.Artifact) error {
+	os.Stdout.Write(artifact.GetContents())
```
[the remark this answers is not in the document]
timburks: `os.Stdout.Write` is explicitly writing '[]byte', `fmt.Print` prints a "human friendly" version of `[]byte`:

```
package main

import (
	"fmt"
	"os"
)

func main() {
	h := "hello\n"
	b := []byte(h)
	fmt.Print(h)
	fmt.Print(b)
	os.Stdout.Write(b)
}
```
produces
```
hello
[104 101 108 108 111 10]hello
```

seaneganx (approved):

theganyo (approved):
LGTM 
(Although the large nested if statement reminds me of how much I kinda hate our name parsing.)

PR #738, opened 2022-09-14
remove extra db query in delete

### LINKED ISSUE #736: Registry server: remove extra db query in delete
In the actions layer, we query the database to check if the entity exists to return an error before performing a delete. This adds an unnecessary database call as we can check the return for RowsAffected == 0 on the target row.

theganyo:
Fixes #736

### DIFF
--- a/server/registry/internal/storage/delete.go
+++ b/server/registry/internal/storage/delete.go
@@ -59,6 +59,9 @@ func (c *Client) DeleteProject(ctx context.Context, name names.Project, cascade 
 		if err := op.Delete(model).Error; err != nil {
 			return err
 		}
+		if _, ok := model.(models.Project); ok && op.RowsAffected == 0 {
+			return status.Errorf(codes.NotFound, "%q not found in database", name)
+		}
 		counts[i] = op.RowsAffected
 	}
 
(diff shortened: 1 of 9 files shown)


server/registry/internal/storage/delete.go
```diff
@@ -303,8 +325,8 @@ func (c *Client) DeleteDeploymentRevision(ctx context.Context, name names.Deploy
 
 func (c *Client) DeleteArtifact(ctx context.Context, name names.Artifact) error {
 	for _, model := range []interface{}{
-		models.Blob{},
 		models.Artifact{},
+		models.Blob{},
```
timburks, line 329: Was this reordering functionally necessary? I was tempted to label it ["drive by fix"](http://tillenius.me/blog/2010/05/03/problems-with-drive-by-fixes/) but I'm wondering if there's an ordering sensitivity that we should note.

timburks (approved):
Nice improvement, thanks!

server/registry/internal/storage/delete.go
theganyo: This is basically the same optimization: There's would be no point in calling the DB to delete blobs if the Artifact doesn't exist.

timburks: Makes sens, thanks!

seaneganx (approved):

PR #742, opened 2022-09-15
remove extra query in non-revision updates

### LINKED ISSUE #737: Registry server: remove extra query in update
In the actions layer, we query the database to check if the entity exists and get the current values from the db before performing an update. This adds an unnecessary database call as we can check the return for RowsAffected == 0 on the target row after update and we already have a mask for columns that shouldn't be updated.

theganyo:
Partially fixes #737

The remaining updates (Deployment, Spec) have to maintain revisions and will require different and more complex logic to handle in a similar manner, so I'll submit that in a separate PR so we can evaluate that approach separately.


theganyo:
The test timeout seems unrelated to this PR. Needs investigation, but possibly db lock contention?

server/registry/actions_apis.go
```diff
@@ -176,22 +176,22 @@ func (s *RegistryServer) UpdateApi(ctx context.Context, req *rpc.UpdateApiReques
...
+		api, err := models.NewApi(name, req.GetApi())
+		if err != nil {
 			return err
-		} else {
+		}
+		mask := models.ExpandMask(req.GetApi(), req.GetUpdateMask()).GetPaths()
+		if err := db.SaveApi(ctx, api, mask); err != nil {
+			if status.Code(err) == codes.NotFound && req.GetAllowMissing() {
+				response, err = s.createApi(ctx, db, name, req.GetApi())
+				if status.Convert(err).Code() == codes.AlreadyExists {
```
seaneganx, line 187: What's the difference between `status.Convert(err).Code()` and `status.Code(err)`? This is my first time seeing the convert method.

server/registry/internal/storage/save.go
```diff
@@ -80,3 +92,23 @@ func (c *Client) save(ctx context.Context, v interface{}) error {
...
+	return grpcErrorForDBError(ctx, err)
+
+	// op := c.db.WithContext(ctx).
+	// 	Select(fieldMask).
+	// 	Clauses(
+	// 		clause.Returning{},
+	// 		clause.OnConflict{UpdateAll: true},
+	// 	)
+	// err := op.Create(v).Error
+	// return grpcErrorForDBError(ctx, err)
```
seaneganx, line 113: Did you intend on checking this in? IMO we shouldn't check in code that's commented out.

server/registry/internal/storage/save.go
```diff
@@ -18,55 +18,67 @@ import (
...
 
 	"github.com/apigee/registry/server/registry/internal/storage/models"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"gorm.io/gorm/clause"
 )
 
-func (c *Client) SaveProject(ctx context.Context, v *models.Project) error {
+// SaveProject will return codes.NotFound if key not found
+func (c *Client) SaveProject(ctx context.Context, v *models.Project, fieldMask []string) error {
```
seaneganx, line 27: Is there a benefit to accepting `[]string` here instead of `*fieldmaskpb.FieldMask`? The only difference this seems to make is whether we call `GetPaths()` in the action handler or in the `saveWithMask` method, so I lean toward keeping the more specialized type with all of its documentation about what a field mask is.

log/log.go
```diff
@@ -43,7 +43,7 @@ type Logger interface {
 
 var defaultOptions = []Option{
 	TextFormat(os.Stderr),
-	InfoLevel,
+	DebugLevel,
```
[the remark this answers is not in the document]
theganyo: Nope. Thanks, good catch.

server/registry/actions_apis.go
theganyo: From the [docs](https://pkg.go.dev/google.golang.org/grpc/status#Convert): `Convert is a convenience function which removes the need to handle the boolean return value from FromError.`

theganyo: Hmm. Looking at `status.Code(err)` now, perhaps I didn't answer your real question of why not use that? I'll try it. Seems like there are plenty of options in this package. LOL

theganyo: Ok, changed.

server/registry/internal/storage/save.go
theganyo: In retrospect, I guess there's no point in avoiding the dependency in this package... we rely on the protobuf stuff everything. I'll change it.

theganyo: Done

seaneganx (approved):

PR #759, opened 2022-09-21
optimize paging

theganyo:
Optimization: If result size is less than the requested limit, it's not necessary to query the next page for more results.

### DIFF
--- a/server/registry/internal/storage/client.go
+++ b/server/registry/internal/storage/client.go
@@ -70,7 +70,6 @@ func NewClient(ctx context.Context, driver, dsn string) (*Client, error) {
 			c.close()
 			return nil, grpcErrorForDBError(ctx, err)
 		}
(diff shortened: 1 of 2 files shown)


server/registry/internal/storage/client.go
```diff
@@ -70,6 +70,7 @@ func NewClient(ctx context.Context, driver, dsn string) (*Client, error) {
 			c.close()
 			return nil, grpcErrorForDBError(ctx, err)
 		}
+		db.Exec("PRAGMA foreign_keys = ON")
```
seaneganx, line 73: Does this affect paging?

seaneganx (approved):

server/registry/internal/storage/client.go
theganyo: No, that shouldn't be there... my git history is screwy. Will remove.

PR #760, opened 2022-09-22
add foreign key functionality

### LINKED ISSUE #636: Registry server: Declarative database associations

theganyo:
Adds foreign key relationships between Project, Api, Spec, Deployment, and Version. Uses the fk constraints to ensure: 1) parent must exist during create; and 2) deletes will be cascaded by the DB, if client requests.

Requires migration for existing databases.

Fixes #636


theganyo:
Two notes: 

1. The migrate script still needs to be written.

3. The CI is failing on postgres, but completes on a clean database locally. Need to investigate this discrepancy.

seaneganx (approved):

timburks (approved):

PR #763, opened 2022-09-25
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/apigee-registry/commit/1340fd20b3eb50baa1a24504c7c48cde14d47ec7#commitcomment-84945593

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/apigee-registry/commit/8683661af799c431b1f28ae5c783cd520c7046fe

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

PR #764, opened 2022-09-29
Make grammar fixes suggested in #763 and add a spelling correction that was lost in a merge.

timburks:


server/registry/internal/storage/models/api.go
```diff
@@ -104,7 +104,7 @@ func (api *Api) Message() (message *rpc.Api, err error) {
 	return message, nil
 }
 
-// Update modifies a api using the contents of a message.
```
timburks, line 107: These probably go back to early times when "api" was called "product" and the article wasn't fixed in the renaming.

theganyo (approved):

PR #766, opened 2022-09-29
improve api create message for invalid IDs

### LINKED ISSUE #765: Registry server: improve api create message for invalid IDs
The api errors for creating with a missing or invalid ID doesn't make it entirely clear how for a client to proceed. For example, this is the error when attempting to create a project via the registry command with no ID specified:

`rpc error: code = InvalidArgument desc = invalid project name "projects/": must match "^projects/([A-Za-z0-9-.]+)$"`

This should be easily fixed with two improvements:

1. Perform the ID check before the entire name regexp check the [Validate()](https://github.com/apigee/registry/blob/main/server/registry/names/project.go#L29) functions for each type.
2. Add an explicit check and error for a missing ID in the common [validateID()](https://github.com/apigee/registry/blob/main/server/registry/names/common.go#L45) function.

theganyo:
New error examples:

`rpc error: code = InvalidArgument desc = invalid identifier "": identifier is required`

`rpc error: code = InvalidArgument desc = invalid identifier "-": must begin with a number or letter`

Fixes #765


server/registry/names/common.go
```diff
@@ -43,7 +43,9 @@ func GenerateID() string {
 
 // validateID returns an error if the provided ID is invalid.
 func validateID(id string) error {
-	if !customIdentifier.MatchString(id) {
+	if "" == id {
+		return fmt.Errorf("invalid identifier %q: identifier is required", id)
```
timburks, line 47: I think you chose this wording for the CLI (makes sense) but this could also be triggered if someone sets the id to "". Would "nonempty identifier is required" be clear enough to cover all cases?

theganyo: I chose it to match all the other error messages below it.

timburks: Right, makes sense. So how about 
"invalid identifier %q: nonempty identifier is required" 
or
"invalid identifier %q: identifier must be nonempty"
? I slightly prefer the first (particularly for the CLI case).

server/registry/names/api.go
```diff
@@ -28,12 +28,16 @@ type Api struct {
 // Validate returns an error if the resource name is invalid.
 // For backward compatibility, names should only be validated at creation time.
 func (a Api) Validate() error {
+	if err := validateID(a.ApiID); err != nil {
+		return err
+	}
```
timburks, line 33: I'm surprised that this is untested because we have [this test case](https://github.com/apigee/registry/blob/616e3a92c71fbb528430f649700215f8ec3c9f06/server/registry/names/names_test.go#L88). Did we miss the real source of the error message? It seems that to get here we need a `names.Api`, and we might be getting that with [names.ParseApi()](https://github.com/apigee/registry/blob/616e3a92c71fbb528430f649700215f8ec3c9f06/server/registry/names/api.go#L98), which will return the error that we don't want to see before we get here.

server/registry/names/api.go
```diff
@@ -28,12 +28,16 @@ type Api struct {
...
+
 	r := apiRegexp()
 	if name := a.String(); !r.MatchString(name) {
 		return fmt.Errorf("invalid API name %q: must match %q", name, r)
```
timburks, line 37: Unfortunately, I think this error message is the same as [this one](https://github.com/apigee/registry/blob/616e3a92c71fbb528430f649700215f8ec3c9f06/server/registry/names/api.go#L101).

server/registry/names/api.go
theganyo: I tested this manually and got the errors I put in the description, so... no?

server/registry/names/common.go
theganyo: sure

PR #782, opened 2022-10-13
artifacts on spec revisions

theganyo:
This PR shows how moving artifacts to spec revisions might be supported without any schema changes. The intent is to spark discussion of whether this approach is appropriate. 

Potential simplifications / alternatives:
The preexisting approach (preserved in this PR) is to have two external entities (Spec and SpecRevision) and a single internal model (Spec) and this approach. This adds some impedance mismatch complexity, of course, so for comparison and potential discussion, we could also consider and discuss these options:
1. Collapse the Spec and SpecRevision entities into a single entity to match the existing internal db model.
2. Separate the internal Spec model (db schema) into Spec and SpecRevision db tables and consider these as completely separate entities.


timburks:
This doesn't change any protos, so could you elaborate on what you mean by "two external entities (Spec and SpecRevision)"?

theganyo:
> This doesn't change any protos, so could you elaborate on what you mean by "two external entities (Spec and SpecRevision)"?

Right, not protos... I wasn't clear on that, I meant external to the `internal/storage`. (So many layers. :))

theganyo:
BTW: The build error is caused by the patterns in the controller patterns all still relying on Spec instead of SpecRevision. Fixing this seems like it may require fairly extensive work and will have to further incorporate the intricacies of the naming package's differentiation between the two. This is one of the motivating factors for asking the question if it would be better to combine the two naming entities - as exists in the API and the DB.

PR #783, opened 2022-10-13
Enhance "registry apply" to infer --parent from registry configuration.

timburks:
This makes `--parent` optional for `registry apply` by using the configured `registry.project` when it is nonempty.

Manually verified... I looked into writing tests for this and didn't see a good way to set the registry configuration in an isolated way. Reviewers, WDYT? I'm thinking maybe the tests should save the configuration, modify it as needed, and put it back afterwards, but that seems risky and messy. Since we're not actually running the registry tool in tests, could there be a better way to do this by writing the configuration data structures directly?

### DIFF
--- a/cmd/registry/cmd/apply/apply.go
+++ b/cmd/registry/cmd/apply/apply.go
@@ -36,6 +36,16 @@ func Command() *cobra.Command {
 		Args:  cobra.ExactArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			ctx := cmd.Context()
+			if parent == "" {
+				c, err := connection.ActiveConfig()
+				if err != nil {
+					log.FromContext(ctx).WithError(err).Fatal("Unable to identify parent: please use --parent or registry configuration")
+				}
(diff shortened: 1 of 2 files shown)


theganyo:
The most straightforward way right now is to set the config dir, create a config, save it and go like is done [here]( https://github.com/apigee/registry/blob/main/cmd/registry/cmd/config/config_test.go#L80). Otherwise, I think we'll have to refactor to add some indirection in front of the global `connection.ActiveConfig()` function that's used throughout the cmd functions.

theganyo (approved):

timburks:
@theganyo I had an idea about this that I got working with some hacks -- essentially modify `connection.ActiveConfig` to cache the configuration when it is initially read, and then modify the cached value so that when it is returned by subsequent calls to `connection.ActiveConfig`, we get the modified configuration. This was bit harder than I had hoped because configurations are returned by value everywhere so I can't just get it, modify it, and expect the cached value to also be modified.

I'll put it in a followup PR so we can discuss it.

PR #784, opened 2022-10-14
Allow overriding the registry tool configuration in tests.

timburks:
This includes modifications to the config package that allow the registry.project to be forced to particular values during testing.

I'm not attached to the implementation, this is just one way to pursue a goal of testing the registry tool under different configurations without modifying files in possibly-surprising ways.

### DIFF
--- a/pkg/connection/config.go
+++ b/pkg/connection/config.go
@@ -33,17 +33,17 @@ type Config struct {
 
 // If set, ActiveConfig() returns this configuration.
 // This is intended for use in testing.
-var override *Config
+var active *Config
 
 // SetConfig forces the active configuration to use the specified value.
 func SetConfig(config Config) {
-	override = &config
+	active = &config
 }
 
 // ActiveConfig returns the active config.
 func ActiveConfig() (Config, error) {
-	if override != nil {
-		return *override, nil
+	if active != nil {
+		return *active, nil
 	}
 
 	name, err := config.ActiveName()
--- a/pkg/connection/config.go
+++ b/pkg/connection/config.go
@@ -31,27 +31,19 @@ type Config struct {
 	Token    string `mapstructure:"token"`    // bearer token
 }
 
-// OverrideActiveConfigRegistryProject forces the active configuration to use the specified project.
-func OverrideActiveConfigRegistryProject(project string) {
-	// force the config to be read and cached
-	useCache = true
-	_, err := ActiveConfig()
-	if err == nil && cachedConfig != nil {
-		cachedConfig.Project = project
-	}
-}
+// If set, ActiveConfig() returns this configuration.
+// This is intended for use in testing.
+var override *Config
 
-var cachedConfig *Config
-var useCache bool
-
-func ClearCachedConfig() {
-	cachedConfig = nil
+// SetConfig forces the active configuration to use the specified value.
+func SetConfig(config Config) {
+	override = &config
 }
 
 // ActiveConfig returns the active config.
 func ActiveConfig() (Config, error) {
-	if cachedConfig != nil {
-		return *cachedConfig, nil
+	if override != nil {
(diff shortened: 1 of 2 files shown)


pkg/connection/config.go
```diff
@@ -31,13 +31,39 @@ type Config struct {
...
+	// force the config to be read and cached
+	useCache = true
+	_, err := ActiveConfig()
+	if err == nil && cachedConfig != nil {
+		cachedConfig.Project = project
+	}
+}
+
+var cachedConfig *Config
+var useCache bool
```
seaneganx, line 45: These globals should probably be at the top of the file. I don't think I'm particularly unusual with that being the first place I look when functions access a variable that isn't a parameter.

Also, do we need globals here? I'm not familiar enough with the code to propose other designs, but maybe there are other more functional options that are less likely to cause surprises in the future.

pkg/connection/config.go
```diff
@@ -31,13 +31,39 @@ type Config struct {
...
+
+func ClearCachedConfig() {
```
seaneganx, line 47: What's the use case for this? I'm curious why it's added but not used.

timburks: Oops, I'll take this out. I added it to try to fix other tests that were broken by this change but found that the `useCache` flag (hack) made this unnecessary.

pkg/connection/config.go
timburks: I put those variables there to be close to their point of use... I don't mind moving them but I'm unsure that it's helpful. We should probably just get rid of them...

if Go supported static variables inside functions, I probably would have used that instead. [This](https://stackoverflow.com/a/30558173) might be a reasonable alternative.

theganyo:
I'd suggest just having a function called "SetConfig()" that sets the config pointer and call it a config, not a cache. I don't believe we need the useCache or clearConfig or anything like that. If the config pointer is populated, we use it; if not, we attempt to load the config.

timburks:
> I'd suggest just having a function called "SetConfig()" that sets the config pointer and call it a config, not a cache. I don't believe we need the useCache or clearConfig or anything like that. If the config pointer is populated, we use it; if not, we attempt to load the config.

@theganyo This seems to mean that when we set any config value in code, we would need to set all of the config values. I think in tests we might want to set the project but not touch the configured server address.

theganyo:
I see... So I think what I hear you saying is that you'd like to load and keep certain external values (from, I assume, env vars), but then override other specific values for specific tests. In that case, why not just get the config, modify the values you want to change, and then SetConfig() back?

timburks:
That sounds good, as long as `SetConfig()` doesn't write to the configuration file -- I don't think we want this to have side effects outside the tests.

pkg/connection/config.go
```diff
@@ -31,12 +31,26 @@ type Config struct {
 	Token    string `mapstructure:"token"`    // bearer token
 }
 
+// If set, ActiveConfig() returns this configuration.
+// This is intended for use in testing.
+var override *Config
```
timburks, line 36: @theganyo this is not called "config" because that collides with the `config` package name and breaks the call to `config.ActiveName()` below.

timburks: @seaneganx the only thing above this in the file is the definition of the `Config` type.

theganyo: Would `active` or `cached` be appropriate?

theganyo (approved):
The potential naming and/or caching behavior is a nit, so LGTM if you don't want to address that.

pkg/connection/config.go
seaneganx: +1 for `active`

seaneganx (approved):

timburks:
I'll go with `active` on your joint recommendation. Thanks!

PR #788, opened 2022-10-19
Replace ubuntu-latest in actions with ubuntu-22.04

timburks:
Feeling a bit burned by the surprise failures that led to #786, I'd like to also pin the ubuntu image used in testing to an explicit version. Any concerns?

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -35,7 +35,7 @@ env:
   DOCKER_REPOSITORY_OWNER: ${{github.repository_owner}}
 jobs:
   build-registry-server:
-    runs-on: ubuntu-2204
+    runs-on: ubuntu-22.04
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


theganyo:
Counter-point: CI did its job of pointing out an issue that we might not have noticed until a user attempted an install of Postgres 15. This change will likewise ensure we don't know about any issues in new versions of ubuntu.

theganyo:
Counter counter-point: Only testing on latest could cause us to accidentally drop support for older versions.

Arguably, we should:

1. Test on each platform version that we support.
2. Also test on latest platform version.

This way:

1. We don't accidentally rely on new features that don't exist in supported versions.
2. We can also see if there are coming support issues with future versions.

theganyo (approved):
Anyway, the bigger issues with our CI aside, nothing wrong with this change per se.

timburks:
@theganyo Thanks - I think the main reason to pin is to avoid surprises that block everything until the upgrade consequence gets fixed. Compatibility testing across versions seems better to do at a different frequency - maybe a weekly "canary" run that picks up the latest versions of all dependencies so we can get a heads up and plan a response, like the issue for Postgres 15 support that I need to file...

giteshk:
After reviewing the Github actions files, I think the ubuntu-latest is the only to define the machine to run the jobs on. 
Eventually we pick specific docker container images that run on the ubuntu-latest runner.

I don't have objections to tying to to a specific version, I just prefer the ubuntu-latest.

giteshk (approved):

PR #790, opened 2022-10-20
Report foreign key errors as InvalidArgument errors.

timburks:
Just curious to see what happens if we return `InvalidArgument` instead of `NotFound` for foreign key violations.

### DIFF
--- a/server/registry/internal/storage/errors.go
+++ b/server/registry/internal/storage/errors.go
@@ -77,7 +77,7 @@ func grpcErrorForDBError(ctx context.Context, err error) error {
 		} else if v.Code.Name() == "query_canceled" {
 			return status.Error(codes.Canceled, cause.Error())
 		} else if v.Code.Name() == "foreign_key_violation" {
-			return status.Error(codes.NotFound, cause.Error())
+			return status.Error(codes.InvalidArgument, cause.Error())
 		}
 		log.Infof(ctx, "Unhandled %T %+v code=%s name=%s", v, v, v.Code, v.Code.Name())
 	case *net.OpError:
--- a/server/registry/internal/storage/errors_cgo_enabled.go
+++ b/server/registry/internal/storage/errors_cgo_enabled.go
@@ -48,7 +48,7 @@ func grpcErrorForSQLite3Error(ctx context.Context, err error) error {
 			return status.Error(codes.Unavailable, err.Error())
 		}
 		if v.Code == sqlite3.ErrConstraint && v.ExtendedCode == sqlite3.ErrConstraintForeignKey {
-			return status.Error(codes.NotFound, v.ExtendedCode.Error())
+			return status.Error(codes.InvalidArgument, v.ExtendedCode.Error())
 		}
 		log.Infof(ctx, "Unhandled %T %+v code=%d extended=%d", v, v, v.Code, v.ExtendedCode)
 	}


server/registry/actions_apis_test.go
```diff
@@ -139,7 +139,7 @@ func TestCreateApiResponseCodes(t *testing.T) {
 				ApiId:  "valid-id",
 				Api:    &rpc.Api{},
 			},
-			want: codes.NotFound,
```
timburks, line 142: @seaneganx Why were we reporting missing parents as `NotFound` errors? I can see that the parent is missing, but I think there is a good case for using `InvalidArgument` instead: Typically `NotFound` corresponds to the object of the RPC - the resource we are getting, updating, or deleting, or even the owner of the collection we are listing. For creates with missing parents, the object of the creation isn't `not found`, it was invalidly specified.

seaneganx: I have two main concerns.

---

Firstly, I'm not understanding the "object of the RPC" concept in the same way you are.

>Typically NotFound corresponds to the object of the RPC - the resource we are getting, updating, or deleting, or even the owner of the collection we are listing.

How can we define "object of the RPC" in a way that includes the parent (owner of the collection we're trying to read) in a List operation but not the parent (owner of the collection we're trying to modify) in a Create operation?

Note: In both List and Create requests the parent is the owner of the affected collection, not the collection itself, e.g. `/v1/{parent=publishers/*}/books`

Perhaps we're coming to different conclusions because I consider resource creation an operation on the parent, while you consider it an operation on the resource being created. Why do create requests ask for a parent + id combination instead of a full resource name if the object of the RPC is the child being created? That leads me to consider creation an operation on the parent.

---

Secondly, using the `InvalidArgument` in the way you're proposing seems to directly violate its definition.

>InvalidArgument indicates client specified an invalid argument. Note that this differs from FailedPrecondition. It indicates arguments that are problematic regardless of the state of the system (e.g., a malformed file name).
>
>https://pkg.go.dev/google.golang.org/grpc/codes#Code

Returning `InvalidArgument` when the parent isn't found and returning `OK` when the parent is found seems to imply the argument is only problematic when the system is in a particular state.

timburks: `FailedPrecondition` seems better to me, thanks for pointing that out. I'm strongly inclined to switch to that for all foreign key violations.

As for the object of creation, I think you might be thinking more in terms of the implementation than a typical user would. It's true that "make me a sandwich" is a request to someone (who is not a sandwich) but the thing that a user wants is a sandwich.

timburks:
From https://grpc.github.io/grpc/core/md_doc_statuscodes.html:

> The operation was rejected because the system is not in a state required for the operation's execution. For example, the directory to be deleted is non-empty, an rmdir operation is applied to a non-directory, etc. Service implementors can use the following guidelines to decide between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE: (a) Use UNAVAILABLE if the client can retry just the failing call. (b) Use ABORTED if the client should retry at a higher level (e.g., when a client-specified test-and-set fails, indicating the client should restart a read-modify-write sequence). (c) Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed. E.g., if an "rmdir" fails because the directory is non-empty, FAILED_PRECONDITION should be returned since the client should not retry unless the files are deleted from the directory.

server/registry/internal/storage/errors.go
```diff
@@ -77,7 +77,7 @@ func grpcErrorForDBError(ctx context.Context, err error) error {
 		} else if v.Code.Name() == "query_canceled" {
 			return status.Error(codes.Canceled, cause.Error())
 		} else if v.Code.Name() == "foreign_key_violation" {
-			return status.Error(codes.NotFound, cause.Error())
```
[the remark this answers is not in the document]
timburks: Updating this to `FailedPrecondition`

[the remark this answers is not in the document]
theganyo: Ironically, `FailedPrecondition` was my first instinct as well - before I ended up switching it to `NotFound` to allow retries and be more consistent with pre-existing behavior.

server/registry/actions_apis_test.go
seaneganx: >`FailedPrecondition` seems better to me, thanks for pointing that out. I'm strongly inclined to switch to that for all foreign key violations.

As far as I'm aware, 100% of resource oriented APIs I've encountered use NOT_FOUND as we've been using it. Can you link me to any examples of APIs that use FAILED_PRECONDITION (or really anything other than NOT_FOUND) for this situation, or do we simply not care that this would be an unusual thing to do?

>As for the object of creation, I think you might be thinking more in terms of the implementation than I think a typical user would.

I've only referred to the API methods and fields on the request messages as our users would experience them. All they should know about is the resources in our API and their relationships _as specified in the protos_. OTOH you're talking about foreign key constraints, which seems much more implementation focused.

>It's true that "make me a sandwich" is a request to someone (who is not a sandwich) but the thing that a user wants is a sandwich.

Right, but we're talking about the design of a resource-oriented API where `CreateSandwichRequest` and every other request message for Create methods describes it as an action performed on a parent resource. You seem to be leaning into REST concepts that I don't think map onto our API's design.

I think we can agree this is what the request would look like.
```
message CreateSandwichRequest {
  // The parent resource where this sandwich will be created.
  // Format: meals/{meal}
  string parent = 1 [
    (google.api.field_behavior) = REQUIRED,
    (google.api.resource_reference) = {
      child_type: "library.googleapis.com/Sandwich"
    }];

  // The sandwich to create.
  Sandwich sandwich = 2 [(google.api.field_behavior) = REQUIRED];
}
```

In an API design that considers the sandwich the "object of the RPC" for the CreateSandwich method, wouldn't we expect this to be the request? Specifically, like UpdateSandwichRequest which acts directly on a sandwich.

```
message CreateSandwichRequest {
  // The sandwich to create.
  //
  // The sandwich's `name` field is used to identify the sandwich to create.
  // Format: meals/{meal}/sandwiches/{sandwich}
  Sandwich sandwich = 1 [(google.api.field_behavior) = REQUIRED];
}
```

>Regardless, I think FailedPrecondition is a great representation of all the ways that the necessary conditions for sandwich-making could be absent.

Do you think that NOT_FOUND is _not_ a great representation of the necessary parent resource _not being found_? Or are you just arguing that FAILED_PRECONDITION is a better status code for this situation? 

To be perfectly clear, my stance is that both of these seem to be valid choices, but NOT_FOUND is much more of a standard/conventional/unsurprising choice than FAILED_PRECONDITION.

timburks:
Update: this is on hold pending a broader discussion of existing patterns and style.

PR #794, opened 2022-10-20
parent spec artifacts on spec revisions

### LINKED ISSUE #774: Registry API: Supporting artifacts on resource revisions
This discussion started when we were thinking about supporting a history of Scores. As we are building more UI features, it is getting more and more difficult to keep track of which spec revision a particular artifact (conformance report, Score, scorecard) belongs to. This is a limitation in conveying the correct information.

Can we evaluate the option of attaching artifacts to a resource's revision?

**Motivation:**

In our case, we will attach artifacts to SpecRevisions and DeploymentRevisions and keep Apis, Versions, Project as is (resources with no revisions).
This goes in line with our data model, the artifacts that get generated for specs are usually properties of a particular revision (lint reports, scores, etc), same with deployments (security analysis of a particular deployed state). The changes to these artifacts are also usually governed by changes in the revision.
For other generic properties, like Related links, Source code repo, those are usually higher level properties outside the scope of a spec/deployment which will normally get attached to an API. For an API, spec becomes the implementation detail of that API (and deployment becomes an operational detail), hence we predict that most of the characteristics related artifacts will be attached at the API level since that is the unit that users are managing in the registry.
Such artifacts at API level will not have to worry about revisions, they will be one time artifacts which will be preserved whereas for Specs and Deployments, artifacts will need a regeneration every time a new revision gets created.
Following [AIP-162](https://google.aip.dev/162#child-resources), the implementation will look like this:

The example is for Specs here, deployments will also have a similar behavior:

- **Create/Update:** Users should create artifacts only under a SpecRevision.
   - The artifact path will look like this: 
   `projects/demo/locations/global/apis/petstore/versions/v1/specs/openapi.yaml@0000/artifacts/conformance-report`
   - The parent of the created artifact must always include the revision ID of the spec.
   - When a user creates an artifact without specifying the revision ID, it should always get attached to the latest revision.
   - The response from the API will always include the revision ID in the artifact path.
- **List/Read:** Users should be able to list artifacts under SpecRevision and for a particular Spec
   - A list request like `projects/demo/locations/global/versions/v1/specs/openapi.yaml@0000/artifacts/-` should list all artifacts 
   under the specified revision.
   - A list request without the revision ID, `projects/demo/locations/global/versions/v1/specs/openapi.yaml/artifacts/-` should list 
   artifacts under the latest revision of that spec.
   - A list request with wildcard, `projects/demo/locations/global/versions/v1/specs/openapi.yaml@-/artifacts/-` should list all artifacts under all revisions of that spec.
   - The response from the API will always include the revision ID in the artifact path.
 
The above implementation is inline with the second scenario explained [here](https://google.aip.dev/162#child-resources)

**Some caveats to call out:**

- Whenever a user attaches an artifact to a spec without specifying the revision, it will be attached to the latest revision, that might have been updated by someone else and might not be the same as the user had created before.
- Users can no longer create artifacts under a spec, they will always be attached to a specific revision.
- Users should be aware that when they create a new revision of spec, the list call projects/demo/locations/global/versions/v1/specs/openapi.yaml/artifacts/- will no longer return artifacts that were available on the previous revision, they will have to create new artifacts on the latest revision for the list call to return artifacts.

theganyo:
Partially addresses #774 by moving artifacts from Specs to Spec Revisions. Deployment Revisions will be addressed similarly in a following PR.

### DIFF
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -23,6 +23,7 @@ import (
 	"github.com/apigee/registry/cmd/registry/patterns"
 	"github.com/apigee/registry/log"
 	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/registry/names"
 )
 
 type Action struct {
@@ -204,9 +205,13 @@ func generateUpdateActions(
 		// if the pattern doesn't explicitly ask for revisions, ignore them
 		if !includeRevisions {
 			switch r := pn.(type) {
-			case patterns.SpecName:
-				r.RevisionID = ""
-				pn = r
+			case patterns.SpecRevisionName:
+				s, err := names.ParseSpec(r.Spec())
+				if err != nil {
+					log.Errorf(ctx, "%s", err)
+					continue
+				}
+				pn = patterns.SpecName{Name: s}
 			}
 		}
 		visited[pn.String()] = true
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -233,10 +233,11 @@ func generateUpdateActions(
 // The filter excludes resources whose `name` property is equal to any of the visited parent names.
 //
 // For example, consider a visited map of parents which are apis:
-// {
-//     "projects/demo/locations/global/apis/example-api1": true,
-//     "projects/demo/locations/global/apis/example-api2": true,
-// }
+//
+//	{
+//	    "projects/demo/locations/global/apis/example-api1": true,
+//	    "projects/demo/locations/global/apis/example-api2": true,
+//	}
 //
 // The resulting CEL filter will be:
 // ["projects/demo/locations/global/apis/example-api1","projects/demo/locations/global/apis/example-api2"].all(parent, !(name==parent))
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -23,7 +23,6 @@ import (
 	"github.com/apigee/registry/cmd/registry/patterns"
 	"github.com/apigee/registry/log"
 	"github.com/apigee/registry/rpc"
-	"github.com/apigee/registry/server/registry/names"
 )
 
 type Action struct {
@@ -196,25 +195,9 @@ func generateUpdateActions(
 		return nil, nil, err
 	}
 
-	includeRevisions := strings.Contains(resourcePattern, "@")
-
 	// Iterate over a list of existing target resources to generate update actions
 	for _, targetResource := range resourceList {
-		pn := targetResource.ResourceName().ParentName()
-
-		// if the pattern doesn't explicitly ask for revisions, ignore them
-		if !includeRevisions {
-			switch r := pn.(type) {
-			case patterns.SpecRevisionName:
-				s, err := names.ParseSpec(r.Spec())
-				if err != nil {
-					log.Errorf(ctx, "%s", err)
-					continue
-				}
-				pn = patterns.SpecName{Name: s}
-			}
-		}
-		visited[pn.String()] = true
+		visited[targetResource.ResourceName().ParentName().String()] = true
 
 		takeAction, err := needsUpdate(
 			targetResource.ResourceName(),
@@ -264,7 +247,8 @@ func excludeVisitedParents(v map[string]bool) string {
 	// Wrap each string with quotes and join them with commas to build a JSON string array.
 	jsonStrings := make([]string, 0, len(v))
 	for parent := range v {
-		jsonStrings = append(jsonStrings, fmt.Sprintf("%q", parent))
+		// filter ignores revisions
+		jsonStrings = append(jsonStrings, fmt.Sprintf("%q", strings.Split(parent, "@")[0]))
 	}
 	return fmt.Sprintf("[%s].all(parent, !(name==parent))", strings.Join(jsonStrings, ","))
 }
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -195,9 +195,21 @@ func generateUpdateActions(
 		return nil, nil, err
 	}
 
+	includeRevisions := strings.Contains(resourcePattern, "@")
+
 	// Iterate over a list of existing target resources to generate update actions
 	for _, targetResource := range resourceList {
-		visited[targetResource.ResourceName().ParentName().String()] = true
+		pn := targetResource.ResourceName().ParentName()
+
+		// if the pattern doesn't explicitly ask for revisions, ignore them
+		if !includeRevisions {
+			switch r := pn.(type) {
+			case patterns.SpecName:
+				r.RevisionID = ""
+				pn = r
+			}
+		}
+		visited[pn.String()] = true
 
 		takeAction, err := needsUpdate(
 			targetResource.ResourceName(),
--- a/cmd/registry/controller/controller_test.go
+++ b/cmd/registry/controller/controller_test.go
@@ -623,7 +623,7 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
 		generatedResourceRequiresRevision bool
 	}{
 		{
-			desc: "create artifacts",
+			desc: "create latest spec revision artifacts",
 			seed: []seeder.RegistryResource{
 				&rpc.Artifact{
 					Name:     "projects/controller-test/locations/global/artifacts/registry-styleguide",
@@ -659,7 +659,7 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
 			},
 		},
 		{
-			desc: "outdated artifacts",
+			desc: "outdated lastest spec revision artifacts",
 			seed: []seeder.RegistryResource{
 				&rpc.Artifact{
 					Name: "projects/controller-test/locations/global/apis/petstore/versions/1.0.0/specs/openapi.yaml/artifacts/conformance-registry-styleguide",
@@ -697,7 +697,7 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
 			generatedResourceRequiresRevision: true,
 		},
 		{
-			desc: "missing dependencies",
+			desc: "missing latest spec revision dependencies",
 			seed: []seeder.RegistryResource{
 				&rpc.Artifact{
 					Name:     "projects/controller-test/locations/global/artifacts/registry-styleguide",
--- a/cmd/registry/controller/controller_test.go
+++ b/cmd/registry/controller/controller_test.go
@@ -792,8 +792,10 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
 						SpecID:    a.SpecID(),
 					}
 					if err := core.GetSpec(ctx, registryClient, sr, false, func(s *rpc.ApiSpec) error {
+						test.want[i].Command = strings.ReplaceAll(test.want[i].Command,
+							fmt.Sprintf("/%s", a.SpecID()), fmt.Sprintf("/%s@%s", a.SpecID(), s.GetRevisionId()))
 						test.want[i].GeneratedResource = strings.ReplaceAll(test.want[i].GeneratedResource,
-							fmt.Sprintf("/%s/", a.SpecID()), fmt.Sprintf("/%s@%s/", a.SpecID(), s.GetRevisionId()))
+							fmt.Sprintf("/%s", a.SpecID()), fmt.Sprintf("/%s@%s", a.SpecID(), s.GetRevisionId()))
 						return nil
 					}); err != nil {
 						t.Fatal("Failed GetSpecRevision", err)
--- a/cmd/registry/controller/controller_test.go
+++ b/cmd/registry/controller/controller_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	"github.com/apigee/registry/cmd/registry/core"
+	"github.com/apigee/registry/gapic"
 	"github.com/apigee/registry/pkg/connection"
 	"github.com/apigee/registry/pkg/connection/grpctest"
 	"github.com/apigee/registry/rpc"
@@ -166,6 +167,7 @@ func TestArtifacts(t *testing.T) {
 				},
 			}
 			actions := ProcessManifest(ctx, lister, projectID, manifest, 10)
+			addSpecRevisions(t, ctx, registryClient, test.want)
 
 			if diff := cmp.Diff(test.want, actions, sortActions); diff != "" {
 				t.Errorf("ProcessManifest(%+v) returned unexpected diff (-want +got):\n%s", manifest, diff)
@@ -263,6 +265,7 @@ func TestAggregateArtifacts(t *testing.T) {
 				},
 			}
 			actions := ProcessManifest(ctx, lister, projectID, manifest, 10)
+			addSpecRevisions(t, ctx, registryClient, test.want)
 
 			if diff := cmp.Diff(test.want, actions, sortActions); diff != "" {
 				t.Errorf("ProcessManifest(%+v) returned unexpected diff (-want +got):\n%s", manifest, diff)
@@ -405,6 +408,7 @@ func TestDerivedArtifacts(t *testing.T) {
 				},
 			}
 			actions := ProcessManifest(ctx, lister, projectID, manifest, 10)
+			addSpecRevisions(t, ctx, registryClient, test.want)
 
 			if diff := cmp.Diff(test.want, actions, sortActions); diff != "" {
 				t.Errorf("ProcessManifest(%+v) returned unexpected diff (-want +got):\n%s", manifest, diff)
@@ -498,6 +502,7 @@ func TestReceiptArtifacts(t *testing.T) {
 				},
 			}
 			actions := ProcessManifest(ctx, lister, projectID, manifest, 10)
+			addSpecRevisions(t, ctx, registryClient, test.want)
 
 			if diff := cmp.Diff(test.want, actions, sortActions); diff != "" {
 				t.Errorf("ProcessManifest(%+v) returned unexpected diff (-want +got):\n%s", manifest, diff)
@@ -606,6 +611,7 @@ func TestReceiptAggArtifacts(t *testing.T) {
 				},
 			}
 			actions := ProcessManifest(ctx, lister, projectID, manifest, 10)
+			addSpecRevisions(t, ctx, registryClient, test.want)
 
 			if diff := cmp.Diff(test.want, actions, sortActions); diff != "" {
 				t.Errorf("ProcessManifest(%+v) returned unexpected diff (-want +got):\n%s", manifest, diff)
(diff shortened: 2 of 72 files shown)


timburks:
I thought it would be useful to exercise this by bulk-uploading discovery specs twice (to get two revisions of every spec) and then attaching some artifacts, but while running the bulk uploader the second time (with a postgres backend), I got a flood of these errors:
```
 INFO[0293] Handling request.         method=UpdateApiVersion request_id=25d73b36 uid=91a27cd8
 ERROR[0293] Aborted.                  duration=1.554354ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=25d73b36 status_code=Aborted uid=91a27cd8
  INFO[0293] Handling request.         method=UpdateApiVersion request_id=deaa6df6 uid=91a27cd8
 ERROR[0293] Aborted.                  duration=1.610005ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=deaa6df6 status_code=Aborted uid=91a27cd8
  INFO[0296] Handling request.         method=UpdateApiVersion request_id=705adf3f uid=91a27cd8
 ERROR[0296] Aborted.                  duration=1.523023ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=705adf3f status_code=Aborted uid=91a27cd8
  INFO[0296] Handling request.         method=UpdateApiVersion request_id=ec09fc4c uid=91a27cd8
 ERROR[0296] Aborted.                  duration=1.598264ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=ec09fc4c status_code=Aborted uid=91a27cd8
  INFO[0296] Handling request.         method=UpdateApiVersion request_id=1189ee48 uid=91a27cd8
 ERROR[0296] Aborted.                  duration=1.520873ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=1189ee48 status_code=Aborted uid=91a27cd8
  INFO[0296] Handling request.         method=UpdateApiVersion request_id=1fc28347 uid=91a27cd8
 ERROR[0296] Aborted.                  duration=1.553934ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=1fc28347 status_code=Aborted uid=91a27cd8
  INFO[0297] Handling request.         method=UpdateApiVersion request_id=8376b607 uid=91a27cd8
 ERROR[0297] Aborted.                  duration=1.522513ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=8376b607 status_code=Aborted uid=91a27cd8
  INFO[0297] Handling request.         method=UpdateApiVersion request_id=38814ce5 uid=91a27cd8
 ERROR[0297] Aborted.                  duration=1.646625ms error=rpc error: code = AlreadyExists desc = pq: duplicate key value violates unique constraint "versions_pkey" method=UpdateApiVersion request_id=38814ce5 status_code=Aborted uid=91a27cd8
```

cmd/registry/controller/controller_test.go
```diff
@@ -774,6 +778,30 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
...
+					a, err := names.ParseArtifact(gr)
+					if err != nil {
+						t.Fatal("Failed to parse GeneratedResource", err)
+					}
+					sr := names.SpecRevision{
+						ProjectID:  a.ProjectID(),
+						ApiID:      a.ApiID(),
+						VersionID:  a.VersionID(),
+						SpecID:     a.SpecID(),
+						RevisionID: a.RevisionID(),
```
shrutiparabgoogle, line 793: This flow is a little confusing. We parse the `GeneratedResource` (which doesn't have a revisionID) into a `SpecRevision`, fetch the `SpecRevision` and use that revisionID to populate the `GeneratedResource`.

Specially the `RevisionID: a.RevisionID()` is quite confusing. Can we instead parse the `GeneratedResource` into a `Spec` -> fetch that spec and then use the spec's revision ID to populate the `GeneratedResource`. This type of flow will avoid the confusion I think.

cmd/registry/patterns/resources.go
```diff
@@ -52,6 +54,10 @@ func (s SpecName) Spec() string {
 	return s.Name.String()
 }
 
+func (s SpecName) Revision() string {
```
shrutiparabgoogle, line 57: This method can be unexported because it won't be accessible from the `ResourceName` interface.

cmd/registry/patterns/resources.go
```diff
@@ -321,6 +331,10 @@ func (ar ArtifactName) ParentName() ResourceName {
 		return SpecName{
 			Name: spec,
 		}
+	} else if rev, err := names.ParseSpecRevision(parent); err == nil {
+		return SpecName{
+			Name: rev.Spec(),
+		}
```
shrutiparabgoogle, line 337: Shouldn't this be:

```
return SpecName{
			Name: rev.Spec(),
                         RevisionID: rev.RevisionID,
		}

```


Should we also support revision collections in the previous block for matching collections?
Example: `projects/demo/locations/global/apis/-/versions/-/specs/-@-/artifacts/-` 
Not sure if the names package supports this yet.

theganyo:
> I thought it would be useful to exercise this by bulk-uploading discovery specs twice (to get two revisions of every spec) and then attaching some artifacts, but while running the bulk uploader the second time (with a postgres backend), I got a flood of these errors

Are you saying this is a new issue? Or just an observation you happened to perform on this PR?

cmd/registry/controller/controller_test.go
theganyo: Sounds good.

cmd/registry/patterns/resources.go
theganyo: ok

cmd/registry/patterns/resources.go
theganyo: Looking into this... Adding RevisionID to this breaks [this controller test](https://github.com/apigee/registry/blob/1777bc258f5719f80dd4cb9d3b8dd4fa285c78de/cmd/registry/controller/controller_test.go#L659) - it generates double the actions: a set of update actions for SpecRevisions and a set of create actions for Specs. So, I'm working to understand that.

timburks:
| Are you saying this is a new issue? Or just an observation you happened to perform on this PR?

I believe this was #795. If #796 is merged into this PR (it's in main), I think it will be fixed.

theganyo:
> I believe this was #795. If #796 is merged into this PR (it's in main), I think it will be fixed.

done

theganyo:
Per offline meeting with @shrutiparabgoogle, controller will continue to support existing syntax for Pattern and `$resource.spec`, no wildcard revisions or revision resource references. However, the generated `controller compute` actions must include the revision in `$resource.spec`. From that, the following are TODOs:

- [ ] Ensure manifest pattern cannot include revisions
- [ ] Ensure generated `controller compute` actions include the revision.

cmd/registry/controller/controller.go
```diff
@@ -195,9 +195,21 @@ func generateUpdateActions(
...
+
+		// if the pattern doesn't explicitly ask for revisions, ignore them
+		if !includeRevisions {
+			switch r := pn.(type) {
+			case patterns.SpecName:
+				r.RevisionID = ""
+				pn = r
+			}
+		}
+		visited[pn.String()] = true
```
shrutiparabgoogle, line 217: I think what we need here is ignoring the revision in the parent name while generating the filter (since the filter applies only on the resource name). 

So instead of this, we can make the change in `excludeVisitedParents` like this:
```
func excludeVisitedParents(v map[string]bool) string {
	// Wrap each string with quotes and join them with commas to build a JSON string array.
        // Ignore revision names if any exist in the parent names.
	jsonStrings := make([]string, 0, len(v))
	for parent := range v {
		jsonStrings = append(jsonStrings, fmt.Sprintf("%q", strings.Split(parent, "@")[0]))
	}
	return fmt.Sprintf("[%s].all(parent, !(name==parent))", strings.Join(jsonStrings, ","))
}
```

Thoughts?

theganyo: Yes, that will work - as long as we're ok dropping the guard to check for revision in the pattern, which we don't need now anyway.

shrutiparabgoogle: There is a function to validate the manifest definition: https://github.com/apigee/registry/blob/main/cmd/registry/controller/parser.go#L38

`ParseResource()` doesn't support `SpecRevisions` right now, so that guard is still in place, if we decide to add the specrevision support later, we can modify the validate function accordingly.

cmd/registry/controller/controller_test.go
```diff
@@ -774,6 +778,31 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
 			}
 			actions := ProcessManifest(ctx, lister, projectID, manifest, 10)
 
+			if test.generatedResourceRequiresRevision {
```
shrutiparabgoogle, line 781: I am not sure if we need this flag, wouldn't all generatedResource should include revision IDs in them? Since we are saying that we want to carry through a spec revision till the action.

IMO every reference of a spec should have a revision ID to it, we can't rely on `projects/demo/locations/apis/petstore/versions/v1/specs/openapi.yaml` anywhere because the latest revision might change anytime while the controller is processing the sate and taking actions. Every spec should include the revision ID in the controller.

cmd/registry/controller/controller_test.go
```diff
@@ -656,7 +659,7 @@ func TestMultipleEntitiesArtifacts(t *testing.T) {
 			},
 		},
 		{
-			desc: "outdated artifacts",
+			desc: "outdated lastest spec revision artifacts",
 			seed: []seeder.RegistryResource{
 				&rpc.Artifact{
 					Name: "projects/controller-test/locations/global/apis/petstore/versions/1.0.0/specs/openapi.yaml/artifacts/conformance-registry-styleguide",
```
shrutiparabgoogle, line 665: These spec level artifacts should also include revisionID in them.

cmd/registry/controller/list.go
```diff
@@ -14,6 +14,7 @@ type listingClient interface {
 	ListAPIs(context.Context, names.Api, string, core.ApiHandler) error
 	ListVersions(context.Context, names.Version, string, core.VersionHandler) error
 	ListSpecs(context.Context, names.Spec, string, core.SpecHandler) error
+	ListSpecRevisions(context.Context, names.SpecRevision, string, core.SpecHandler) error
```
shrutiparabgoogle, line 17: Are we using the ListSpecRevisions anywhere? I think Revision support should be implicitly added to the existing Specs support that we have in the controller.

cmd/registry/controller/parser.go
```diff
@@ -61,6 +61,8 @@ func validateGeneratedResourceEntry(parent string, generatedResource *rpc.Genera
 			return resourceName.Version() != ""
 		case "spec":
 			return resourceName.Spec() != ""
+		case "specrev":
```
shrutiparabgoogle, line 64: We are not supporting `$resource.specrev` for now. 

I think if we want to support revisions in some way, we can maybe support `$resource.spec@latest` instead of `specrev`, because which revision to pick will again be a question. This can be a discussion outside the scope of this PR though.

cmd/registry/controller/parser.go
```diff
@@ -169,7 +171,10 @@ func generateCommand(action string, resourceName string) (string, error) {
 		case "version":
 			entityVal = resource.Version()
 		case "spec":
-			entityVal = resource.Spec()
+			entityVal = resource.SpecRevision()
```
shrutiparabgoogle, line 174: I thought we were not going to add a `SpecRevision` type and going to stick to only `Specs`

cmd/registry/patterns/resources.go
```diff
@@ -82,6 +87,52 @@ func (s SpecName) ParentName() ResourceName {
 	return VersionName{}
 }
 
+type SpecRevisionName struct {
```
shrutiparabgoogle, line 90: Let's not support both Spec and SpecRevision, that will create a lot of confusion.

shrutiparabgoogle (commented):
I have reservations on how the controller modifications are made right now, let's sync on this. I will also add some comments on the issue to explain my thought process more clearly.

cmd/registry/controller/parser.go
theganyo: Sorry, yes, this was done before we discussed not supporting SpecRevisions in the controller... which now it looks like we're going to discuss next week? I imagine I should leave this alone until then.

theganyo:
Sorry, yes, there is cleanup on work that was done before we discussed not supporting SpecRevisions in the controller... which now it looks like we're going to discuss next week? I imagine I should leave this alone until then.

theganyo:
I removed a bunch of the old stuff related to specrev on the client (cmd/*) and addressed much of the issues mentioned. Not all issues are resolved yet, though, as some commands are failing (see failing tests) because of Spec / SpecRevision naming complexity.

cmd/registry/patterns/resources.go
```diff
@@ -41,15 +41,16 @@ type ResourceName interface {
...
+	RevisionID string
 }
 
 func (s SpecName) Artifact() string {
 	return ""
 }
 
 func (s SpecName) Spec() string {
-	return s.Name.String()
+	return s.String()
```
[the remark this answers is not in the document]
theganyo: Yes, it does:  https://github.com/apigee/registry/blob/a171157aca4f0d47d09486c97d2c772ee7fa97a2/cmd/registry/patterns/resources.go#L68-L74

cmd/registry/patterns/parser.go
```diff
@@ -49,8 +49,8 @@ func parseResource(resourcePattern string) (ResourceName, error) {
 		return ApiName{Name: api}, nil
 	} else if version, err := names.ParseVersion(resourcePattern); err == nil {
 		return VersionName{Name: version}, nil
-	} else if spec, err := names.ParseSpec(resourcePattern); err == nil {
-		return SpecName{Name: spec}, nil
+	} else if spec, err := names.ParseSpecRevision(resourcePattern); err == nil {
```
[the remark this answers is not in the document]
theganyo: Just using `names.ParseSpecRevision` should always work as I made the revision part optional. I think your other questions are addressed here:

https://github.com/apigee/registry/blob/a171157aca4f0d47d09486c97d2c772ee7fa97a2/server/registry/actions_spec_revisions_test.go#L330

https://github.com/apigee/registry/blob/a171157aca4f0d47d09486c97d2c772ee7fa97a2/server/registry/actions_spec_revisions_test.go#L300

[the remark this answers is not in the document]
shrutiparabgoogle: Thanks, this makes sense.

cmd/registry/patterns/resources.go
```diff
@@ -213,17 +218,18 @@ func (ar ArtifactName) Artifact() string {
...
+		SpecID:     ar.Name.SpecID(),
+		RevisionID: ar.Name.RevisionID(),
 	}
 
 	// Validate the generated name
-	if spec, err := names.ParseSpec(specPattern.String()); err == nil {
+	if spec, err := names.ParseSpecRevision(specPattern.String()); err == nil {
 		return spec.String()
-	} else if _, err := names.ParseSpecCollection(specPattern.String()); err == nil {
+	} else if _, err := names.ParseSpecRevisionCollection(specPattern.String()); err == nil {
```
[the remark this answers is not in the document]
theganyo: Hmm. This looks like a bug that's not being hit in testing. I'll take a look.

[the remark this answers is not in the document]
theganyo: Since this is preexisting and unclear, I think this deserves a look independent of this PR, I'll create a new issue.

theganyo:
@shrutiparabgoogle I updated the [complexity.go](https://github.com/apigee/registry/pull/794/files#diff-a0e166107d9b4f3bb4d456333ba8f8b657bdd8d755485ba18061738bd768a4b0) cmd to work with specs and/or revisions. I assume I'll need to do the same with each of the following. True? Are there any others I'm missing?

- [ ] conformance
- [ ] lint
- [ ] lintStats
- [ ] references
- [ ] score
- [ ] scoreCard
- [ ] vocabulary

shrutiparabgoogle:
> @shrutiparabgoogle I updated the [complexity.go](https://github.com/apigee/registry/pull/794/files#diff-a0e166107d9b4f3bb4d456333ba8f8b657bdd8d755485ba18061738bd768a4b0) cmd to work with specs and/or revisions. I assume I'll need to follow that same pattern with each of the following. True? Are there any others I'm missing?
> 
> * [ ]  conformance
> * [ ]  lint
> * [ ]  lintStats
> * [ ]  references
> * [ ]  score
> * [ ]  scoreCard
> * [ ]  vocabulary

`lint` and `lintstats` should be deprecated soon since we have now replaced those with conformance and score commands. Everything else looks correct

theganyo:
The latest 2 commits should complete the controller support (aside from lint and lintStats, as you suggested). If you see anything I missed, please let me know.

shrutiparabgoogle:
In my testing, I noticed that the conformance command was failing because it needs an update here: https://github.com/apigee/registry/blob/main/cmd/registry/conformance/conformance-task.go#L117

Update to `ParseSpecRevision()`

theganyo:
> In my testing, I noticed that the conformance command was failing because it needs an update here: https://github.com/apigee/registry/blob/main/cmd/registry/conformance/conformance-task.go#L117
> 
> Update to `ParseSpecRevision()`

done

shrutiparabgoogle (approved):
Thanks for adding those parser tests, that gives me confidence that supporting revisions in controller manifest should be straightforward. 

This PR has LGTM from my side, after testing and verifying  the controller changes.

timburks:
One observation, which I'm not sure is new: when I list specs with `registry list`, they are in ascending order alphabetically, but when I list revisions, they seem to be in descending order, but perhaps first sorted by revision time and then alphabetically (both descending).

If possible, I think it would be best to sort alphabetically (ascending) by resource name and secondarily sort descending by revision time.
```
$ registry list apis/-/versions/-/specs/-
projects/disco/locations/global/apis/abusiveexperiencereport/versions/v1/specs/discovery.json
projects/disco/locations/global/apis/acceleratedmobilepageurl/versions/v1/specs/discovery.json
projects/disco/locations/global/apis/accessapproval/versions/v1/specs/discovery.json
projects/disco/locations/global/apis/accesscontextmanager/versions/v1/specs/discovery.json
...
```
$ registry list apis/-/versions/-/specs/-@-
projects/disco/locations/global/apis/youtube/versions/v3/specs/discovery.json@258ade1c
projects/disco/locations/global/apis/workflows/versions/v1/specs/discovery.json@16f9d38d
projects/disco/locations/global/apis/workflows/versions/v1beta/specs/discovery.json@d5ffe22e
projects/disco/locations/global/apis/versionhistory/versions/v1/specs/discovery.json@98aa8265
projects/disco/locations/global/apis/transcoder/versions/v1/specs/discovery.json@14f704d2
```

theganyo:
Added the existing data migration for spec revision artifacts

timburks (approved):
LGTM. Apart from tripping over unrelated FK issues, the only concern that I saw while testing this was list order. I mainly exercised "registry compute complexity" and that worked as I expected for both "specs/-" and "specs/-@-". So I think we can go ahead with this, and adding support for deployments can come after this is in.

PR #813, opened 2022-11-04
Add --jobs flag to registry compute subcommands.

timburks:


cmd/registry/cmd/compute/conformance_test.go
```diff
@@ -463,11 +463,11 @@ func TestConformance(t *testing.T) {
 			}
 
 			// setup the command
-			conformanceCmd := conformanceCommand()
+			conformanceCmd := Command()
 			conformanceCmd.Flags().String("filter", "", "Filter selected resources")
 			conformanceCmd.Flags().Bool("dry-run", false, "if set, computation results will only be printed and not uploaded to registry")
```
shrutiparabgoogle, line 468: These two flag definitions are not needed anymore since the initialization uses `Command()` which comes with those flags.

shrutiparabgoogle (approved):

cmd/registry/cmd/compute/conformance_test.go
timburks: Thanks! I've removed those two lines.

PR #816, opened 2022-11-04
Enhance UpdateSequence tests to verify create and update timestamps.

### LINKED ISSUE #792: Registry server: CreateTime being updated on Update actions
When updating resources, the CreateTime time property is being updated also. I don't know if this is the expected behavior, but this was not happening before. Looking into the source code, I've found that the actions related to update are calling a function models.New(). I verified that this happens for Project, Artifact, Api, and Version.

image
image

timburks:
This fixes #792 and verifies with tests. Projects, APIs, and versions were already setting timestamps correctly, and artifacts weren't (and are fixed). Although specs and deployments weren't mentioned in #792, this also adds similar tests for them, too.

### DIFF
--- a/server/registry/actions_apis_test.go
+++ b/server/registry/actions_apis_test.go
@@ -1156,6 +1156,9 @@ func TestUpdateApiSequence(t *testing.T) {
 	}
 	var createTime time.Time
 	var updateTime time.Time
+	// NOTE: in the following sequence of tests, each test depends on its predecessor.
+	// Resources are successively created and updated using the "Update" RPC and the
+	// tests verify that CreateTime/UpdateTime fields are modified appropriately.
 	for i, test := range tests {
 		t.Run(test.desc, func(t *testing.T) {
 			var result *rpc.Api
--- a/server/registry/actions_apis_test.go
+++ b/server/registry/actions_apis_test.go
@@ -1143,7 +1143,7 @@ func TestUpdateApiSequence(t *testing.T) {
 				Api: &rpc.Api{
 					Name: "projects/my-project/locations/global/apis/a",
 				},
-				AllowMissing: true,
+				AllowMissing: false,
 			},
 			want: codes.OK,
 		},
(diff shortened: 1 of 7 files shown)


server/registry/actions_apis_test.go
```diff
@@ -1153,11 +1154,29 @@ func TestUpdateApiSequence(t *testing.T) {
...
+				} else {
+					if !createTime.Equal(result.CreateTime.AsTime()) {
+						t.Errorf("UpdateApi create time changed after update (%v %v)", createTime, result.CreateTime.AsTime())
+					}
+					if !updateTime.Before(result.UpdateTime.AsTime()) {
+						t.Errorf("UpdateApi update time did not increase after update (%v %v)", updateTime, result.UpdateTime.AsTime())
+					}
+					updateTime = result.UpdateTime.AsTime()
+				}
+			}
```
seaneganx, line 1179: This seems like it belongs in a separate test. As a rule of thumb, any time you're hiding assertions inside an if-statement to only check them in a subset of the test cases, you're probably trying to squeeze more than one kind of behavioral assertion into a single test.

Here it's really tempting because "TestUpdateApiSequence" is a vague name, where many different tests are probably going to use a sequence of UpdateApi calls. It shouldn't be a catch-all for every type of test that does that though, and should be renamed instead.

timburks: Ah, ok, I think if you look more closely, you'll see that there's a dependency between these tests - they must run in sequence. I used the existing "test" structure to accomplish this, but could easily rename "test" and "tests" to "step" and "steps", and put the test.Run() outside the loop. WDYT?

timburks: OK, since these tests will be nonideal without a significant rewrite, I added comments that make their sequential dependency harder to miss and suggest we move forward. We're testing a relatively small feature here, though one that we *have* broken in the past, and I don't think it will be hard to figure out what we've broken if these tests ever fail again. (Happy to add other wording if  you want to more clearly say "these tests are bad, don't do this again!" :) )

seaneganx (approved):

PR #818, opened 2022-11-07
artifacts on deployment revisions

### LINKED ISSUE #774: Registry API: Supporting artifacts on resource revisions

theganyo:
Fixes #774


server/registry/actions_artifacts_test.go
```diff
@@ -1638,3 +1599,247 @@ func TestSpecRevisionArtifacts(t *testing.T) {
...
+				},
+				want: &rpc.ListArtifactsResponse{
+					Artifacts: []*rpc.Artifact{
+						{Name: "projects/my-project/locations/global/apis/my-api/deployments/my-deployment2@" + deployment2r1.GetRevisionId() + "/artifacts/my-artifact-r1"},
+						{Name: "projects/my-project/locations/global/apis/my-api/deployments/my-deployment@" + deployment1r2.GetRevisionId() + "/artifacts/my-artifact-r2"},
+					},
+				},
+			},
+			{
+				desc: "all revisions of all deployment",
```
timburks, line 1783: super-nit: "deployments"

timburks (approved):

PR #820, opened 2022-11-08
correct default ordering for all types

### LINKED ISSUE #807: Adjust sort order of spec revisions
When I list specs with `registry list`, they are in ascending order alphabetically, but when I list revisions, they seem to be in descending order, but perhaps first sorted by revision time and then alphabetically (both descending).

If possible, I think it would be best to sort alphabetically (ascending) by resource name and secondarily sort descending by revision time.

```
$ registry list apis/-/versions/-/specs/-
projects/disco/locations/global/apis/abusiveexperiencereport/versions/v1/specs/discovery.json
projects/disco/locations/global/apis/acceleratedmobilepageurl/versions/v1/specs/discovery.json
projects/disco/locations/global/apis/accessapproval/versions/v1/specs/discovery.json
projects/disco/locations/global/apis/accesscontextmanager/versions/v1/specs/discovery.json
...
```

```
$ registry list apis/-/versions/-/specs/-@-
projects/disco/locations/global/apis/youtube/versions/v3/specs/discovery.json@258ade1c
projects/disco/locations/global/apis/workflows/versions/v1/specs/discovery.json@16f9d38d
projects/disco/locations/global/apis/workflows/versions/v1beta/specs/discovery.json@d5ffe22e
projects/disco/locations/global/apis/versionhistory/versions/v1/specs/discovery.json@98aa8265
projects/disco/locations/global/apis/transcoder/versions/v1/specs/discovery.json@14f704d2
...
```

_Originally posted by @timburks in https://github.com/apigee/registry/issues/794#issuecomment-1298966661_

theganyo:
Fixes #807

### DIFF
--- a/server/registry/internal/storage/list.go
+++ b/server/registry/internal/storage/list.go
@@ -27,6 +27,26 @@ import (
 	"gorm.io/gorm"
 )
 
+var tableFieldsLookup = map[string]map[string]filtering.FieldType{
+	"projects":             projectFields,
+	"apis":                 apiFields,
+	"versions":             versionFields,
+	"specs":                specFields,
+	"deployments":          deploymentFields,
+	"artifacts":            artifactFields,
+	"revisioned_artifacts": revisionedArtifactFields,
+}
+
+var defaultOrder = map[string]string{
+	"projects":             "project_id",
+	"apis":                 "project_id, api_id",
+	"deployments":          "project_id, api_id, deployment_id, revision_create_time desc",
+	"versions":             "project_id, api_id, version_id",
+	"specs":                "project_id, api_id, version_id, spec_id, revision_create_time desc",
+	"artifacts":            "project_id, api_id, version_id, spec_id, deployment_id, artifact_id, create_time desc",
+	"revisioned_artifacts": "project_id, api_id, version_id, spec_id, deployment_id, revision_create_time desc, artifact_id",
+}
+
 var projectFields = map[string]filtering.FieldType{
 	"name":         filtering.String,
 	"project_id":   filtering.String,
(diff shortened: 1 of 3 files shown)


timburks (approved):
Slick! :)

server/registry/internal/storage/list.go
```diff
@@ -99,24 +119,45 @@ var deploymentFields = map[string]filtering.FieldType{
...
 // gormOrdering accepts a user-specified order_by string and returns a gorm-compatible equivalent.
 // For example, the user-specified string `name,description` returns `key,description`.
-// An error is returned if the string is invalid or refers to a field that isn't included in the provided `fields` map.
-func gormOrdering(ordering string, fields map[string]filtering.FieldType) (string, error) {
+// An error is returned if the string is invalid or refers to a field that isn't included in the `fields` map.
+func gormOrdering(ordering, table string) (string, error) {
+	fields, ok := tableFieldsLookup[table]
+	if !ok {
+		return "", status.Errorf(codes.Internal, "unknown order table: %q", table)
+	}
```
seaneganx, line 158: This failure mode seems unnecessary to me. Each caller of gormOrdering should be able to provide the data at compile time, which avoids using lookup tables (and having to handle these errors) at runtime. Or am I missing something?

theganyo: You're right, good catch. A previous iteration required the table name for namespacing, but I neglected to back this out after I was able to eliminate that requirement.

theganyo: Both the fields and default order are accessed by table name, so it didn't make sense to pass in both (opening a new failure mode if they didn't match). If you'd rather, I can make the table name an enum instead of a string... I had considered that, but it just didn't seem to add a lot of value for an purely internal lookup.

seaneganx (approved):

PR #845, opened 2022-11-17
Add the original resource to patterns.ResourceInstance

shrutiparabgoogle:
This the preparation for enabling the following feature in the scoring tool:

Currently, a score can be derived only from the artifacts that are attached to the primary entities (apis, versions, specs, deployments). However, we would also like the ability to derive a score from the primary entity itself, an example of this is the completeness score for an API, how well hydrated (with information) is a particular API stored in the registry.

### DIFF
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -286,7 +286,7 @@ func generateCreateActions(
 		}
 		parentList = []patterns.ResourceInstance{
 			patterns.ProjectResource{
-				Project: &rpc.Project{Name: parentName.String()},
+				ProjectName: parentName.String(),
 			},
 		}
 
--- a/cmd/registry/controller/controller.go
+++ b/cmd/registry/controller/controller.go
@@ -286,7 +286,7 @@ func generateCreateActions(
 		}
 		parentList = []patterns.ResourceInstance{
 			patterns.ProjectResource{
-				ProjectName: parentName,
+				Project: &rpc.Project{Name: parentName.String()},
 			},
 		}
 
--- a/cmd/registry/controller/list.go
+++ b/cmd/registry/controller/list.go
@@ -49,7 +49,7 @@ func listResources(ctx context.Context, client listingClient, pattern, filter st
 	} else if spec, err := names.ParseSpecCollection(pattern); err == nil {
 		err2 = client.ListSpecs(ctx, spec, filter, generateSpecHandler(&result))
 	} else if artifact, err := names.ParseArtifactCollection(pattern); err == nil {
-		err2 = client.ListArtifacts(ctx, artifact, filter, false, generateArtifactHandler(&result))
+		err2 = client.ListArtifacts(ctx, artifact, filter, true, generateArtifactHandler(&result))
 	}
 
 	// Then try to match resource names.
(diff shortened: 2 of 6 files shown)


cmd/registry/controller/controller.go
```diff
@@ -286,7 +286,7 @@ func generateCreateActions(
 		}
 		parentList = []patterns.ResourceInstance{
 			patterns.ProjectResource{
-				ProjectName: parentName,
+				Project: &rpc.Project{Name: parentName.String()},
```
timburks, line 289: Note that "Project" is only defined in the Admin service.

theganyo:
I'm likely missing something, but it appears that the ResourceInstance implementations (eg. SpecResource, ApiResource, etc) are now wrappers without a purpose?

cmd/registry/controller/controller.go
shrutiparabgoogle: I understand the confusion, `ProjectResource` is only used in this particular instance and we are mainly interested in it's name. At no time is the `rpc.Project` resource being fetched from the registry, since the controller is the registry client command.

To mimimize the confusion, I removed the `rpc.Project` resource from the struct, we still need to keep the timestamp function (which returns null value), for `ProjectResource` to be able to inherit `ResourceInstance`.

See the latest commit `Handle ProjectResource properly` for the fix I made.

shrutiparabgoogle:
> I'm likely missing something, but it appears that the ResourceInstance implementations (eg. SpecResource, ApiResource, etc) are now wrappers without a purpose?

The `ResourceInstance` is the abstraction which the controller and also the scoring commands use to represent the resources in the registry. For these commands, everything is an instance of `ResourceInstance` in the registry and they are only interested in their timestamps and names (for now). 

These wrappers make sure that the different resource types in the registry can be represented as the `ResourceInstance`s for these commands. 

Now that we are interested in the actual resource as well (only in a few occasions), this change is storing the original resource in these wrappers but still behaves as a `ResourceInstance`.  

Does this explanation help? I am open to other implementation approaches that I might have missed.

theganyo:
> > I'm likely missing something, but it appears that the ResourceInstance implementations (eg. SpecResource, ApiResource, etc) are now wrappers without a purpose?
> 
> Does this explanation help? I am open to other implementation approaches that I might have missed.

Thanks. I understood what was happening and the need for a common interface, I just didn't see a need for the wrappers... the key bit of functionality that this provides seems to be the ResourceName implementations. I was imagining this could've simply been supported directly in the names package instead of an additional set of wrappers. But perhaps not.

theganyo (approved):

shrutiparabgoogle:
> I was imagining this could've simply been supported directly in the names package instead of an additional set of wrappers. But perhaps not.

I agree, I imagine eventually this interface going into the names package, but it still needs to mature out of specific use-cases and be generic enough to be included in names package. I am sure there will be a refactor on this part of the code, but that is outside the scope of this PR.

shrutiparabgoogle:
> > I was imagining this could've simply been supported directly in the names package instead of an additional set of wrappers. But perhaps not.
> 
> I agree, I imagine eventually this interface going into the names package, but it still needs to mature out of specific use-cases and be generic enough to be included in the `names` package.
> 
> The `names` package was developed after the introduction of the `ResourceName` and `ResourceInstance` interfaces. And the interfaces have changed a lot since then, we should look at consolidating the `names` package and `resource.go` in the future, thanks for calling this out.

Here is an issue for future scope: https://github.com/apigee/registry/issues/848

Issue #601, opened 2022-06-11
Registry server: tagApiSpecRevision and tagApiDeploymentRevision should allow implicit tagging of the latest revision


timburks:
Currently these two methods in the API require an explicit reference to a revision of the spec or deployment to be tagged -- i.e. the revision id must be included. When the revision id is omitted, the API returns an error like this one:
```
rpc error: code = InvalidArgument desc = invalid spec revision name "projects/apigeek-86/locations/global/apis/registry/versions/v1/specs/openapi.yaml": must match "^projects/([A-Za-z0-9-.]+)/locations/global/apis/([A-Za-z0-9-.]+)/versions/([A-Za-z0-9-.]+)/specs/([A-Za-z0-9-.]+)@([a-z0-9-]+)$"
```
Let's consider allowing the revision tag to be omitted, and when it is omitted, the operation will tag the latest revision of the specified spec or deployment. AIP-162 currently says "The name field should require an explicit revision ID to be provided" but I think this is debatable and worth discussing for the usability benefits of this. 

We might get this same capability with #600 if `@0` or `@1` is added as a shorthand for the latest revision id.

theganyo:
Fixed by #826

PR #847, opened 2022-11-21
Ensure that spec and revision tags contain no uppercase characters and are not "-".

### LINKED ISSUE #250: TagApiSpecRevision - mixed case tags

timburks:
Fixes #250.

### DIFF
--- a/server/registry/actions_deployment_revisions_test.go
+++ b/server/registry/actions_deployment_revisions_test.go
@@ -162,6 +162,51 @@ func TestTagApiDeploymentRevision(t *testing.T) {
 	})
 }
 
+func TestTagApiDeploymentRevisionResponseCodes(t *testing.T) {
+	tests := []struct {
+		desc string
+		tag  string
+		want codes.Code
+	}{
+		{
+			desc: "contains uppercase leters",
+			tag:  "TestTag",
+			want: codes.InvalidArgument,
+		},
+		{
+			desc: "single dash",
+			tag:  "-",
+			want: codes.InvalidArgument,
+		},
+		{
+			desc: "valid one-character tag",
+			tag:  "x",
+			want: codes.OK,
+		},
+		{
+			desc: "valid tag",
+			tag:  "latest",
+			want: codes.OK,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			ctx := context.Background()
+			server := defaultTestServer(t)
+			if err := seeder.SeedDeployments(ctx, server, &rpc.ApiDeployment{Name: "projects/my-project/locations/global/apis/a/deployments/d"}); err != nil {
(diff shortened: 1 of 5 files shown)


server/registry/actions_deployment_revisions_test.go
```diff
@@ -162,6 +162,51 @@ func TestTagApiDeploymentRevision(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			ctx := context.Background()
+			server := defaultTestServer(t)
+			if err := seeder.SeedDeployments(ctx, server, &rpc.ApiDeployment{Name: "projects/my-project/locations/global/apis/a/deployments/d"}); err != nil {
+				t.Fatalf("Setup/Seeding: Failed to seed registry: %s", err)
+			}
+			if _, err := server.TagApiDeploymentRevision(ctx, &rpc.TagApiDeploymentRevisionRequest{
+				Name: "projects/my-project/locations/global/apis/a/deployments/d",
```
timburks, line 201: @theganyo I noticed we have an open issue #601 about implicit tagging of the latest revision - which I believe is what is happening here (successfully)

theganyo (approved):

server/registry/actions_deployment_revisions_test.go
theganyo: Yes, and I just looked through the code to verify: The changes I made [in this PR](https://github.com/apigee/registry/pull/826) allow revision requests to use a non-revision name which will assume the latest revision.

PR #849, opened 2022-11-22
ensure filtering honors requested page size

### LINKED ISSUE #846: Filter resources using labels not working properly when having too many items
On my instance of the Registry server I added the label "source" on all resources of type API. If I try some filters like the one below, the result might not return all matches when having too many items in the database.

`registry rpc list-apis -j --page_size 1000 --parent projects/my-project-name/locations/global --filter 'labels.source=="example-source"'`

On this example, I have 530 APIs on the database, even with the explicit page_size flag, it's always returning 500 items, so, I need to make another call to get the other 30 APIs. I did not find any limitations about the page_size when using along with filters.

theganyo:
Fixes #846

### DIFF
--- a/server/registry/actions_apis_test.go
+++ b/server/registry/actions_apis_test.go
@@ -842,12 +842,13 @@ func TestListApisSequence(t *testing.T) {
 	}
 }
 
-// This test prevents the list sequence from ending before a known filter match is listed.
+// This test ensures filtering works correctly with paging and prevents the list
+// sequence from ending before a known filter match is listed.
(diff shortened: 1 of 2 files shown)


theganyo:
Note: I only added the tests for filter paging to the APIs collection, as this is enough to prove the issue and all collections use the same pattern. However, if you'd like, I can easily clone these tests to the other collections.

timburks (approved):
@theganyo thanks! These tests look great. Let's not add tests for any other resource types unless we find significant gaps in code coverage reports.

Issue #843, opened 2022-11-16
golangci-lint warnings about deprecated functions


timburks:
```
$ golangci-lint run
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649. 
examples/grpc-client/main.go:38:23: SA1019: grpc.WithInsecure is deprecated: use WithTransportCredentials and insecure.NewCredentials() instead. Will be supported throughout 1.x. (staticcheck)
                opts = append(opts, grpc.WithInsecure())
                                    ^
cmd/registry/scoring/extensions/extensions.go:44:3: SA1019: cel.Functions is deprecated: use Function() instead to declare the function, its overload signatures, and the overload implementations. (staticcheck)
                cel.Functions(
                ^
pkg/connection/client.go:34:42: SA1019: grpc.WithInsecure is deprecated: use WithTransportCredentials and insecure.NewCredentials() instead. Will be supported throughout 1.x. (staticcheck)
                conn, err := grpc.Dial(config.Address, grpc.WithInsecure())
```

timburks:
@shrutiparabgoogle could you address the remaining warning?
```
$ golangci-lint run
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649. 
cmd/registry/scoring/extensions/extensions.go:44:3: SA1019: cel.Functions is deprecated: use Function() instead to declare the function, its overload signatures, and the overload implementations. (staticcheck)
                cel.Functions(
                ^
```

PR #852, opened 2022-11-22
Replace deprecated `grpc.WithInsecure()`

timburks:
with `grpc.WithTransportCredentials(insecure.NewCredentials())`

https://stackoverflow.com/a/70482635

This is one of two fixes required to address #843

### DIFF
--- a/examples/grpc-client/main.go
+++ b/examples/grpc-client/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/apigee/registry/rpc"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
 )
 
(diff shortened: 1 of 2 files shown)


examples/grpc-client/main.go
```diff
@@ -26,16 +26,16 @@ import (
...
 )
 
 func main() {
 	// This makes a raw gRPC connection.
 	// see the connection package for a simpler way to get a Go client.
 	var opts []grpc.DialOption
-	insecure := os.Getenv("APG_REGISTRY_INSECURE")
-	if insecure != "" {
-		opts = append(opts, grpc.WithInsecure())
+	if os.Getenv("APG_REGISTRY_INSECURE") != "" {
```
theganyo, line 37: Nit and out of scope for this PR, but I believe it would be more bulletproof if we used `strconv.Boolean()` rather than assuming all non-blank values are true (eg. "false" "0").

theganyo (approved):

Issue #446, opened 2022-01-20
Add missing functionality to names package


seaneganx:
### Background
The [names package](https://github.com/apigee/registry/tree/4b43762e57f64242cad80b8c7d2e3add157fde2d/server/registry/names) supports all necessary functionality for the server to work with the names of resources. Its purpose is to abstract away from the strings containing resource names and the details of their expected format so developers can easily work with resource names.

While developing the registry controller, we encountered a situation where we needed to [work with resources names generically](https://github.com/apigee/registry/blob/4b43762e57f64242cad80b8c7d2e3add157fde2d/cmd/registry/controller/parser.go), but the existing names package wasn't designed to work nicely in generic cases. The controller package needed to [extend the existing name types](https://github.com/apigee/registry/blob/4b43762e57f64242cad80b8c7d2e3add157fde2d/cmd/registry/controller/resource-name.go) to fill the gaps.

### Ideas

1. Support full abstraction of resource names: an abstract type that can represent any resource.
2. Support partial abstraction of resource names: abstract types that can represent different kinds of resources (e.g. collections).
3. Add separate types to represent resource collections.
4. Improve identification of parent type when resource names contain a [wildcard](https://google.aip.dev/159#guidance).

timburks:
Would you be open to replacing the word "missing" in the title with "additional"?

seaneganx:
Agreed "missing" isn't the ideal word for it. I don't want to use "Add additional functionality to names package" as a title either though. It doesn't convey that there are gaps between what the package offers and what the needed use cases are.

timburks:
Makes sense, let's describe the missing use case and the impact of addressing it. "Adding generic name parsing to the names package would simplify some places where we parse names" ??

PR #859, opened 2022-12-02
Registry tool: improve `export yaml` to support all artifact types and to include artifact parent when needed

### LINKED ISSUE #856: Registry tool: include `parent` in exported yaml of non-top-level resources
`registry apply` supports a metadata field that specifies the parent of resources that are not top-level (i.e. directly under `project/location`). When exporting these resources with `registry export yaml`, we should include this field so that resources can be exported and reimported. To make exported resources portable across projects, these names should exclude the `project/location` segments. 

Related: this includes verification that all resource types can be exported and reimported using YAML.

timburks:
Fixes #856

`registry export yaml` now exports all resource types and accepts a `--nested` (`-n`) flag that includes child resources in a "nested" YAML representation of the resource (e.g. `--nested` with an API resource causes the versions, specs, deployments, and artifacts of that API to be included in a single file). Note that `--nested` now is *not* the default.

Exported resources include `parent` metadata when resource parents are not the project/location where the registry resides. `parent` fields exclude the project and location to allow exported resources to be easily moved from one project to another. Also, only the current revisions of specs and deployments are exported (there's no mechanism for importing a revision with a specified revision ID).  

This PR includes command-level tests that verify that `registry export yaml` can be called, but most of the functional verification is in the tests for `registry apply`, which import and reexport a registry and verify that representations match. #863 proposes test restructuring that would move the functional tests to the `patch` package, where YAML i/o is implemented.


cmd/registry/cmd/apply/testdata/sample/apis/registry.yaml
```diff
@@ -25,7 +25,7 @@ data:
         state: Staging
         specs:
           - metadata:
-              name: openapi.yaml
+              name: openapi
```
timburks, line 28: Since we export resources into YAML files, I think we want to avoid creating resources with names that include ".yaml" extensions. This was an arbitrary convention introduced by the OpenAPI bulk uploader that I think we can easily change.

server/registry/names/common.go
```diff
@@ -85,3 +85,27 @@ const Location = "global"
...
+// and any revision ID is removed since revision ids cannot be imported.
+//
+// Note that currently revision ids are only removed from trailing segments.
+// That's all that we need now, but support for removing internal revision ids
+// could be added if needed.
+//
+// Also note: this would be a nice function to add to the Name interface
+// but that might require improving the regularity of access to ProjectIDs
+// for different types of Names.
+func ExportableName(name string, projectID string) string {
```
theganyo, line 100: I'm not sure I understand why this can't be put into the Name interface... and why do we need to pass in the projectID? A name is well-structured on it's own... so we should be able to easily identify and remove the `projects/x` and `locations/x` segments, no?

cmd/registry/cmd/apply/apply_test.go
```diff
@@ -271,3 +272,172 @@ func TestApplyProject(t *testing.T) {
...
+		}
+		actual, _, err := patch.ExportArtifact(ctx, registryClient, message)
+		if err != nil {
+			t.Fatalf("ExportArtifact(%+v) returned an error: %s", message, err)
+		}
+		if diff := cmp.Diff(expected, actual); diff != "" {
+			t.Errorf("GetArtifact(%q) returned unexpected diff: (-want +got):\n%s", message, diff)
+		}
+	}
+	{
```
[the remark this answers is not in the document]
timburks: They have a test setup in common. I think I've seen preferences to not have tests share common setup, but I don't think it's a big change either way.

server/registry/names/common.go
timburks: > "so we should be able to easily identify and remove the projects/x and locations/x segments, no?"

Yes, true, I considered doing this but it seemed to go against the spirit of the `names` package, which I think is trying to encapsulate knowledge about name structure.

timburks: #867

cmd/registry/cmd/apply/apply_test.go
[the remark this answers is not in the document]
theganyo: I think I wasn't clear: You can run subtests with t.Run().

cmd/registry/cmd/export/yaml.go
```diff
@@ -28,21 +27,22 @@ import (
...
 		Use:   "yaml RESOURCE",
 		Short: "Export a subtree of the registry as YAML",
 		Args:  cobra.ExactArgs(1),
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
 			c, err := connection.ActiveConfig()
 			if err != nil {
-				log.FromContext(ctx).WithError(err).Fatal("Failed to get config")
+				return err
```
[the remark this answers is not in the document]
theganyo: Ok. There are many places that add text to the err message (nearly all) in logging, so we should try to be consistent as we move from log usage to RunE.

server/registry/names/common.go
theganyo: > seemed to go against the spirit of the names package, which I think is trying to encapsulate knowledge about name structure

This explanation leaves me more confused than before, but I trust that you've given it thought.

theganyo (approved):

cmd/registry/cmd/apply/apply_test.go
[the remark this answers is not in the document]
timburks: Sounds good, I'll wrap these in `t.Run()`

server/registry/names/common.go
timburks: [Here](https://github.com/apigee/registry/commit/b289e729991d0b67448278bdcda7cfbe3b1a01b9) is @seaneganx's original PR creating the `names` package. A goal of that was to reduce the amount of pasting together of resource names that was scattered throughout the code. Looking back, I think it has definitely been helpful, but we've also been aware that it was incomplete (#446).

timburks:
Merged. See #863, #865, and #867 for followup issues. I expect to address #860 next.

PR #861, opened 2022-12-05
Update field name for parent resource in logs

### LINKED ISSUE #857: Registry server: call logger prints misleading "collection" information that omits the collection name
[This code](https://github.com/apigee/registry/blob/fd90f2115f1eeb4e7f0b9fedb44e114185fbd33f/log/interceptor/interceptor.go#L52) is logging `parent` arguments with the `collection` key.

That produces log messages that look like this:
```
  INFO[0006] Handling request.         collection=projects/test/locations/global/apis/registry/versions/v1/specs/openapi.yaml method=CreateArtifact request_id=7b9d2cc9 uid=7d82bfb9
  INFO[0006] Handling request.         collection=projects/test/locations/global/apis/registry/versions/v1 method=CreateArtifact request_id=2899b7a7 uid=7d82bfb9
  INFO[0006] Handling request.         collection=projects/test/locations/global/apis/registry/deployments/prod method=CreateArtifact request_id=5d13ae75 uid=7d82bfb9
```

This appears incorrect because the `artifacts` collection name is omitted. It seems less confusing to log this with the `parent` key, which also matches the field in the request.

So I think we should replace
```
reqInfo["collection"] = r.GetParent()
```
with
```
reqInfo["parent"] = r.GetParent()
```

@seaneganx are you aware of anything that this change to logging might break?

seaneganx:
Fixes #857.

### DIFF
--- a/log/interceptor/interceptor.go
+++ b/log/interceptor/interceptor.go
@@ -49,7 +49,7 @@ func CallLogger(opts ...log.Option) grpc.UnaryServerInterceptor {
 		}
 
 		if r, ok := req.(collectionOperation); ok {
-			reqInfo["collection"] = r.GetParent()
+			reqInfo["parent"] = r.GetParent()
 		}
 
 		// Bind request-scoped and inbound attributes to the context logger before handling the request.


timburks (approved):

timburks:
@seaneganx thanks!

PR #868, opened 2022-12-09
Remove v1alpha1 protos and related code.

### LINKED ISSUE #541: Move v1alpha1 protos to registry-experimental
To avoid confusion and over-indexing on them, I think we should move the application/v1alpha1 protos to the registry-experimental repo. Anything in the registry tool that uses them should go to the registry-experimental tool, but without the protos, `registry get` won't be able to print their contents. Previously that seemed to me to be a blocker, but as we build out our supported application protos, I think these are needed less and less.

timburks:
If needed, this code can be added to registry-experimental (https://github.com/apigee/registry-experimental/issues/120).

Closes #541.


seaneganx (approved):
Have you confirmed that none of this is depended on by packages in registry-experimental?

timburks:
@seaneganx Good question - no, but that's built to a tag, so the effects of this wouldn't show up until the tag is updated. I'll take a look at the code before merging regardless.

Issue #863, opened 2022-12-06
Registry tool: improve testing of YAML import/export


timburks:
Currently all tests of YAML I/O are under the `registry apply` command, and tests rely on round-tripping (importing YAML, exporting it, and verifying that the export matches the imported YAML). Tests of core I/O logic should be moved to the `patch` package and modified so that imports compare against an internal structure built with the seeder and exports use that seeder. Command-specific tests should be limited to verifying that specified options are supported properly.

timburks:
Closing since the core issues are addressed in #872, #874, and #877 and the remaining concerns are covered by #875 and #879

timburks:
Reopening after noticing that `patch/apply.go` has zero test coverage.

timburks:
`patch/apply.go` test coverage will be fixed when tests are moved for #879

PR #877, opened 2022-12-14
YAML I/O tests for apis, versions, specs, and deployments

timburks:
These tests directly test the `patch` package and verify import and export of YAML against internally-built references. Contributes to #863. After this, we might only want to pare down the tests for `registry apply` before closing #863.


cmd/registry/patch/patch_test.go
```diff
@@ -25,7 +25,423 @@ func TestMain(m *testing.M) {
...
+	root := "projects/patch-api-test/locations/global"
+	tests := []struct {
+		resourceID string
+		parent     string
+		yamlFile   string
+		nested     bool
+		message    proto.Message
+	}{
+		{
+			resourceID: "registry",
```
theganyo, line 38: The tests are using the same resourceID, so the output of t.Run() won't be useful to identify which test failed. (Usually, we'd use a description of what is being tested.)

timburks: I noticed that "#1" is appended for the second test, so left it at that - but if you'd prefer I'll make it explicit.

theganyo: ah. ok. thanks.

theganyo (approved):

PR #882, opened 2022-12-15
Always use "openapi" as the spec id for bulk-uploaded openapi specs.

### LINKED ISSUE #865: Registry tool: bulk uploaders should avoid using file extensions in spec ids
Currently the bulk uploaders use `openapi.yaml`, `swagger.yaml`, `discovery.json`, and `protos.zip` as spec identifiers. To avoid confusion in YAML (and possibly other future) externalizations of registries, let's not use the file type extensions in resource identifiers.

Detail: YAML exports of resources are stored in files named with the resource id followed by a `.yaml` extension. So a resource with id `openapi.yaml` would be exported into `openapi.yaml.yaml` (and I think we want to avoid this).

timburks:
Closes #865.

This only updates the bulk uploader, related tests, and the demo script that depends on the bulk uploader.

Numerous tests use `openapi.yaml` as a spec ID. I think updating these would obscure the change in this PR and also that updating these tests is not particularly urgent since the ids are isolated within the tests. If that seems ok, I'll create a followup PR to update the tests.

### DIFF
--- a/cmd/registry/cmd/upload/bulk/openapi.go
+++ b/cmd/registry/cmd/upload/bulk/openapi.go
@@ -174,8 +174,7 @@ func (task *uploadOpenAPITask) populateFields() error {
 	versionPart := parts[len(parts)-2]
 	task.versionID = sanitize(versionPart)
 
-	specPart := parts[len(parts)-1]
-	task.specID = sanitize(specPart)
+	task.specID = "openapi"
 
 	var err error
 	task.contents, err = os.ReadFile(task.path)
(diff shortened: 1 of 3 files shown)


cmd/registry/cmd/upload/bulk/openapi.go
```diff
@@ -174,8 +174,7 @@ func (task *uploadOpenAPITask) populateFields() error {
 	versionPart := parts[len(parts)-2]
 	task.versionID = sanitize(versionPart)
 
-	specPart := parts[len(parts)-1]
-	task.specID = sanitize(specPart)
+	task.specID = "openapi"
```
seaneganx, line 177: Why make this a field on the struct at all anymore?

seaneganx (approved):

theganyo (approved):

cmd/registry/cmd/upload/bulk/openapi.go
timburks: good question - I guess I was subconsciously uncommitted to the permanence of this :) 

I'll remove it before merging

PR #886, opened 2022-12-19
Consolidate type-related helpers in new `types` package.

timburks:
This collects type-related helpers in the registry tool into a new `types` package (internal to the tool). Along with chipping away at `core`, it reduces duplicate name-to-proto mappings previously in `core/print.go` and the `patch` package down to one.


theganyo (approved):
LGTM. Would be nice to add test coverage in the future. :)

timburks:
@theganyo ack #887

seaneganx (approved):

PR #895, opened 2022-12-23
Set command output in yaml export tests to keep tests from being too noisy.

timburks:


cmd/registry/cmd/export/yaml_test.go
```diff
@@ -57,6 +58,8 @@ func TestExportYAML(t *testing.T) {
 		cmd := Command()
 		args := []string{"yaml", r}
 		cmd.SetArgs(args)
+		out := bytes.NewBuffer(make([]byte, 0))
```
seaneganx, line 61: Can this be replaced with io.Discard?

cmd/registry/cmd/export/yaml_test.go
```diff
@@ -57,6 +58,8 @@ func TestExportYAML(t *testing.T) {
...
+		cmd.SetOutput(out)
```
seaneganx, line 62: FYI SetOutput is deprecated. The docs suggest using SetOut and SetErr. If both really do need silencing we can keep using this, but might have to fix the bug if we try to upgrade after they remove the function in a new version of cobra. I think I prefer two lines of code over someone having to deal with that eventually.

https://pkg.go.dev/github.com/spf13/cobra#Command.SetOutput

timburks: Good to know - I was following [this](https://github.com/apigee/registry/blob/9bf1213bc7bf2ca5740bd12dd19a8fc50f7eeee0/cmd/registry/cmd/export/csv_test.go#L98), which also uses `SetOutput`. For clarity, I think it would be best to address these together with a PR when necessary.

cmd/registry/cmd/export/yaml_test.go
timburks: Good point, but I think that's going in the wrong direction -- we should probably be verifying the output in some way.

seaneganx (approved):
Can you help me understand how these sets of beliefs can be held simultaneously?
- We should verify command output. Using io.Discard in this PR would be the wrong direction, so we shouldn't do it.
- We should prefer non-deprecated functions. We might address that in the future, so it's best to use deprecated functions until then.

Aren't we going in the wrong direction by adding calls to deprecated functions in this PR? Isn't it best to address output verification in a future PR if necessary?

I'm rubber stamping this PR because I don't think it's worth spending any more time discussing. I still think you should make the changes I suggested, but feel free to disregard them.

Issue #884, opened 2022-12-17
Registry tool: `registry export` should export subtrees of registry contents, `registry get` should get individual yaml files


timburks:
`registry export csv` is blocking us from moving `registry export yaml` up to `registry export` -- let's consider moving it to `registry-experimental`, maybe as `registry-experimental export-csv`.

Issue #741, opened 2022-09-15
Registry tool: improve alignment of `registry get` and `registry list` with `kubectl`


timburks:
`kubectl get` allows users to list and view various resources in Kubernetes. Used with just a resource name, it acts as a `list` command, returning resources in a table that include their name and a few other relevant type-specific values. When the `-o yaml` option is specified, results are returned as YAML objects. When the `get` matches a single resource, the YAML for that resource is returned and when the `get` matches multiple resources, a YAML list of resources is returned. For example:
```
$ kubectl get deployments
NAME              READY   UP-TO-DATE   AVAILABLE   AGE
tiny              1/1     1            1           183d
shakes            1/1     1            1           150d
playtime          1/1     1            1           57d

$ kubectl get deployments -o yaml
apiVersion: v1
items:
- apiVersion: apps/v1
  kind: Deployment
  metadata:
...
```
In the above, matching deployments are returned as entries in the `items` array.

This seems like an appropriate pattern for the `registry` tool. `registry get` would accept patterns and by default would return lists of matching resources, and with the `-o yaml` option, it would return YAML representations of matching resources. Whether or not to return arrays of resources seems open (to me). If we do that, then I think `registry apply` should accept these arrays.

`kubectl` accepts many other output formats in addition to `yaml`. From the command line help:
```
  -o, --output='':
	Output format. One of: (json, yaml, name, go-template, go-template-file, template, templatefile, jsonpath,
	jsonpath-as-json, jsonpath-file, custom-columns, custom-columns-file, wide). See custom columns
	[https://kubernetes.io/docs/reference/kubectl/#custom-columns], golang template
	[http://golang.org/pkg/text/template/#pkg-overview] and jsonpath template
	[https://kubernetes.io/docs/reference/kubectl/jsonpath/].
```
Of these, we might want to support `json`, but I wouldn't prioritize any others without user input.

timburks:
#884

PR #896, opened 2022-12-23
Improve `registry get` to include listing capabilities and yaml output.

timburks:
The `registry get` interface is simplified to the following:
```
Usage:
  registry get PATTERN [flags]

Flags:
      --filter string   Filter selected resources
  -o, --output string   Output type (name, yaml, contents) (default "name")

Global Flags:
      --registry.address string    the server and port of the registry api (eg. localhost:8080)
      --registry.location string   the registry location
      --registry.project string    the registry project
      --registry.token string      the token to use for authorization to registry
```

When `PATTERN` is a collection or a resource name with wildcards, `registry get` finds all matching resources (using a `--filter` if provided). Names of matching resources are returned by default, and the `--output` (`-o`) flag can be used to select other output formats. Initially `name` (the default), `yaml`, and `contents` are provided. `contents` is only supported for specs and artifacts and returns the contents of these resources. When `contents` is specified, `PATTERN` must match a single resource. For other response types, when multiple resources match, an appropriate list is returned. `yaml` lists follow the `kubectl` pattern and look like this:
```
% registry get projects/googleapis/locations/global/apis -o yaml 
apiVersion: apigeeregistry/v1
items:
  - apiVersion: apigeeregistry/v1
    kind: API
    metadata:
      name: abusiveexperiencereport
    data:
      displayName: Abusive Experience Report API
      description: Views Abusive Experience Report data, and gets a list of sites that have a significant number of abusive experiences.
  - apiVersion: apigeeregistry/v1
    kind: API
    metadata:
      name: acceleratedmobilepageurl
    data:
      displayName: Accelerated Mobile Pages (AMP) URL API
      description: 'Retrieves the list of AMP URLs (and equivalent AMP Cache URLs) for a given list of public URL(s). '
  - apiVersion: apigeeregistry/v1
    kind: API
    metadata:
      name: accessapproval
    data:
      displayName: Access Approval API
      description: An API for controlling access to data by Google personnel.
...
```
 
This will eliminate the need for `registry list` and following this change, `registry export yaml` will be changed to only write files -- all of the cases that write YAML to stdout will be supported by `registry get`.

See #884 and #741 for justification.


server/registry/names/deployment_revision.go
```diff
@@ -22,7 +22,7 @@ import (
 var deploymentRevisionRegexp = regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/deployments/%s(?:@%s)?$",
 	identifier, Location, identifier, identifier, revisionTag))
 
-var deploymentRevisionCollectionRegexp = regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/deployments/%s@",
+var deploymentRevisionCollectionRegexp = regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/deployments/%s@$",
```
timburks, line 25: @theganyo a bug in the regex?

server/registry/names/spec_revision.go
```diff
@@ -22,8 +22,8 @@ import (
 var specRevisionRegexp = regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/versions/%s/specs/%s(?:@%s)?$",
 	identifier, Location, identifier, identifier, identifier, revisionTag))
 
-var specRevisionCollectionRegexp = regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/versions/%s/specs@$",
-	identifier, Location, identifier, identifier))
+var specRevisionCollectionRegexp = regexp.MustCompile(fmt.Sprintf("^projects/%s/locations/%s/apis/%s/versions/%s/specs/%s@$",
```
timburks, line 25: @theganyo maybe another bug in the regex? I found these writing tests and checking code coverage -- the associated branches weren't being taken until I updated these regexes.

cmd/registry/cmd/get/get.go
```diff
@@ -15,90 +15,329 @@
...
+type getHandler struct {
+	ctx         context.Context
+	client      connection.RegistryClient
+	adminClient connection.AdminClient
+	writer      io.Writer
+	name        string
+	filter      string
+	output      string
+	results     []interface{} // result values to be returned in a single message
+}
```
seaneganx, line 88: Which of these actually need to be fields on a struct, as opposed to parameters of a function? In a 60 second time box I was able to figure out that `traverse()` and `write()` are the two methods meant for direct use, and the others are helpers. It also seems like everything except `writer` and `results` are just used as inputs. Why isn't this just two functions?

The context for example, definitely shouldn't be a field on a struct. Each function that needs a context should accept it as an argument.

https://pkg.go.dev/context#pkg-overview
https://google.github.io/styleguide/go/decisions#contexts

server/registry/names/spec_revision.go
seaneganx: Should these be addressed in a separate PR?

cmd/registry/types/types.go
```diff
@@ -71,6 +71,19 @@ func IsZipArchive(mimeType string) bool {
...
+func IsPrintableType(mimeType string) bool {
+	if mimeType == "text/plain" {
+		return true
+	} else if strings.HasPrefix(mimeType, "application/yaml") {
+		return true
+	} else if strings.HasPrefix(mimeType, "application/json") {
+		return true
+	} else {
+		return false
+	}
```
seaneganx, line 84: `return mimeType == "text/plain" || strings.HasPrefix(mimeType, "application/yaml") || strings.HasPrefix(mimeType, "application/json")`

cmd/registry/cmd/get/get.go
timburks: This is struct collects state that is captured in closures passed to the `core.List*` and `core.Get*` functions.

server/registry/names/spec_revision.go
timburks: Maybe, certainly if we had more complete tests on the `names` package #911

cmd/registry/types/types.go
timburks: I find this easier to read and more extensible than one long line, but ok.

timburks: Updated, but since the single-line expression was over 130 columns wide, I included line breaks.

server/registry/names/deployment_revision.go
timburks: merged in #912

server/registry/names/spec_revision.go
timburks: merged in #912

cmd/registry/cmd/get/get.go
timburks: #913

pkg/models/project.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2022 Google LLC. All Rights Reserved.
```
theganyo, line 1: 2023? ;)

timburks: :) fixed

PR #900, opened 2023-01-04
Increase gRPC client timeout to 60 seconds

timburks:
This works around some issues that we're seeing in hosted implementations which cause API calls to be cancelled prematurely. Proxies are subtracting time from the deadlines that we send and cancelling if the remaining time is less than some threshold. Specifically, we see our 20s timeout become 10s, and then with a fraction of a second removed during processing, the operation is canceled because the hosting platform requires 10s minimums. This seems like a bug in the hosting configuration, so we might revisit this when that's better understood.

See also: https://github.com/googleapis/gapic-generator-go/issues/1206

### DIFF
--- a/gapic/admin_client.go
+++ b/gapic/admin_client.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2023 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


theganyo (approved):
If we haven't already, we should create the issue to follow up on this.

timburks:
@theganyo #901 awaits!

PR #910, opened 2023-01-06
Improvements to `registry delete`

### LINKED ISSUE #897: Registry tool: `registry delete` should support all resource types, wildcards, filters, and `--force` for cascading deletions
The `registry delete` subcommand was created to get past the awkwardness of direct calls to the `delete` RPCs. With a single subcommand, any resource should be deletable, and if wildcards and filters are specified, all matching resources should be deleted. When provided, the `registry.project` configuration should be used to simplify resource names. A `--force` option should be supported to allow cascading deletes.

Essentially, `registry delete` should be the same as `registry get` (following #896), but instead of listing or returning resources, it should delete them.

Although a basic command structure exists, there are no command tests, not all resource types are supported, and much of the above remains to be done.

timburks:
Fixes #897 - `registry delete` now takes the same basic arguments as `registry get` (replacing `-o` for output format with `--force` for cascading deletes and `--jobs` for parallel deletion).

Although the command supports them for consistency with `get`, deletions of revision collections are not tested because they are known to fail on the server since a single remaining revision cannot be deleted, and when these errors are returned by the server, they are caught  in concurrent tasks which do not return their errors back to the command handler.


cmd/registry/cmd/delete/delete.go
```diff
@@ -30,145 +31,332 @@ import (
...
+		return core.ListVersions(ctx, client, version, filter, h.apiVersionHandler())
+	} else if spec, err := names.ParseSpecCollection(name); err == nil {
+		return core.ListSpecs(ctx, client, spec, filter, false, h.apiSpecHandler())
+	} else if rev, err := names.ParseSpecRevisionCollection(name); err == nil {
+		return core.ListSpecRevisions(ctx, client, rev, filter, false, h.apiSpecRevisionHandler())
+	} else if artifact, err := names.ParseArtifactCollection(name); err == nil {
+		return core.ListArtifacts(ctx, client, artifact, filter, false, h.artifactHandler())
+	}
+
+	// Then try to match resource names containing wildcards, these also are treated as collections.
```
theganyo, line 122: Hmm. Should this be part of the collections' regexp?

timburks:
@theganyo agreed - #917

cmd/registry/cmd/delete/delete.go
```diff
@@ -30,145 +31,332 @@ import (
...
-	api names.Api,
-	filterFlag string,
-	taskQueue chan<- core.Task) error {
-	return core.ListAPIs(ctx, client, api, filterFlag, func(api *rpc.Api) error {
-		taskQueue <- &deleteTask{
-			client:       client,
-			resourceName: api.Name,
-			resourceKind: "api",
-		}
+func (h *deleteHandler) enqueue(task core.Task) {
```
theganyo, line 171: This kind of handler structure that can handle a list of any type is also a place that seems ripe for a refactored DRY solution. (I did it one way in lint.checker, you've done it slightly different.)

theganyo (approved):
No complaints, but some discussion notes.

cmd/registry/cmd/delete/delete.go
```diff
@@ -30,145 +31,332 @@ import (
...
+	cmd.Flags().BoolVarP(&force, "force", "f", false, "Force deletion of child resources")
 	return cmd
 }
 
-type deleteTask struct {
-	client       connection.RegistryClient
-	resourceName string
-	resourceKind string
+type deleteHandler struct {
+	ctx         context.Context
```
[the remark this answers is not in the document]
timburks: Interesting... maybe it's an advanced setting. I'm not opposed to addressing this in #913.

cmd/registry/cmd/delete/delete.go
```diff
@@ -30,145 +31,332 @@ import (
...
-		return task.client.DeleteApi(ctx, &rpc.DeleteApiRequest{Name: task.resourceName})
-	case "version":
-		return task.client.DeleteApiVersion(ctx, &rpc.DeleteApiVersionRequest{Name: task.resourceName})
-	case "spec":
-		return task.client.DeleteApiSpec(ctx, &rpc.DeleteApiSpecRequest{Name: task.resourceName})
-	case "artifact":
-		return task.client.DeleteArtifact(ctx, &rpc.DeleteArtifactRequest{Name: task.resourceName})
-	default:
-		return nil
+	// First try to match collection names.
```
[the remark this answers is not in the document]
timburks: +1

cmd/registry/cmd/delete/delete.go
timburks: I think we might benefit from keeping these cases separate -- currently when the collections regexp matches, we know that the last segment is the collection name. That wouldn't be true if we merged these cases, and I'm not sure the more powerful regexp would be worth its complexity.

PR #912, opened 2023-01-06
Fix patterns for revision collection names.

timburks:
Patterns used to match spec and deployment revision collections were breaking changes that I was testing for #896. Here's the fix in a standalone PR.


seaneganx (approved):

timburks:
@seaneganx thanks for the review - I'll proceed but @theganyo PTAL

PR #921, opened 2023-01-10
Move "registry export yaml" to "registry export"

timburks:
Finishes #884.

With this set of changes, `registry export` now exports the entire contents of a registry (given a project argument) or all of the information for a single API (given an API argument). Results are always written to a local directory. YAML to stdout is now available only through the `registry get` command.

`registry export projects/myproject` and `registry export projects/myproject/locations/global/apis/myapi` both write files into identical directory hierarchies. Argument types besides projects and APIs might be desirable (and a `--filter` flag), but these are omitted to keep this change from getting too complex.


cmd/registry/cmd/export/export_test.go
```diff
@@ -1,24 +1,15 @@
-// Copyright 2022 Google LLC. All Rights Reserved.
```
theganyo: Did you mean to delete the copyright?

cmd/registry/patch/patch_project_test.go
```diff
@@ -240,3 +240,100 @@ func TestProjectExport(t *testing.T) {
...
+				}
+				if diff := cmp.Diff(newBytes, refBytes); diff != "" {
+					t.Errorf("mismatched export %s %+v", newFilename, diff)
+				}
+				return nil
+			}); err != nil {
+				t.Fatalf("Setup: Failed to export api: %s", err)
+			}
+
+			if err := adminClient.DeleteProject(ctx, &rpc.DeleteProjectRequest{
```
theganyo, line 344: This should be put into a `t.Cleanup()` or it may not run.

cmd/registry/cmd/export/export_test.go
timburks: Nope, I caught that just as you did -- it should be back now.

cmd/registry/patch/patch_project_test.go
timburks: thanks, fixed

theganyo (approved):

PR #930, opened 2023-01-11
Fix some previously-undetected style concerns in `registry check` flags.

timburks:
I noticed that our new registry tool style checker was missing some flags because it was iterating over `Flags()` and not `LocalFlags()` *shrug* and when I changed that, I found that the missed flags were declared with `PersistentFlag()` instead of `Flag()` - @theganyo, was that needed? I changed it to `Flag()` for consistency but could change it back if you prefer. I also changed the checker to hopefully catch more problems and saw that problems in the check flags were still seen when I iterate over `LocalFlags()` (both `PersistentFlag()` and `Flag()`-based).


theganyo (approved):
SGTM

timburks:
@theganyo [This](https://github.com/spf13/cobra/blob/4fa4fdf5cd34cb78973561d077461caea8eeb48a/command.go#L1619) is what I'm trusting to prefer `LocalFlags()` over `Flags()`, which is defined above and lacks this call.

theganyo:
Yet Flags() says:
// Flags returns the complete FlagSet that applies
// to this command (local and persistent declared here and by all parents).

theganyo:
Hmm. Looking around that code, I believe that Flags() _should_ actually also make the mergePersistentFlags() call. Also, LocalFlags() has a filter against persistent flags... so I'm unsure what you're actually seeing here? Hmm.

timburks:
@theganyo I'm just reading the Cobra code - maybe it's incorrect?

theganyo:
\It seems like it to me, but I didn't spend too much time reviewing.

theganyo:
There is a related bug that's never been addressed. https://github.com/spf13/cobra/issues/412

timburks:
A related nit, I've noticed that when flags are declared like [this](https://github.com/apigee/registry/blob/c075813560f9896b32f1fdceade58ad21de020d1/cmd/registry/cmd/get/get.go#L74), then there's no need to call `cmd.Flags().GetString(...)` which can fail and have [hard-to-test error handling](https://github.com/apigee/registry/blob/c075813560f9896b32f1fdceade58ad21de020d1/cmd/registry/cmd/check/check.go#L63).

theganyo:
Yeah, I usually do that. In this case, I think I just grabbed another CLI command we had and followed it.

Issue #875, opened 2022-12-14
Registry tool: YAML I/O tests needed for scoring and style-related messages.


timburks:
Upon completion of #874, the following artifact types will still need testing:

- google.cloud.apigeeregistry.v1.scoring.Score
- google.cloud.apigeeregistry.v1.scoring.ScoreDefinition
- google.cloud.apigeeregistry.v1.scoring.ScoreCard
- google.cloud.apigeeregistry.v1.scoring.ScoreCardDefinition
- google.cloud.apigeeregistry.v1.style.ConformanceReport
- google.cloud.apigeeregistry.v1.style.Lint

Tests should be added in `registry/cmd/patch/patch_tests.go` (see `TestArtifactPatches`) and should match the style of other artifact tests there. Sample artifacts do not need to be large or true-to-life but should exercise all fields.

PR #931, opened 2023-01-11
Add yaml tests for score/conformance protos

shrutiparabgoogle:
Fixes https://github.com/apigee/registry/issues/875

The Lint proto is not directly used to set any artifact currently, it is indirectly used in the ConformanceReports. Eventually we should keep only the required messages and merge the Conformance and Lint proto files into one.


timburks (approved):

timburks:
I was surprised that this didn't increase our coverage percentage, but I've noticed in other PRs that we get credit for coverage (by exercising code) without necessarily verifying that the results produced are correct, so this was still worthwhile. Thanks!

shrutiparabgoogle:
These tests are not essentially covering new lines of code, assuming the logic to convert a yaml to a proto is shared between different types. But these tests are still a good safeguard to verify that all our defined protos work as expected.

Issue #933, opened 2023-01-12
Registry tool: add registry rules for Api


theganyo:
Add Api resource checks to `registry check` command.

rule group | rule number | resource | field | type | severity | check
-- | -- | -- | -- | -- | -- | --
Registry | 100 | Api | availability | string | INFO | single word
Registry | 101 | Api | recommended_version | string | ERROR | valid ApiVersion name under this Api
Registry | 102 | Api | recommended_deployment | string | ERROR | valid ApiDeployment name under this Api

PR #935, opened 2023-01-12
add registry rules for Api

theganyo:
WIP toward #933

Note: Check framework needs ability to do name lookup (#936) before the issue can be completed. However, this PR works standalone, so if we can get it in, it would be helpful from a dependency standpoint.


cmd/registry/cmd/check/lint/problem.go
```diff
@@ -26,6 +26,18 @@ const (
 	ERROR
 )
 
+func (s Severity) String() string {
+	switch s {
+	case INFO:
+		return "INFO"
+	case WARNING:
+		return "INFO"
```
timburks, line 34: return "WARNING" ?

cmd/registry/cmd/check/rules/rule110/rule110.go
```diff
@@ -1,18 +1,18 @@
-// Copyright 2022 Google LLC
+// Copyright 2023 Google LLC. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     https://www.apache.org/licenses/LICENSE-2.0
+//    http://www.apache.org/licenses/LICENSE-2.0
```
timburks, line 7: Did you mean to change https to http?

timburks (approved):

cmd/registry/cmd/check/lint/problem.go
theganyo: fixed

cmd/registry/cmd/check/rules/rule110/rule110.go
theganyo: yes, this was inconsistent with all other notices

PR #940, opened 2023-01-13
Add tests for `names` package.

### LINKED ISSUE #911: Registry server: 100% test coverage of the `names` package?
Currently the [test coverage of the names package](https://app.codecov.io/gh/apigee/registry/tree/main/server/registry) is just under 40%. Because this package is so widely used (and useful), we think that its transitive coverage could be much higher, but transitive coverage is not currently being measured, so we have no way to be sure what's actually not covered by tests.

Recently we encountered a couple of possible errors in less-used expressions ([example](https://github.com/apigee/registry/pull/896#discussion_r1063837984), fixed in #912) that we didn't see until we found problems in dependent code. It's probably fair to think that we would have found these sooner if we had directly tested the expressions in the `names` package.

Why aim for 100%? That's arbitrary, but the `names` code seems relatively easy to test (compared to other packages), and  100% test coverage would give us confidence to proceed with #851 and make other improvements that we've discussed in issues.

timburks:
I think this gets close enough to close #911.

Uncovered code appears to be unreachable.

Tests could be better (not all failures are clearly explained), but since the `names` package is so simple, I think it's better to not overinvest in this and leave further test improvement for when/if it's needed.

### DIFF
--- a/server/registry/names/common.go
+++ b/server/registry/names/common.go
@@ -36,11 +36,6 @@ const (
 // User provided identifiers should be validated according to this format.
 var customIdentifier = regexp.MustCompile(`^[a-z0-9-.]+$`)
 
-// GenerateID generates a random resource ID.
-func GenerateID() string {
-	return uuid.New().String()[:8]
-}
-
 // validateID returns an error if the provided ID is invalid.
 func validateID(id string) error {
 	if "" == id {
--- a/server/registry/names/api_test.go
+++ b/server/registry/names/api_test.go
@@ -1,0 +1,64 @@
+// Copyright 2023 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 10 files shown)


server/registry/names/common.go
```diff
@@ -36,11 +36,6 @@ const (
 // User provided identifiers should be validated according to this format.
 var customIdentifier = regexp.MustCompile(`^[a-z0-9-.]+$`)
 
-// GenerateID generates a random resource ID.
-func GenerateID() string {
```
timburks, line 40: Nothing was using this (I think it was orphaned when we stopped automatically generating IDs), so I lazily decided to delete it rather than test it.

theganyo:
Nice! Nearly 100%!  It says the only uncovered code is this line? 

https://github.com/apigee/registry/blob/main/server/registry/names/project.go#L36

theganyo (approved):

timburks:
@theganyo that's right, and I don't believe it's reachable. Any `String()` with a valid `ProjectID` would match the regex. It seemed too controversial for this PR but I wouldn't mind taking it out in a followup.

PR #943, opened 2023-01-14
Move files only needed for `registry compute` commands out of `core`

timburks:
This is probably not the last stop for these files -- `cmd/compute` looks like it should be divided by subcommand -- but this helps cut `core` down to a smaller set of essentials.


cmd/registry/cmd/compute/complexity-discovery.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
```
theganyo, line 1: 2023

theganyo (approved):

PR #947, opened 2023-01-17
add context to rules with client access

### LINKED ISSUE #936: Registry tool: check framework needs ability to do name lookup
When evaluating internal reference fields, rules need to be able to lookup resources by name to verify their existence.

theganyo:
Closes #936


cmd/registry/cmd/check/lint/checker.go
```diff
@@ -28,6 +28,12 @@ import (
 	"github.com/apigee/registry/server/registry/names"
 )
 
+type contextKey int
+
+const (
+	ContextKeyClient contextKey = iota
```
timburks, line 34: We might want to say `ContextKeyRegistryClient` here -- I know it's more verbose :( but it would align better with `ContextKeyAdminClient` if we decide we need that and it would make the purpose of this more obvious -- it wasn't until I saw the usage below that I realized what "Client" was referring to here :)

timburks (approved):

cmd/registry/cmd/check/lint/checker.go
theganyo: ack

Issue #946, opened 2023-01-17
Registry tool: store structured artifacts as YAML instead of serialized protobuf


timburks:
Informal feedback has suggested that our practice of storing artifacts as serialized protos puts too much of a burden on registry users, requiring them to write code to convert YAML or JSON equivalents into serialized protos before posting them in [artifact contents fields](https://github.com/apigee/registry/blob/665921c74025ace8aadfe6035c3838869c8aa71a/google/cloud/apigeeregistry/v1/registry_models.proto#L373).

It seems overly-restrictive to require all artifacts to be text, but most of the artifacts that we use could be stored as text (e.g. YAML), and this would allow users easily to view and modify artifact contents with HTTP methods.

Note that this isn't a problem for users of the `registry` tool, which automatically serializes the appropriate types.

Moving forward, we might take the following steps:
1. Begin by modifying `registry apply` to take an optional argument that causes it to store artifacts as YAML instead of serializing them as proto. Mime types for artifacts might become `application/yaml;type=KIND`, where `KIND` is the value of the `Kind` field in YAML representations.
2. Update `registry` subcommands that read artifacts to support both serialized protos and YAML.
3. Update applications that use artifacts to support both serialized protos and YAML.
4. Make YAML the default representation.
5. Consider deprecating serialized proto representations.

theganyo:
I'm wondering if we should keep fully qualified types for the yaml mimes as we do with the proto types. 

Ie. Instead of:  `application/yaml;type=Score` we'd use: `application/yaml;type=google.cloud.apigeeregistry.v1.scoring.Score`

This would keep name-spacing to avoid any potential conflicts and make it trivial to map between the "octet-stream" and "yaml" versions.  

It would also enables us to easily distinguish our "well-known" internally-used kinds of archives from other ad-hoc types. Of course, this also opens the question of whether we should be doing the same in our archive yaml. Right now, it seems we're loosely "reserving" these types.

theganyo:
The first 2 tasks are now complete so this is blocked pending API Hub updates to be able to read yaml artifacts.

Following that, when we continue with task 4 - along with making `apply` use yaml as default, we'll also want to look at updating the places that we're directly using `proto.Marshal()` in the `cmd/` hierarchy to store compute, conformance, and scoring artifacts.

PR #949, opened 2023-01-17
system artifacts: add `system-buildinfo` to return buildinfo

timburks:
This adds support for an experimental (and controversial?) special type of artifact that we might call "system artifacts".

System artifacts are defined at the project level and their ids begin with `system-`. They cannot be created, listed, deleted, or replaced, but they can be read with `get-artifact` and `get-artifact-contents`.

One system artifact is defined, `system-buildinfo`, which returns a YAML-formatted description of the build and its dependencies. It is for use in environments that don't have access to [Admin.GetStatus](https://github.com/apigee/registry/blob/bd0123ccb6189b2fa11b3d783d7ef90337803d52/google/cloud/apigeeregistry/v1/admin_service.proto#L45).

We could access the artifact with the following:
```
$ registry get artifacts/system-buildinfo -o yaml
apiVersion: apigeeregistry/v1
kind: BuildInfo
metadata:
  name: system-buildinfo
data:
  goVersion: go1.18.1
  path: github.com/apigee/registry/cmd/registry-server
  main:
    path: github.com/apigee/registry
    version: (devel)
    sum: ""
    replacement: null
  dependencies:
    - path: cloud.google.com/go
      version: v0.104.0
      sum: h1:gSmWO7DY1vOm0MVU6DNXM11BWHHsTUmsC5cv1fuW5X8=
      replacement: null
  ...
  settings:
    -compiler: gc
    CGO_CFLAGS: ""
    CGO_CPPFLAGS: ""
    CGO_CXXFLAGS: ""
    CGO_ENABLED: "1"
    CGO_LDFLAGS: ""
    GOAMD64: v1
    GOARCH: amd64
    GOOS: linux
    vcs: git
    vcs_modified: "true"
    vcs_revision: cfa42e6a3d6ddbbc1724543556c68d17e01b6933
    vcs_time: "2023-01-17T23:56:10Z"
```

And with tools like `jq` and `yq`, we could pick out specific values as needed:
```
$ registry rpc get-artifact-contents --name "projects/foo/locations/global/artifacts/system-buildinfo" -j | jq .data -r | base64 -d | yq .settings.vcs_revision
https://github.com/timburks/registry/commit/ea5d4c317c3aa93ef48d117820db8904a0ff6f8a

$ registry get artifacts/system-buildinfo -o yaml | yq .data.settings.vcs_revision
https://github.com/timburks/registry/commit/ea5d4c317c3aa93ef48d117820db8904a0ff6f8a
```

Some possible questions:
- Why `system-` and not some other prefix? It's arbitrary, but I think we want to use valid artifact ids, which means no special prefixes like underscores or exclamation points.
- Why not return these from `List` operations? At first it's laziness - including them in list operations would complicate listings - but also these might confuse users and as "system" features, they might be better hidden anyway.
- What else might we want to have in system artifacts? Maybe some database-related info or registry configuration, but I don't think these are for application-specific data, meaning I wouldn't change any existing prefixes if we added these.
- Would `system-` artifacts always be read-only? Maybe...
- Does this break any proxying layers that expect to cache artifact metadata? Worth investigating... @Krpate?
- Why return YAML and not serialized proto? User-friendliness
- Why not JSON? Consistency with our current usage of YAML in the `registry` tool and the possible changes in #946.

(If we like this, I'll add tests)


server/registry/actions_artifacts.go
```diff
@@ -60,6 +61,9 @@ func (s *RegistryServer) CreateArtifact(ctx context.Context, req *rpc.CreateArti
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	if system.Reserved(parent.Artifact(req.ArtifactId)) {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
theganyo, line 65: This needs to create or use a different error than the nil one left over from the parseArtifactParent() call above it.

server/registry/actions_artifacts.go
```diff
@@ -128,6 +132,9 @@ func (s *RegistryServer) DeleteArtifact(ctx context.Context, req *rpc.DeleteArti
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	if system.Reserved(name) {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
theganyo, line 136: This needs to create or use a different error than the nil one left over from the parseArtifact() call above it.

theganyo (changes requested):

server/registry/actions_artifacts.go
timburks: oops, yes, I would have caught that if I'd written a test :)

theganyo:
It would be fairly easy to argue that this should be an explicit API. Is this a truly completely separate class of thing than operational health (eg. kubernetes) or runtime metrics (eg. prometheus) endpoints? That said, I do like the dynamic aspect of being able to expose new classes of system metadata without having to add new APIs... so I'm ambivalent at this point and easily swayed.

timburks:
> Is this a truly completely separate class of thing than operational health (eg. kubernetes) or runtime metrics (eg. prometheus) endpoints?

That's a very broad question! Yes? No? Failing to answer that, I think this is a practical compromise that accounts for the high cost of adding new API methods and the relatively low value of this information (it is high to a few developers but low to most users). If someone wanted to push through an API method that provided build info, I'd be happy to use the new method but I wouldn't be able to confidently say they'd spent their time productively considering all the other things they might have done. But we do have situations where we would like to verify the build version that we're working with, and there's nothing better than providing that directly (IMO).

theganyo:
Yes and opportunity cost is likely the strongest point here: Since this solution extends and existing pattern in an easily understandable way and reasonably clean way with minimal effort, that's a huge benefit. And, since I can't think of any truly painful side-effects, I feel this is a reasonable compromise.

server/registry/internal/system/artifacts.go
```diff
@@ -0,0 +1,137 @@
...
+func contents(name names.Artifact) ([]byte, string, error) {
+	switch name.ArtifactID() {
+	case "system-buildinfo":
+		return system_buildinfo()
+	default:
+		return nil, "", status.Errorf(codes.NotFound, "%q not found", name)
+	}
+}
+
+func system_buildinfo() ([]byte, string, error) {
```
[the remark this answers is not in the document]
seaneganx: Also, camelCase function names please.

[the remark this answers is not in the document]
timburks: > ...camelCase...

Thanks, fixed.

> The return values seem unnecessarily generic. Can we return fully formed artifact messages instead? 

I prefer not to do that because I think it would create a pattern in which any function returning a system artifact would return a fully formed artifact and redundantly compute the content size and hash. Here there's a clear division of responsibility with content generation separate from artifact construction (and I am thinking that additional artifacts would be added, though I don't have any to suggest here).

> Are these system artifacts always going to be manifested (for lack of a better term) by the server at runtime? The implementation seems designed for that: allowing us to add more recognized IDs that generate a response.

Yes, exactly. This PR includes a "name grab" of the `system-` prefix to hold it in reserve for our use. I think we should clarify (I think I didn't make this clear enough) that system artifacts are experimental features that do not have the same guarantees of stability as the API surface. I think we can and should reserve the right to change system artifacts and their contents at any time (including removing them).

>> Would system- artifacts always be read-only? Maybe...

> This seems like it would be a pretty big change because it would introduce statefulness to system artifacts. I think that change would challenge a lot of the assumptions proposed in this PR, e.g. not including system artifacts in list operations. Maybe worth thinking a bit more about whether this will ever happen.

Indeed. I don't see that uncertainty blocking our usage of read-only system artifacts, but it is something that we should consider if we see a use for mutable system artifacts.

> Have we considered adding a collection specifically for system artifacts? Not one method per artifact as you two have been discussing, but also not special behavior for a subset of resources in the projects/*/locations/*/artifacts/* collection. Instead we would add an additional_binding on GetArtifact for a collection like projects/*/locations/*/systemArtifacts/*.

Yes, and I think this would be bad for the API design for the same reasons that we don't have application-specific collections of artifacts (there is no "apihubArtifacts" collection, for example). Codifying these subgroups would suggest that such new collections are expected or necessary when they aren't at all - simply by using nonoverlapping sets of ids, applications could share the artifact collection space and potentially benefit from sharing information among them -- though the biggest benefit is the reduction of friction for application developers. Focusing back on system artifacts, creating a "systemArtifacts" collection would also make it harder for us to change our mind about having system artifacts (because then we would also have to remove the binding from our protos) -- I think it's too soon to make that level of commitment.

server/registry/actions_artifacts.go
timburks: Fixed and updated with tests, thanks!

(I wrote the tests before fixing the code and the code crashed on the err.Error() call)

server/registry/actions_artifacts.go
timburks: fixed, thanks

Krpate:
>Does this break any proxying layers that expect to cache artifact metadata?

We don't cache artifact metadata today but to consider plans down the line for making artifacts CCFE managed - If we are saying that system artifacts are always created by server during runtime and are **read only** then it shouldn't break anything. 

> but also these might confuse users and as "system" features, they might be better hidden anyway.

I realize how `system-buildinfo` can be useful for us as Apigee Registry API producers to provide support and debug issues. I am not seeing how this would be useful to API hub customers? If it's only for our use then maybe the internal producer API can be used for exposing this info?

timburks:
> I realize how system-buildinfo can be useful for us as Apigee Registry API producers to provide support and debug issues. I am not seeing how this would be useful to API hub customers? If it's only for our use then maybe the internal producer API can be used for exposing this info?

@Krpate thanks for this suggestion - I think there's value in keeping this solution open so that application-builders can also independently confirm the version of the API that they are using (or access whatever other metadata that we decide to share here). That certainly comes with caveats: this is not an official API feature and can be changed or removed at any time.

seaneganx (changes requested):
>>Have we considered adding a collection specifically for system artifacts? Not one method per artifact as you two have been discussing, but also not special behavior for a subset of resources in the projects/locations/artifacts/* collection. Instead we would add an additional_binding on GetArtifact for a collection like projects/locations/systemArtifacts/*.

>Yes, and I think this would be bad for the API design

We should discuss the public-facing API changes separately in an API design review before proceeding. Not only because we can't agree on an approach that isn't bad, but primarily because API design review is mandatory for changes to public APIs like ours.

I described this change to the Cloud API design team. They don't consider it an acceptable solution and requested you escalate to them with a business reason for an exception before proceeding with the change.

theganyo:
One possibility that we haven't discussed here involves no change to the API at all and follows an existing pattern: That is, simply use the controller to create / populate an regular artifact.

PR #968, opened 2023-01-25
Bugfix for getting current user email

stokedout:
Closed original #967  due to not having signed the CLA.

The existing way had two different env variables and `gcloud config get account` doesn't just print the email it also includes the project:

```
$ gcloud config get account                                                                                                                                                           ─╯
$ Your active configuration is: [api-registry]
$ john@smith.com
```

```
$ gcloud config list account --format "value(core.account)"
$ john@smith.com
```

I believe you want the latter?!

### DIFF
--- a/auth/HOSTED.sh
+++ b/auth/HOSTED.sh
@@ -36,7 +36,7 @@ APG_REGISTRY_INSECURE="false"
 
 APG_REGISTRY_PROJECT="$(gcloud config get project)"
 APG_REGISTRY_LOCATION="global"
-APG_REGISTRY_CLIENT_EMAIL="$(gcloud config list account --format "value(core.account)")"
+APG_REGISTRY_CLIENT_EMAIL="$(gcloud auth list --filter=status:ACTIVE --format="value(account)")"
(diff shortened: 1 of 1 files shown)


theganyo:
Ah, you have multiple configurations... I only have one, so I didn't see that. I did a little snooping, however, and it seems like the issue you see might also occur in some cases with the list / format command you suggested? See: https://stackoverflow.com/questions/35311686/how-to-get-the-active-authenticated-gcloud-account.

The suggestion there is to add an additional filter param to ensure that doesn't happen:

`gcloud auth list --filter=status:ACTIVE --format="value(account)"`

WDYT?

stokedout:
> Ah, you have multiple configurations... I only have one, so I didn't see that. I did a little sleuthing, however, and it seems like the issue you see might also occur in some cases with the list / format command you suggested? See: https://stackoverflow.com/questions/35311686/how-to-get-the-active-authenticated-gcloud-account.
> 
> The suggestion there is to add an additional filter param to ensure that doesn't happen:
> 
> `gcloud auth list --filter=status:ACTIVE --format="value(account)"`
> 
> WDYT?

That works too, thanks! tried locally before applying

theganyo (approved):
Thanks for your contribution!

PR #970, opened 2023-01-25
Allow hosted projects to be exported with `registry export`

timburks:
We've observed that `registry export projects/PROJECT` fails to export hosted PROJECT instances, because these lack the Admin service. This modifies the `core.GetProject()` helper to optionally provide dummy projects when the Admin service is unavailable. This is limited to `registry export`, which will fail subsequently if a nonexistent project is provided. `registry get` and `registry delete` will continue to fail for hosted project because these operations don't make sense for hosted projects (deleting an entire project is conceivable but is not something that we should support in this way).

Exporting an existing project now just works:
```
$ registry export projects/timburks-test
  INFO[0001] Exported projects/timburks-test/locations/global/apis/apigeeregistry uid=916c601f
  INFO[0001] Exported projects/timburks-test/locations/global/apis/artifactregistry uid=916c601f
  INFO[0001] Exported projects/timburks-test/locations/global/apis/helloworld uid=916c601f
```

Exporting a non-existent project fails soon enough, I think:
```
timburks@penguin:~/Desktop/exports$ registry export projects/timburks-nonexistent
Error: rpc error: code = PermissionDenied desc = Permission denied on resource project timburks-nonexistent.
error details: name = ErrorInfo reason = CONSUMER_INVALID domain = googleapis.com metadata = map[consumer:projects/timburks-nonexistent service:apigeeregistry.googleapis.com]
error details: name = DebugInfo detail = Project 'project:timburks-nonexistent' not found or deleted. stack = 
error details: name = Help desc = Google developer console API key url = https://console.developers.google.com/project/timburks-nonexistent/apiui/credential
Usage:
  registry export PATTERN [flags]

Flags:
      --filter string   filter selected resources
  -h, --help            help for export
  -j, --jobs int        number of actions to perform concurrently (default 10)
  -r, --recursive       include child resources in export
      --root string     root directory for export

Global Flags:

----
Need more help?
```

Getting and deleting hosted projects continues to fail:
```
$ registry get projects/timburks-nonexistent
Error: rpc error: code = Unimplemented desc = The GRPC target is not implemented on the server, host: apigeeregistry.googleapis.com, method: /google.cloud.apigeeregistry.v1.Admin/GetProject.
Usage:
  registry get PATTERN [flags]

Flags:
      --filter string   filter selected resources
  -o, --output string   output type (name|yaml|contents) (default "name")

Global Flags:

----
Need more help?
$ registry delete projects/timburks-nonexistent
Error: rpc error: code = Unimplemented desc = The GRPC target is not implemented on the server, host: apigeeregistry.googleapis.com, method: /google.cloud.apigeeregistry.v1.Admin/GetProject.
Usage:
  registry delete PATTERN [flags]

Flags:
      --filter string   filter selected resources
  -f, --force           force deletion of child resources
  -h, --help            help for delete
  -j, --jobs int        number of actions to perform concurrently (default 10)

Global Flags:

----
Need more help?
```

### DIFF
--- a/cmd/registry/cmd/delete/delete.go
+++ b/cmd/registry/cmd/delete/delete.go
@@ -148,7 +148,7 @@ func (h *deleteHandler) traverse() error {
 	}
 
 	if project, err := names.ParseProject(name); err == nil {
-		return core.GetProject(ctx, adminClient, project, h.projectHandler())
+		return core.GetProject(ctx, adminClient, project, false, h.projectHandler())
 	} else if api, err := names.ParseApi(name); err == nil {
 		return core.GetAPI(ctx, client, api, h.apiHandler())
 	} else if deployment, err := names.ParseDeployment(name); err == nil {
--- a/cmd/registry/cmd/export/export.go
+++ b/cmd/registry/cmd/export/export.go
@@ -149,7 +149,7 @@ func (h *exportHandler) traverse() error {
 	}
 
 	if project, err := names.ParseProject(pattern); err == nil {
-		return core.GetProject(ctx, adminClient, project, h.projectHandler())
+		return core.GetProject(ctx, adminClient, project, true, h.projectHandler())
 	} else if api, err := names.ParseApi(pattern); err == nil {
 		return core.GetAPI(ctx, client, api, h.apiHandler())
 	} else if deployment, err := names.ParseDeployment(pattern); err == nil {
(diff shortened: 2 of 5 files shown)


theganyo:
Exporting a non-existent project fails with `PermissionDenied`... which seems misleading (despite the DebugInfo eventually saying it's missing). Is that something that we could / should address to avoid confusion?

theganyo (approved):

timburks:
@theganyo yes, we'll have to address this for #971 and can bring something from that to `registry export`.

timburks:
(@theganyo - actually, the hosted project might not be missing, it might be someone else's. I think the platform might be returning PERMISSION_DENIED for everything to prevent hostile probing for project names)

Issue #827, opened 2022-11-11
Registry tool: commands that call `log...Fatal()` are untestable.


timburks:
Throughout the `registry` tool, command implementations handle fatal errors by calling functions like [this one](https://github.com/apigee/registry/blob/937e76e0e8617d3358b53e9022387d183727fb15/cmd/registry/cmd/upload/bulk/protos.go#L65):
```
log.FromContext(ctx).WithError(err).Fatal("Failed to identify parent project")
```

However, if we attempt to write tests to verify these error handlers, the tests always fail because the call to `Fatal` calls `os.Exit()` ([discussion](https://stackoverflow.com/questions/30688554/how-to-test-go-function-containing-log-fatal)).

One way to fix this would be add a mode to `Fatal()` to prevent it from exiting in the situations that we want to test, but this would probably also prevent the cobra wrapper code from reporting the error to the caller in test code like this:
```
cmd := Command()
cmd.SetArgs(test.args)
err := cmd.Execute()
```

Fortunately, cobra also provides an alternate way to define run handlers that return error codes. If we define commands by setting the [RunE](https://github.com/spf13/cobra/blob/b43be995ebb4bee335a787bd44498b91aef7619c/command.go#L130) fields, they are expected to return errors, and these errors are passed back to callers of `cmd.Execute()`.

We might begin to do this incrementally as we write tests for command error-handling, but I think generally we should consider deprecating calls to `Fatal` and moving to `RunE` for all `registry` tool command handlers.

timburks:
For cases where we expect errors, we can prevent cobra from printing error and usage descriptions (which clutter up test logs) by setting `SilenceUsage` and `SilenceErrors` before calling `cmd.Execute()`.
```
cmd := Command()
cmd.SilenceUsage = true
cmd.SilenceErrors = true
cmd.SetArgs(test.args)
err := cmd.Execute()
```

theganyo:
I agree with returning errors directly via RunE instead of using log.Fatal(). Doing so also ensures that it emits help and the exit value for the command is non-zero.

timburks:
As noted by @theganyo, let's be mindful of how we report errors when we replace calls to `log.Fatal` with `return err`.  Returned errors should be descriptive without adding unnecessary verbosity. If we need to wrap errors, we'll probably want to follow an [idiomatic strategy](https://go.dev/blog/go1.13-errors).

 https://github.com/apigee/registry/pull/859#discussion_r1042657457

timburks:
[...]

PR #972, opened 2023-01-26
Allow uploads of CSV to hosted registries.

### LINKED ISSUE #971: Registry tool: `registry upload csv` requires the Admin service
Using an Admin service call [here](https://github.com/apigee/registry/blob/989942a561d89fd1f51ec1bdca34b496899be252/cmd/registry/cmd/upload/csv.go#L63) to verify the existence of a project causes `upload csv` to fail whenever it is used on a hosted project.

We should instead use [this](https://github.com/apigee/registry/blob/989942a561d89fd1f51ec1bdca34b496899be252/cmd/registry/core/verify.go#L10) or something like it that doesn't require the Admin service.

timburks:
This addresses #971 by using [core.VerifyLocation()](https://github.com/apigee/registry/blob/989942a561d89fd1f51ec1bdca34b496899be252/cmd/registry/core/verify.go#L10) instead of the Admin service to verify that a parent project exists.

Other incidental changes:
- `upload csv` no longer creates projects on demand, as with other uploaders, destination projects must be preexisting.
- this addresses #827 by converting the `upload csv` command to use `RunE` and return errors instead of logging fatal errors.
- the `--parent` flag is used instead of `--project_id` and is processed by a common function used by other `upload` commands.
- Names are used instead of strings in the command implementation.

(new tests for error cases are coming to this PR)

### DIFF
--- a/cmd/registry/cmd/upload/csv.go
+++ b/cmd/registry/cmd/upload/csv.go
@@ -35,7 +35,6 @@ import (
 
 func csvCommand() *cobra.Command {
 	var (
-		projectID string
 		delimiter string
 		jobs      int
 	)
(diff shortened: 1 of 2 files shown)


cmd/registry/cmd/upload/csv.go
```diff
@@ -44,37 +43,33 @@ func csvCommand() *cobra.Command {
 		Use:   "csv file --project-id=value [--delimiter=value]",
```
theganyo, line 43: We shouldn't include the flags in the `Use:` as they are listed automatically. (Also, `--project-id` is deprecated.)

theganyo (approved):

theganyo:
Also: This has a `fixes` tag in the description... but #971 won't be completely fixed by this PR as it also requires a wiki entry.

cmd/registry/cmd/upload/csv.go
timburks: thanks, I'll take those out

seaneganx (approved):

timburks:
"Also: This has a fixes tag in the description... but https://github.com/apigee/registry/issues/971 won't be completely fixed by this PR as it also requires a wiki entry."

@theganyo I think I've removed that "fixes" tag, if not I'll reopen the issue until we have the wiki page.

Issue #913, opened 2023-01-06
Registry tool: modify Handler interfaces to accept context arguments


timburks:
The handler interfaces defined [here](https://github.com/apigee/registry/blob/17331c741852e6ae285f1b8704a34af18b177753/cmd/registry/core/handlers.go#L21) currently accept a single message value and return an error. To comply with Go style guidance mentioned in [this comment](https://github.com/apigee/registry/pull/896#discussion_r1063860442), we should expand each of these interfaces to take a context as their first argument so that no other mechanism is used to pass a context into these functions.

PR #979, opened 2023-01-26
Factor a "visitors" pattern out of the get/delete/export implementations.

timburks:
This defines a new `Visitor` interface and a `Visit` function that applies the visitor to resources matching a pattern and filter. This was factored out of the `registry get`, `registry export`, and `registry delete` commands. It builds on the `List*` and `Get*` functions in `core` without making any changes to those. The `Visitor` interface returns closures that are passed to these functions. If usage expands, I think we might consider new `List` and `Get` methods that just take `Visitor` instances and call appropriate functions in the interface - then the `Visitor` interface would just be handlers and not need to return closures. But this seems like a good incremental step forward.

Note that there are some contexts in structs pending #913.

Looking ahead to testing, I think this and related functions should move to a new package under `pkg` and be tested there. Thoughts? `pkg/visitor`?

### DIFF
--- a/cmd/registry/core/visitors.go
+++ b/cmd/registry/core/visitors.go
@@ -35,11 +35,11 @@ type Visitor interface {
 }
 
 type VisitorOptions struct {
-	RegistryClient   connection.RegistryClient
-	AdminClient      connection.AdminClient
-	Pattern          string
-	Filter           string
-	AllowUnavailable bool
+	RegistryClient          connection.RegistryClient
+	AdminClient             connection.AdminClient
+	Pattern                 string
+	Filter                  string
+	AllowUnavailableProject bool
 }
 
 // Visit traverses a registry, applying the Visitor to each selected resource.
@@ -95,7 +95,7 @@ func Visit(ctx context.Context, v Visitor, options VisitorOptions) error {
 	}
 	// Finally, match individual resources
 	if project, err := names.ParseProject(name); err == nil {
-		return GetProject(ctx, ac, project, options.AllowUnavailable, v.ProjectHandler())
+		return GetProject(ctx, ac, project, options.AllowUnavailableProject, v.ProjectHandler())
 	} else if api, err := names.ParseApi(name); err == nil {
 		return GetAPI(ctx, rc, api, v.ApiHandler())
 	} else if deployment, err := names.ParseDeployment(name); err == nil {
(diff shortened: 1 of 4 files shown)


cmd/registry/core/visitors.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Visit traverses a registry, applying the Visitor to each selected resource.
+func Visit(ctx context.Context, v Visitor, options VisitorOptions) error {
+	filter := options.Filter
+	name := options.Pattern
+	ac := options.AdminClient
+	rc := options.RegistryClient
+
+	// First try to match collection names.
```
theganyo, line 52: One thing I adopted in my [tree class](https://github.com/theganyo/registry/blob/main/cmd/registry/cmd/check/tree/tree.go#L41) is to dispatch in a switch. It can look like this, which I find much easier to read:

```
name := names.Parse()
switch name := root.(type) {
  case names.Project:
    return  ListProjects(ctx, ac, name, filter, v.ProjectHandler())
  case names.API:
    return  ListApis(ctx, ac, name, filter, v.ApiHandler())
  // etc
}
```

cmd/registry/core/visitors.go
```diff
@@ -0,0 +1,116 @@
...
+	SpecRevisionHandler() SpecHandler
+	ArtifactHandler() ArtifactHandler
+}
+
+type VisitorOptions struct {
+	RegistryClient   connection.RegistryClient
+	AdminClient      connection.AdminClient
+	Pattern          string
+	Filter           string
+	AllowUnavailable bool
```
theganyo, line 42: `AllowUnavailable` is specific to Project... should we name it `AllowUnavailableProject` to make it more clear what this actually does?

timburks: Yes, sounds good, I'll make that change.

cmd/registry/core/visitors.go
timburks: thanks, I think this is evolving - for this step I wanted to touch as little as possible.

theganyo (approved):
General note: I also adopted the visitor pattern for my checker work. The code is so much cleaner and I'm very excited to see this being adopted broadly in our code! (Soon I want to see the name type case statement as well!)

timburks:
@theganyo I'd like to move this, plus the `Handler` interface and the `Get`/`List` methods out of `core` and into `pkg/visitor` (are you ok with this name?) next and test them there -- then we can work on aligning it with the tree class. I have some uses for this in `registry-experimental` and establishing it in a top-level package would be very helpful.

theganyo:
SGTM!

PR #983, opened 2023-01-27
pkg/visitor

timburks:
Move Get*(), List*(), Fetch*(), Visit(), and Handlers to pkg/visit.

### DIFF
--- a/pkg/visitor/fetch_test.go
+++ b/pkg/visitor/fetch_test.go
@@ -1,0 +1,129 @@
+// Copyright 2023 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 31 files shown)


pkg/visitor/fetch_test.go
```diff
@@ -0,0 +1,243 @@
...
+func TestFetch(t *testing.T) {
+	projectID := "fetch-test"
+	project := names.Project{ProjectID: projectID}
+	parent := project.String() + "/locations/global"
+
+	specContents := "hello"
+	artifactContents := "hello"
+
+	ctx := context.Background()
+	adminClient, err := connection.NewAdminClient(ctx)
```
theganyo, line 37: Do we want to start using `grpctest.SetupRegistry()`?

pkg/visitor/visit_test.go
```diff
@@ -0,0 +1,600 @@
...
+
+	const ProjectCount = 1
+	const ApiCount = 3
+	const VersionCount = 2
+	const DeploymentCount = 2
+	const SpecCount = 2
+	const ArtifactCount = 2
+
+	ctx := context.Background()
+	adminClient, err := connection.NewAdminClient(ctx)
```
theganyo, line 51: grpctest.SetupRegistry?

pkg/visitor/fetch_test.go
timburks: Yes, let's do some dedicated PRs for that change. I would have used it here but didn't feel that I knew the pattern well enough yet.

pkg/visitor/visit_test.go
timburks: ack, agreed

PR #988, opened 2023-01-30
labels and annotations docs

### LINKED ISSUE #977: Registry API: labels and annotations rules clarifications
Documentation is inconsistent for resource labels. The proto files document that the only special characters allowed are "underscores and dashes", but then makes explicit mention of a system prefix that includes periods and a slash:

https://github.com/theganyo/registry/blob/main/google/cloud/apigeeregistry/v1/registry_models.proto#L308-L316

This leads to a few questions about labels:

- Is a prefix separated by a slash a general concept?
- Should periods and a slashes be allowed? 
- If so, should it only be part of the prefix?
- Does the prefix count toward the length restriction?

Also, for annotations:

- Do the label's prefix rules also apply to annotations?
- A couple of basic annotations rules (#957) are being adopted, these should also be documented in the protos.

@timburks

theganyo:
Fixes #977

### DIFF
--- a/google/cloud/apigeeregistry/v1/registry_models.proto
+++ b/google/cloud/apigeeregistry/v1/registry_models.proto
@@ -71,19 +71,22 @@ message Api {
   // Label keys and values can be no longer than 64 characters
   // (Unicode codepoints), can only contain lowercase letters, numeric
   // characters, underscores and dashes. International characters are allowed.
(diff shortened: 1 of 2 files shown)


timburks (approved):
LGTM. Dropping the whole "system labels" designation seems good. This will take some time to make its way to github.com/googleapis/googleapis, but it will.

PR #990, opened 2023-01-31
allow an implicit project to be visited with ListProjects 

theganyo:
allow an implicit project to be visited with ListProjects 
also renames allowUnavailable to implicitProject

### DIFF
--- a/cmd/registry/cmd/check/tree/tree.go
+++ b/cmd/registry/cmd/check/tree/tree.go
@@ -47,7 +47,7 @@ func ListSubresources(ctx context.Context,
 	handler HandlerSet) (err error) {
 	switch name := root.(type) {
 	case names.Project:
-		err = visitor.ListProjects(ctx, adminClient, name, filter, handler.ProjectHandler())
+		err = visitor.ListProjects(ctx, adminClient, name, true, filter, handler.ProjectHandler())
 		if err != nil {
 			return err
 		}
--- a/cmd/registry/cmd/check/tree/tree.go
+++ b/cmd/registry/cmd/check/tree/tree.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/apigee/registry/pkg/connection"
 	"github.com/apigee/registry/pkg/visitor"
+	"github.com/apigee/registry/rpc"
 	"github.com/apigee/registry/server/registry/names"
 )
 
@@ -47,7 +48,8 @@ func ListSubresources(ctx context.Context,
 	handler HandlerSet) (err error) {
 	switch name := root.(type) {
 	case names.Project:
-		err = visitor.ListProjects(ctx, adminClient, name, true, filter, handler.ProjectHandler())
+		implicit := &rpc.Project{Name: "Implicit"}
+		err = visitor.ListProjects(ctx, adminClient, name, implicit, filter, handler.ProjectHandler())
 		if err != nil {
 			return err
 		}
--- a/cmd/registry/cmd/export/export.go
+++ b/cmd/registry/cmd/export/export.go
@@ -65,11 +65,11 @@ func Command() *cobra.Command {
 			}
 			// Visit the selected resources.
 			if err = visitor.Visit(ctx, v, visitor.VisitorOptions{
-				RegistryClient:          registryClient,
-				AdminClient:             adminClient,
-				Pattern:                 pattern,
-				Filter:                  filter,
-				AllowUnavailableProject: true,
+				RegistryClient:  registryClient,
+				AdminClient:     adminClient,
+				Pattern:         pattern,
+				Filter:          filter,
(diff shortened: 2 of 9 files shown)


pkg/visitor/list.go
```diff
@@ -40,6 +43,13 @@ func ListProjects(ctx context.Context,
 		Filter: filter,
 	})
 	for r, err := it.Next(); err != iterator.Done; r, err = it.Next() {
+		if err != nil && status.Code(err) == codes.Unimplemented && implicitProject {
+			// If the admin service is unavailable, provide a placeholder project.
+			// If the project is invalid, downstream actions will fail.
+			p := &rpc.Project{Name: name.String()}
```
theganyo, line 49: Questionable naming: This would likely fail validation, but the other option would be to hard-code a name. Thoughts?

timburks: Could we pass in the project name instead of a bool?

cmd/registry/cmd/check/tree/tree.go
```diff
@@ -47,7 +47,7 @@ func ListSubresources(ctx context.Context,
 	handler HandlerSet) (err error) {
 	switch name := root.(type) {
 	case names.Project:
-		err = visitor.ListProjects(ctx, adminClient, name, filter, handler.ProjectHandler())
+		err = visitor.ListProjects(ctx, adminClient, name, true, filter, handler.ProjectHandler())
```
timburks, line 50: Answering my  own question (below), you could pass name.String() or name.ProjectID here instead of bool, and we could treat an empty string as false.

pkg/visitor/list.go
theganyo: Good idea!

theganyo:
Ok. To keep things simple and symmetrical, I changed the whole visitor pattern to use a `*rpc.Project` instead of a bool. Now we have parameter lists for `GetProject()` (et al.) where we're passing in the name plus an implicitProject, which seemed initially strange to me. OTOH, after a bit of contemplation, I like how explicit it is. I feel it's really no stranger than the bool and the explicitness has a side-benefit of allowing the caller to check the result to determine if the result was implicit or actual, should that need arise.

cmd/registry/cmd/export/export.go
```diff
@@ -65,11 +65,11 @@ func Command() *cobra.Command {
...
-				RegistryClient:          registryClient,
-				AdminClient:             adminClient,
-				Pattern:                 pattern,
-				Filter:                  filter,
-				AllowUnavailableProject: true,
+				RegistryClient:  registryClient,
+				AdminClient:     adminClient,
+				Pattern:         pattern,
+				Filter:          filter,
+				ImplicitProject: &rpc.Project{Name: "Implicit"},
```
timburks, line 72: I would expect "Implicit" to fail name validation. I think we know the name that we want here, so I think we could use that instead.

timburks: oh, actually we don't, because we don't know what "pattern" is.

pkg/visitor/get.go
```diff
@@ -27,15 +27,15 @@ import (
...
 	handler ProjectHandler) error {
 	project, err := client.GetProject(ctx, &rpc.GetProjectRequest{
 		Name: name.String(),
 	})
-	if err != nil && status.Code(err) == codes.Unimplemented && allowUnavailable {
+	if err != nil && status.Code(err) == codes.Unimplemented && implicitProject != nil {
 		// If the admin service is unavailable, provide a placeholder project.
 		// If the project is invalid, downstream actions will fail.
-		project = &rpc.Project{Name: name.String()}
+		project = implicitProject
```
timburks, line 38: Here (in the deleted code), we know the name that we want. We could set implicitProject.Name to that if we wanted.

cmd/registry/cmd/export/export.go
timburks: How about leaving it as &rpc.Project{} and filling in the name when we know it? (for gets, and I guess lists could use the value that we pass in)

pkg/visitor/list.go
```diff
@@ -40,6 +43,12 @@ func ListProjects(ctx context.Context,
 		Filter: filter,
 	})
 	for r, err := it.Next(); err != iterator.Done; r, err = it.Next() {
+		if err != nil && status.Code(err) == codes.Unimplemented && implicitProject != nil {
+			// If the admin service is unavailable, provide a placeholder project.
+			// If the project is invalid, downstream actions will fail.
+			return handler(implicitProject)
```
timburks, line 49: I see: here we don't know the name that we want, so we have to trust the value set in implicitProject.

timburks (approved):
Interesting, I think this is a good step forward.

PR #991, opened 2023-01-31
Add `visitor.Unsupported` to allow easy inclusion of default visitor implementations.

timburks:
This allows visitors to be written that only support a subset of types and return errors for all others. For example:

```
type projectVisitor struct {
	visitor.Unsupported
}

func (v *projectVisitor) ProjectHandler() visitor.ProjectHandler {
	return func(message *rpc.Project) error {
           ...
	}
}

// all other handlers are inherited from Unsupported and return "unsupported" errors.
```

(Originally this was called `DefaultVisitor`)

### DIFF
--- a/pkg/visitor/default_visitor.go
+++ b/pkg/visitor/default_visitor.go
@@ -1,0 +1,74 @@
+// Copyright 2023 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package visitor
+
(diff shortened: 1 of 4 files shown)


theganyo (approved):
Love it. But can we call it an ErrorVisitor or something that indicates it's default action? I can imagine a default visitor that does nothing (perhaps a NullVisitor), for example.

timburks:
> Love it. But can we call it an ErrorVisitor or something that indicates its default action? I can imagine a default visitor that does nothing (perhaps a NullVisitor), for example.

I can see wanting different types of visitors, but I think the default should be to return errors for unsupported types -- visitors are "sent" to resources by patterns specified by users, and if a user specifies an incorrect type of resource, I think that should be an error and not fail silently.

This is like the `Unimplemented*Server` mixins that are added to gRPC servers ([example](	rpc.UnimplementedRegistryServer)). They [return errors for all unsupported methods](https://github.com/apigee/registry/blob/7d440ef0521ebae0b1511df73d91638d10af194a/rpc/registry_service_grpc.pb.go#L533)).

So how about `visitor.Unimplemented`?
```
type projectVisitor struct {
	visitor.Unimplemented
}
```

theganyo:
SGTM

timburks:
OK, I think I follow you. `Unsupported` seems closer to matching our intents - this is for mixing-in handlers for resource types that we don't want to support.

theganyo (approved):

PR #997, opened 2023-02-03
Move registry/server/names to pkg/names

timburks:
Let's encourage the use of `names` in `registry-experimental` and other external applications. It's useful, well-tested, and already is a dependency of `pkg/visitor`.


theganyo (approved):
SGTM

PR #1004, opened 2023-02-06
accept context in handlers

### LINKED ISSUE #913: Registry tool: modify Handler interfaces to accept context arguments

theganyo:
Fixes #913


timburks (approved):
Looks great, thanks. I will follow up and replace my contexts-in-structs with these now.

Issue #1010, opened 2023-02-07
Registry tool: `registry delete` bulk deletions require multiple passes.


timburks:
When deleting large collections of APIs, I found that some APIs remained after I had run `registry delete apis`. Running it again, a smaller number of APIs remained, and after a few iterations, all were gone.

I believe this is occurring because the server-side pagination is using numerical offsets, and as `registry delete` is iterating over items to delete, it begins deleting them in concurrent workers, so by the time the offset is used to get the next page, some of the APIs that were supposed to be skipped are gone, and the offset skips APIs that haven't been listed yet.

Summarizing, the registry List APIs don't support deletion of items while paginating over them. To fix this, we might:
1. Client side: collect all items to be deleted before we begin deleting them. This will slightly delay the start of deletion, particularly for large collections.
2. Client side: build iteration into `registry delete` and count the deletions in each pass, looping until no deletions occur.
3. Server side: modify list pagination to use keys (and paginate by returning results "after" the last key returned or beginning with the "next" key) and verify that this allows deletion to occur during pagination.

I think # 2 is a good short-term solution because it keeps deletion running quickly and is easier than # 1 to undo if we decide to do # 3. 

Update: # 2 is not great because subsequent iterations can return APIs before queued-up tasks have deleted them. When this happens, we try to delete the same API twice, which generates an error. So to avoid more complexity, I'll step back and use approach # 1 instead.

timburks:
Fixed by #1011.

PR #1011, opened 2023-02-07
Collect and then delete resources to work around problems listing-while-deleting.

timburks:
Addresses #1010


theganyo (approved):

theganyo:
The only issue here is our lack of tests to prove it works.

PR #1012, opened 2023-02-07
use visitor for subtree

theganyo:
subtree now integrates with visitor
fix a couple of bugs found in filter fieldmaps


pkg/visitor/subtree.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// SubtreeHandlers is a Visitor implementation that can be passed to the
+// Visit() function. It will run the handlers in Visitor against each
+// resource directly matched in Visit() as well as each resource that is
+// a descendant of those resources. (For example, if the pattern matches
+// a Version, the SubtreeVisitor will be run against all of that Versions's
+// Artifacts and Specs, and the Spec's Artifacts as well.)
+// Only the current revisions of Specs and Deployments are visited.
+type SubtreeHandlers struct {
```
timburks, line 31: Could we call this "SubtreeVisitor" instead of "SubtreeHandlers"? I see both names used in the comment and I think "SubtreeVisitor" more clearly reminds me that this is a visitor implementation that does something special (visits subtrees). Thinking out loud, I'd say this is a kind of meta-visitor, a visitor based on a visitor, that applies its underlying visitor in a specified way.

timburks (approved):
LGTM, just a comment about naming that I think makes this easier (for me) to mentally digest.

pkg/visitor/subtree.go
theganyo: Sure. It's both in the comments because I renamed it from SubtreeVisitor to SubtreeHandlers and missed a couple places. FWIW, I renamed it because "visitor" gets overloaded because of the package name stutter and we call the functions themselves handlers, but I'll change it as you suggest.

timburks: Thanks! I think it's helpful to describe this as a visitor because it implements the Visitor interface. The package name stutter doesn't bother me much, [context.Context](https://pkg.go.dev/context), of all things, is an easy-to-find example of that!

Issue #1007, opened 2023-02-07
Registry tool: eliminate `registry/core`


timburks:
With much moved to `pkg/visitor`, the `registry/core` package is relatively small and could be broken into more descriptively-named packages.

Here's a draft organization:

- create `cmd/registry/compression` for `gzip.go` and `zip.go`. This stays local to the `registry` but public so that `registry-experimental` can still use it.
- create `cmd/registry/tasks` for `tasks.go`, this is also used by `registry-experimental`.
- create `cmd/registry/labeling` for `labeling.go`, this is an odd one-off helper shared by the `annotate` and `label` commands.
- move `artifacts.go`, `specs.go`, and `verify.go` to `pkg/visitor` because these are generally useful helpers for traversing registries (the function in `specs.go` might be easily eliminated with refactoring).

Comments or alternatives?

theganyo:
Random thoughts:
- Go uses "archive" and "compress" package names for the dealing with archives and compression. 
- Tasks seems like a reasonable package
- Moving some stuff to visitor makes sense
- the labeling function is truly an odd duck. is there nowhere else it can land?

timburks:
Thanks for these suggestions. How about:
* `compress` instead of `compression`
* moving the labeling stuff into `cmd/label` and importing it from `cmd/annotate`

PR #1017, opened 2023-02-10
move remaining functions from core to pkg/visitor

timburks:
With #1016, this should finish #1007


pkg/visitor/content_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/apigee/registry/rpc"
+	"github.com/apigee/registry/server/registry/test/seeder"
+)
+
+func TestContentHelpers(t *testing.T) {
+	specName, err := names.ParseSpec("projects/content-test/locations/global/apis/a/versions/v/specs/s")
+	if err != nil {
+		t.Fatalf("failed to parse spec name %s", specName)
+	}
```
seaneganx, line 33: Parsing and checking the error is unnecessary. We know what name we want, and the error should never happen.
```
specName := names.Spec{
    ProjectID: "content-test",
    ApiID: "a",
    VersionID: "v",
    SpecID: "s",
}
```

seaneganx (approved):

pkg/visitor/content_test.go
timburks: thanks - along those lines, I think it would be nice to pull a pattern from [regexp](https://pkg.go.dev/regexp#MustCompile) and be able to do this in one line (in tests):
```
specName := names.MustParseSpec("projects/content-test/locations/global/apis/a/versions/v/specs/s")
```

PR #1026, opened 2023-02-14
Add `-o raw` support to `registry get`.

### LINKED ISSUE #1024: Registry tool: add `-o raw` to `registry get`.
With recent changes, `registry get` no longer provides a way to access all fields of resources, e.g. create/update times and spec and artifact sizes and hashes. While some of these might be appropriate to add to Registry YAML, it also seems useful to have direct access to the information returned by `registry rpc get-api-spec`, etc. with the convenience of `registry get`.

Proposed: add `-o raw` as an output mode for `registry get` and return the JSON-formatted output that is returned by the `registry rpc get-*` methods.

timburks:
Closes #1024

This adds support for `--output raw` (`-o raw`) to `registry get` and returns a JSON array of the raw message content of matching resources. Because `registry get` can return multiple values and to simplify downstream processing of the resulting JSON, results are always returned as a JSON array.

Here are some thoughts on using `raw` as the option name:
- It's a direct (JSON) representation of the API message and matches what is returned by the generated `rpc` commands.
- It is not `--output json` because I think we should save that for a JSON-equivalent of Registry YAML.
- It is not `--output message` because I think that would suggest results were the encoded protobuf messages.

### DIFF
--- a/cmd/registry/cmd/get/get.go
+++ b/cmd/registry/cmd/get/get.go
@@ -29,6 +29,8 @@ import (
 	"github.com/apigee/registry/pkg/visitor"
 	"github.com/apigee/registry/rpc"
 	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
 )
 
 func Command() *cobra.Command {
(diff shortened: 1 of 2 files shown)


theganyo:
Can you elaborate on why we wouldn't add (or optionally add) the missing fields to our YAML / JSON outputs instead of adding a new format?

timburks:
> Can you elaborate on why we wouldn't add (or optionally add) the missing fields to our YAML / JSON outputs instead of adding a new format?

With `raw` output, there's no expectation of round-tripping, so we don't have to figure out how to handle dealing with output-only fields like hash, sizeBytes, and the creation/update times in YAML. So why this? Mainly because it's easier... it might be a temporary gap-filler until we do the YAML export work, but even if we keep `raw` output support, the implementation and support cost seems small to me.

theganyo (approved):
Seems reasonable.

PR #1034, opened 2023-02-21
Pass filenames of patch files to all patch helpers.

timburks:
When `registry apply` uploads spec files, it looks for them in the same directory as the YAML file that referenced them, i.e. the specs should be alongside the YAML. This requires us to pass the path to the containing YAML file all the way down to the spec handler. Previously YAML representations of specs were getting the path name, but higher-level nested representations (APIs and versions) did not have their filenames passed down during processing. This fixes that and passes the containing filename down to all handlers, including artifact handlers, which don't need them now but would use them if/when we support importing external (binary?) files as artifact contents.

### DIFF
--- a/cmd/registry/patch/version.go
+++ b/cmd/registry/patch/version.go
@@ -83,13 +83,14 @@ func applyApiVersionPatchBytes(
 	ctx context.Context,
 	client connection.RegistryClient,
 	bytes []byte,
-	project string) error {
+	project string,
+	filename string) error {
 	var version models.ApiVersion
 	err := yaml.Unmarshal(bytes, &version)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


cmd/registry/patch/version.go
```diff
@@ -129,13 +131,13 @@ func applyApiVersionPatch(
 		return fmt.Errorf("UpdateApiVersion: %s", err)
 	}
 	for _, specPatch := range version.Data.ApiSpecs {
-		err := applyApiSpecPatch(ctx, client, specPatch, name.String(), "")
```
timburks, line 132: This was the problem, which @seaneganx recently observed was breaking spec uploads from YAML.

theganyo (approved):

PR #1036, opened 2023-02-22
Move artifact generated support code into pkg/application/*

timburks:
Updated to follow #1037 (this is addressing item 2 of 4 listed there):

I think it would be helpful to separate the generated artifact support code from the generated code for the API. For one thing, this would encourage external applications to use these types when building YAML with the `models` package, and then we could set the polymorphic `Artifact.Data` field with something like this:

```
rl := &apihub.ReferenceList{...}
a := &models.Artifact{
   ...
   Data: models.ArtifactData(rl) // this function would convert a proto.Message into a yaml.Node
}
```

More generally, this would also make it easier to replace or move these artifact types again if that seems appropriate.

Questions answered in #1037:
- Should this be part of another package (`pkg/types` or `pkg/models`)? I think not because I think we dislike mixing generated and handwritten code. For example, feedback from @seaneganx led to the creation of [cmd/registry/cmd/rpc/generated](https://github.com/apigee/registry/tree/main/cmd/registry/cmd/rpc/generated). 
- Should this be named something else? Like maybe [a singular `pkg/artifact` vs. `pkg/artifacts`](https://rakyll.org/style-packages/#:~:text=In%20go%2C%20package%20names%20are,IT%2C%20USE%20SINGULAR%20FORM!!)? If we do make this singular, we should probably fix the other package names for consistency (which I wouldn't mind). We could also further subdivide these types into separate packages for each functional area, but I'm inclined to leave that for later.


theganyo:
> Should this be part of another package (`pkg/types` or `pkg/models`)? I think not because I think we dislike mixing generated and handwritten code. For example, feedback from @seaneganx led to the creation of [cmd/registry/cmd/rpc/generated](https://github.com/apigee/registry/tree/main/cmd/registry/cmd/rpc/generated).

Since `types` is also specifically for artifacts, we could maybe reorganize like this?

- pkg/artifact
   - pkg/artifact/types
   - pkg/artifact/models


> Should this be named something else? Like maybe [a singular `pkg/artifact` vs. `pkg/artifacts`](https://rakyll.org/style-packages/#:~:text=In%20go%2C%20package%20names%20are,IT%2C%20USE%20SINGULAR%20FORM!!)? If we do make this singular, we should probably fix the other package names for consistency (which I wouldn't mind). We could also further subdivide these types into separate packages for each functional area, but I'm inclined to leave that for later.

I'm ok with changing to singular to match style.

timburks:
After a brief walk and some thought, #1037 is a redraft that addresses more Go style guidance.

theganyo (approved):

timburks:
Generally I think this is a nice change, and I don't regret the names of any of the `pkg/application/` subdirectories, but their names collided with a few internal packages like `cmd/registry/cmd/controller/scoring`. Whenever conflicts occurred in files, I imported the `application` package to a name  with `_message` appended, e.g. `scoring_message` (I have no affection for this suffix).

PR #1044, opened 2023-02-23
Move YAML encoding support code to pkg/encoding.

### LINKED ISSUE #1041: Registry packages: Add support code to `pkg/encoding`
As proposed in #1036, add a helper for converting proto messages to YAML nodes:
```
func NodeForMessage(message proto.Message) (yaml.Node, error)
```

Also move [this code](https://github.com/apigee/registry/blob/bdfd66836a91ec41db7b1ce5d5dd9cf76d4c42e4/cmd/registry/patch/yaml.go#L32) from `cmd/registry/cmd/patch` to `pkg/encoding`... but `Encode` probably isn't the right name for this, maybe something with "YAML" in it (`EncodeYAML`)?

timburks:
This will close #1041.

See `TestYAML()` in `pkg/encoding/yaml_tests.go` for a simple usage example.


timburks:
Looking forward, I think we could add constructors for `Artifact`, etc. that automatically set the `ApiVersion` field and possibly others.

theganyo (approved):

PR #1046, opened 2023-02-24
Support for field definition and storage.

### LINKED ISSUE #1048: Registry tool: representation for groups of fields.
We've discussed and heard of applications where users would want to add custom fields to a registry in ways that our existing `labels` and `annotations` don't address.

- Users would like to control visibility of different groups of values based on relevance or need-to-know considerations. If these are stored in `labels` or `annotations` of the main resources, these would be difficult to segment in this way.
- Similarly, users would like to govern control of these fields, and putting them in separate artifacts would make it easier for different tools or contributors to make updates without collisions.
- Users would like to add a *lot* of fields, perhaps too many for convenient storage in `labels` or `annotations`.
- `labels` put length and character restrictions on their contents that generally make them unsuitable for arbitrary field storage.

Based on the above, I think it would be good to define an artifact type for storing groups of fields in the registry, and allow instances of this type to be attached to any of the main resources in the registry (projects, apis, versions, specs, deployments). We might have two types: 1) a per-instance type that stores values, and 2) a definition type that stores text to be displayed alongside values (title and description) and any constraints on attribute values (formats or allowed values). These could then be created and displayed by tools in standard ways, and this would also make it easier to index groups of attributes for searches.

`Facet` seemed like a good word for this, with "facet" defined as "a particular aspect or feature of something." `Facets` would be collections of attribute values stored on resources with an associated `FacetDefinition` stored at the project level. `Facets` could (at user discretion) be dedicated to specific groupings of information and independently generated by different tools or importers. However, favoring literal definitions, we're going with `FieldSet` and `FieldSetDefinition`.

timburks:
This adds two new `apihub` artifact types: `FieldSet` and `FieldSetDefinition`. `FieldSet` is a group of fields, typically related to a particular domain or application and stored per-resource. `FieldSetDefinition` provides display-related and format-specific information and is stored at the project level.

Closes #1048


google/cloud/apigeeregistry/v1/apihub/fields.proto
```diff
@@ -0,0 +1,86 @@
...
+    string id = 1;
+
+    // The display_name of the field, used when the field is displayed.
+    string display_name = 2;
+
+    // A description of the field, possibly displayable as a tooltip.
+    string description = 3;
+
+    // Optional string specifying the field format.
+    string format = 4;
```
theganyo, line 82: Is this just for display?

google/cloud/apigeeregistry/v1/apihub/fields.proto
```diff
@@ -0,0 +1,86 @@
...
+
+  // A description of the field set.
+  string description = 4;
+
+  // Full resource name of an FieldSetDefinition artifact that
+  // describes this field set.
+  string definition_name = 5 [(google.api.field_behavior) = REQUIRED];
+
+  // The field values, stored using field ids as keys.
+  map<string,string> values = 6;
```
theganyo, line 49: Is a field set essentially a display grouping? If so, should we consider separating the display metadata from the data?

google/cloud/apigeeregistry/v1/apihub/fields.proto
timburks: I don't think so. Since values are coming in from YAML, we can't have significant type constraints there, but here we could require values to be numbers, dates, URLs, email addresses, etc.

google/cloud/apigeeregistry/v1/apihub/fields.proto
timburks: There is display metadata in the FieldSetDefinition - if you mean that the grouping itself is metadata, that seems to get burdensome for applications (viewers) that would have to fetch all these things individually. But we might consider this like a ScoreCard that is derived from individual artifacts.

However, the overhead of storing individual fields as separate artifacts is pretty high, so storing fields in groups seems more practical.

I like to think of these field sets as conceptually-related groups of fields, but that's up to applications and users to eventually decide.

google/cloud/apigeeregistry/v1/apihub/fields.proto
theganyo: Ok, should we define how this field operates... or is it a TODO?

google/cloud/apigeeregistry/v1/apihub/fields.proto
theganyo: I guess what's confusing me is the `display_name` and `description` fields here.

theganyo: Are these just for convenience in browsing artifacts?

google/cloud/apigeeregistry/v1/apihub/fields.proto
timburks: TODO: I'd like to defer that until we have some uses.

google/cloud/apigeeregistry/v1/apihub/fields.proto
timburks: Maybe it's overkill, but we have these fields for Scores, which is somewhat similar: https://github.com/apigee/registry/blob/a2537feb98ba82b98607d6a9e14519a1455afb3f/google/cloud/apigeeregistry/v1/scoring/score.proto#L41

They are instance-specific values, which, at scale, we probably won't want to be using.

timburks: I've removed display_name and description from FieldSet. We can add them later if we find a need for them.

google/cloud/apigeeregistry/v1/apihub/fields.proto
theganyo: If we don't know how this will work, perhaps we should leave it out for now?

timburks: I wouldn't mind that, but `allowed_values` is already a requested field (people want to use it), and I think logically, `format` should precede that in the list of proto fields.

theganyo: Ok... Well, let's at least document that it's not used currently and may be a reserved format.

timburks: ok, done.

theganyo (approved):

PR #1049, opened 2023-02-24
Add a script that installs the latest release of the registry tool.

### LINKED ISSUE #1043: Registry tool: installer
We would like to have a one-line installer for the `registry` tool, and possibly also related plugins and tools.

Examples:
`curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs/ | sh`
`curl -L https://raw.githubusercontent.com/apigee/apigeecli/main/downloadLatest.sh | sh -`

timburks:
Fixes #1043, but this only installs `registry`. We will probably also want to install the linter plugins, but these currently aren't included in release builds.

### DIFF
--- a/downloadLatest.sh
+++ b/downloadLatest.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script downloads and installs the latest version of the registry tool.
+# The binary is installed in the user's home directory in $HOME/.registry/bin.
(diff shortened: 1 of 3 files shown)


downloadLatest.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
theganyo, line 1: I assume we can even just make this `#!/bin/sh` since the README command is using `sh`.

downloadLatest.sh
```diff
@@ -0,0 +1,100 @@
...
+rm -r "${tmp}"
+
+# Print message
+printf "\n"
+printf "Add the registry to your path with:"
+printf "\n"
+printf "  export PATH=\$PATH:\$HOME/.registry/bin \n"
+printf "\n"
+
+export PATH=$PATH:$HOME/.registry/bin
```
theganyo, line 100: JFYI: This won't actually export the PATH change to the calling shell when running the script piped through `sh`.

theganyo (approved):
I think we should include the install directions at or near the top of the root README.md as well.

downloadLatest.sh
timburks: Right, I kept that from https://raw.githubusercontent.com/apigee/apigeecli/main/downloadLatest.sh, but since it's so annoying to have this script call exit when I source it, I'll take this out so we don't encourage that.

downloadLatest.sh
timburks: Thanks, I'll switch this.

theganyo (approved):

PR #1050, opened 2023-02-27
Run ./tools/pretty.sh to format README.md files.

timburks:
This mainly adjusts line lengths and removes trailing characters. I try to remember to run it before commits but don't always. If we put it in the Makefile under the main target, builds will require a local installation of prettier (I think that's too burdensome for new contributors).


theganyo:
We can install prettier into VSCode. https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

PR #1062, opened 2023-03-02
Replace wildcards with specific file names in download script.

timburks:
Running the latest download script with `v0.6.6`, I get the following error on Linux:
```
$ sh downloadLatest.sh 
https://github.com/apigee/registry/releases/download/v0.6.6/registry_0.6.6_linux_amd64.tar.gz

Downloading registry_v0.6.6 from https://github.com/apigee/registry/releases/download/v0.6.6/registry_0.6.6_linux_amd64.tar.gz ...

registry v0.6.6 Download Complete!

mv: cannot stat '/tmp/registry.PaMH31/registry-lint-*': No such file or directory
```
It appears that the globbing is not being expanded properly. This replaces the * with the specific filenames that we want to install.

### DIFF
--- a/downloadLatest.sh
+++ b/downloadLatest.sh
@@ -84,10 +84,12 @@ printf "\n"
 cd "$HOME" || exit
 mkdir -p "$HOME/.registry/bin"
 mv "${tmp}/registry" "$HOME/.registry/bin"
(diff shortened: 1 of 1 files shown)


theganyo (approved):
That's unfortunate. I was hoping we wouldn't have to update this each time we add a plugin.

Issue #470, opened 2022-02-26
Registry packages: The names package should include a representation of locations


timburks:
Currently the names package includes representations of project names and API names, but in between these names, there are location entities that are not represented. Location support was added after the package was created and was only incompletely added to project name parsing. But in some applications, we would like to represent the name of a "parent" resource, which for APIs and some artifacts, is a location.

PR #1068, opened 2023-03-02
Add benchmark that lists artifacts under each resource type.

timburks:
This adds a benchmark test that lists artifacts under different resource types.

I've casually observed that listing artifacts under specs takes longer than listing artifacts under APIs, and guessed that this might be related to revision support. The test here creates N=100 APIs, each with one version, spec, and deployment, and with three artifacts ("a", "b", and "c") under each resource. Three revisions of each spec and deployment are created with artifacts only associated with the last ones created. Then we list all of the "b" artifacts under each resource type.

Running locally with Postgres on my Chromebook, I don't see much difference in listing times:
```
BenchmarkListArtifacts
BenchmarkListArtifacts/ListApiArtifacts
BenchmarkListArtifacts/ListApiArtifacts-8                     75          14497073 ns/op
BenchmarkListArtifacts/ListVersionArtifacts
BenchmarkListArtifacts/ListVersionArtifacts-8                 51          20340059 ns/op
BenchmarkListArtifacts/ListSpecArtifacts
BenchmarkListArtifacts/ListSpecArtifacts-8                    50          22871343 ns/op
BenchmarkListArtifacts/ListDeploymentArtifacts
BenchmarkListArtifacts/ListDeploymentArtifacts-8              54          19693489 ns/op
```

If I increase the number of APIs to N=1000, a distinction is more clear:
```
BenchmarkListArtifacts
BenchmarkListArtifacts/ListApiArtifacts
BenchmarkListArtifacts/ListApiArtifacts-8                      3         418735575 ns/op
BenchmarkListArtifacts/ListVersionArtifacts
BenchmarkListArtifacts/ListVersionArtifacts-8                  2         551807149 ns/op
BenchmarkListArtifacts/ListSpecArtifacts
BenchmarkListArtifacts/ListSpecArtifacts-8                     2         807084090 ns/op
BenchmarkListArtifacts/ListDeploymentArtifacts
BenchmarkListArtifacts/ListDeploymentArtifacts-8               2         743311222 ns/op
```

This doesn't seem to be as severe as it subjectively seemed to me, but this is local and my observations were running with a remote database.

Setup time dominates the runtime of these tests, so this PR keeps N=100.

### DIFF
--- a/tests/benchmark/benchmark_apis_test.go
+++ b/tests/benchmark/benchmark_apis_test.go
@@ -30,7 +30,7 @@ func apiId(i int) string {
 }
 
 func apiName(apiId string) string {
-	return fmt.Sprintf("%s/apis/%s", root(), apiId)
+	return fmt.Sprintf("%s/apis/%s", root().String()+"/locations/global", apiId)
 }
 
 func getApi(b *testing.B, ctx context.Context, client connection.RegistryClient, apiId string) error {
@@ -41,26 +41,13 @@ func getApi(b *testing.B, ctx context.Context, client connection.RegistryClient,
 
 func listApis(b *testing.B, ctx context.Context, client connection.RegistryClient) error {
 	b.Helper()
-	it := client.ListApis(ctx, &rpc.ListApisRequest{Parent: root()})
+	it := client.ListApis(ctx, &rpc.ListApisRequest{Parent: root().String() + "/locations/global"})
 	for _, err := it.Next(); err != iterator.Done; _, err = it.Next() {
 		if err != nil {
 			return err
 		}
 	}
 	return nil
-}
-
-func createApi(b *testing.B, ctx context.Context, client connection.RegistryClient, apiId string) error {
-	b.Helper()
-	_, err := client.CreateApi(ctx, &rpc.CreateApiRequest{
-		Parent: root(),
-		ApiId:  apiId,
-		Api: &rpc.Api{
-			DisplayName: apiId,
-			Description: fmt.Sprintf("Description for %s", apiId),
-		},
-	})
-	return err
 }
 
 func updateApi(b *testing.B, ctx context.Context, client connection.RegistryClient, apiId string) error {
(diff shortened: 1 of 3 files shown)


tests/benchmark/benchmark_apis_test.go
```diff
@@ -41,7 +41,7 @@ func getApi(b *testing.B, ctx context.Context, client connection.RegistryClient,
 
 func listApis(b *testing.B, ctx context.Context, client connection.RegistryClient) error {
 	b.Helper()
-	it := client.ListApis(ctx, &rpc.ListApisRequest{Parent: root()})
+	it := client.ListApis(ctx, &rpc.ListApisRequest{Parent: root().String() + "/locations/global"})
```
theganyo, line 44: Apropos of nothing, every time I see this omnipresent "/locations/global" constant in our code it sends me down a rabbit hole of wondering how we could eliminate it globally.

theganyo (approved):

tests/benchmark/benchmark_apis_test.go
timburks: I think we should do something in the names package. #470?

timburks:
Here's a pretty interesting result from benchmarking a networked test server running with Postgres:
```
BenchmarkListArtifacts/ListApiArtifacts-16         	      38	  30631349 ns/op
BenchmarkListArtifacts/ListVersionArtifacts-16     	      30	  37002730 ns/op
BenchmarkListArtifacts/ListSpecArtifacts-16        	       1	2729463578 ns/op
BenchmarkListArtifacts/ListDeploymentArtifacts-16  	       8	 125910637 ns/op
```

`registry rpc admin get-status` returns this for the `vcs.revision` of my build: [891333a8615e69ee3a2c6043116b78e5ba70f5b7](https://github.com/apigee/registry/commit/891333a8615e69ee3a2c6043116b78e5ba70f5b7), so it's a fairly fresh server.

The server has a couple of big projects, total database usage is summarized here:
```
$ registry rpc admin get-storage --json
{
  "description": "postgres",
  "collections": [
    {
      "name": "apis",
      "count": "5894"
    },
    {
      "name": "artifacts",
      "count": "34448"
    },
    {
      "name": "blobs",
      "count": "47555"
    },
    {
      "name": "deployment_revision_tags"
    },
    {
      "name": "deployments"
    },
    {
      "name": "projects",
      "count": "6"
    },
    {
      "name": "spec_revision_tags",
      "count": "1"
    },
    {
      "name": "specs",
      "count": "13107"
    },
    {
      "name": "versions",
      "count": "9211"
    }
  ]
}
```

timburks:
#1069

PR #1073, opened 2023-03-03
When `registry get` matches nothing, print "Not Found" to stderr and don't report an error.

### LINKED ISSUE #1072: Registry tool: when `registry get` finds no resources, it reports an error.
This seems harsh and the large usage message makes it hard to see what actually happened:
```
$ registry get apis/-/versions/-/specs/-/artifacts
Error: no matching results found
Usage:
  registry get PATTERN [flags]

Flags:
      --filter string   filter selected resources
      --nested          include nested subresources in YAML output
  -o, --output string   output type (name|yaml|contents) (default "name")

Global Flags:

----
Need more help?
```
I think it would be better to just return nothing.
```
$ registry get apis/-/versions/-/specs/-/artifacts
$ 
```
I think this is analogous to running `ls` in an empty directory.
```
$ ls
$ 
```

timburks:
Fixes #1072


timburks:
When we try to get something specific that doesn't exist, we still get an error:
```
$ registry get apis/a/versions/foo/specs/z
Error: rpc error: code = NotFound desc = "projects/menagerie/locations/global/apis/a/versions/foo/specs/z" not found in database
Usage:
  registry get PATTERN [flags]

Flags:
      --filter string   filter selected resources
      --nested          include nested subresources in YAML output
  -o, --output string   output type (name|yaml|contents) (default "name")

Global Flags:

----
Need more help?
```
I think that's appropriate, but the usage message seems a bit much.

theganyo (approved):
I agree.

cmd/registry/cmd/get/get.go
```diff
@@ -75,6 +77,9 @@ func Command() *cobra.Command {
 				Pattern:        pattern,
 				Filter:         filter,
 			}); err != nil {
+				if status.Code(err) == codes.NotFound {
+					cmd.SilenceUsage = true
```
timburks, line 81: We might silence usage for other cases, but this does so for specific resource names that aren't found.
```
% registry get apis/missing          
Error: rpc error: code = NotFound desc = "projects/menagerie/locations/global/apis/missing" not found in database
```

We still print usage for invalid patterns. 
```
% registry get invalid     
Error: unsupported pattern projects/menagerie/locations/global/invalid
Usage:
  registry get PATTERN [flags]

Flags:
      --filter string   filter selected resources
      --nested          include nested subresources in YAML output
  -o, --output string   output type (name|yaml|contents) (default "name")

Global Flags:
----
Need more help?
```

cmd/registry/main.go
```diff
@@ -32,10 +33,21 @@ func main() {
...
+Need more help?
+https://github.com/apigee/registry/wiki
+`)
+		return err
+	})
+
 	if err := cmd.ExecuteContext(ctx); err != nil {
-		fmt.Println("----")
-		fmt.Println("Need more help?")
-		fmt.Println("https://github.com/apigee/registry/wiki")
```
timburks, line 38: Unfortunately `cmd.SilenceUsage` isn't helpful here. `cmd` is the top-level command, and in our implementation, we set it on the get command. It seemed that the best way to silence this was to add it to the `UsageString()` with the code above.

theganyo: Hmm. I'll have to think about this more, but first blush is that maybe we're not properly handling what should be returned as an error if this is a problem here.

timburks: "Should all errors print usage?" seems to be the key question for us to answer.

theganyo: The reason for adding the "Need more help" was to provide help to users when they encounter errors, so I would argue that it is good to do so.  For anything that is potentially expected and not an *error*, we should probably handle within the command. In this case, we've decided that NotFound is not actually an error, just a valid result.

theganyo (approved):
Approving as I don't want to hold this improvement up and the proper error handling / reporting discussion probably deserves its own space.

cmd/registry/main.go
timburks: This is messy. I'm not saying that I want to match this, but the generated `rpc list-` commands return an error when no items match a list. 
```
registry rpc list-apis --parent projects/test/locations/global --filter "api_id == '12'"

Error: no more items in iterator
Usage:
  registry rpc list-apis [flags]
```

Similarly, getting a non-existing resource returns an error in the `rpc` commands.
```
$ registry rpc get-api --name projects/test/locations/global/apis/missing
Error: rpc error: code = NotFound desc = "projects/test/locations/global/apis/missing" not found in database
```

So while I'm not sure that we want to match this, I think not matching it should be for a good reason. Turning off the distracting and irrelevant usage information is one reason (if you ask for a resource that doesn't exist, I don't think you need to be reminded of the usage information, and the usage info gets in the way of seeing what actually happened). I think returning an error would also affect the return value of the `registry` tool, which could be significant in scripts. Do we want to return OK when the `registry` tool is run to `get` something that doesn't exist? I think that might be a bad idea.

timburks: Rambling on about this, I think we could narrow this to just being about `registry get` (which it is) and we could say that `registry get` is a command that returns something if it exists and nothing if nothing matches (with no error) and that it only errors on bad inputs like invalid filters or ...invalid patterns? If I forget to include "/locations/global" in a full path or leave out the collection identifiers ("apis", "versions", etc), then the underlying Visitor code should detect this and report an error (as it should, I think).

theganyo: My argument was just that the root of problem is that each command may have different interpretations of what an actual "error" should be... and if an error isn't handled and bubbles up to the top level, it should probably go ahead and print usage information for that command to be of help the user. (Which is the default for cobra.) NotFound seems to be in a No Man's Land of "is it an user error or not" and is pretty common... so I'm good with just checking for that and moving on.

timburks: OK, I agree that we'll treat NotFound as not-an-error in `registry get`, so we can leave everything else as is.

cmd/registry/cmd/get/get.go
```diff
@@ -75,9 +77,19 @@ func Command() *cobra.Command {
 				Pattern:        pattern,
 				Filter:         filter,
 			}); err != nil {
+				if status.Code(err) == codes.NotFound {
+					fmt.Fprintln(cmd.ErrOrStderr(), "Not Found")
```
theganyo, line 81: This is an interesting detail as well... normally this cmd emits yaml. If this cmd is being used for scripting, should the output be nothing? Or just the yaml header? It seems that non-yaml might not be the right choice?

theganyo: Never mind, this is printed to stderr.

theganyo (approved):

cmd/registry/cmd/get/get.go
timburks: Right, I thought we should keep stdout clear here.

PR #1074, opened 2023-03-03
Unhide generated rpc commands for Admin and Provisioning methods.

timburks:
Hiding these seemed to me to be a good idea (at the time), but now it seems to be more confusing to have hidden commands than to explain the differences between hosted and non-hosted installations.


theganyo (approved):
I agree hiding commands is confusing. I also feel it's inherently confusing to have a CLI with two interfaces where some commands work for the API Hub product and others for the OS project. It probably doesn't help that we don't actually use the name "Google API Hub" and instead say "Google-hosted installations only" - which could be (?) confused with deploying an API Registry on Google cloud. Maybe the "provisioning" submenu should be directly named "apihub"?

PR #1076, opened 2023-03-07
instrument for prometheus

### LINKED ISSUE #1066: Registry server: instrument for prometheus
Add instrumenting for host, golang, and endpoints

theganyo:
Closes #1066

Here's a very basic prometheus integration that exposes Golang VM metrics, grpc call counters, and optionally grpc call histograms (using default buckets of .005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10 seconds). We can, of course, modify or extend these metrics as desired to include just about anything.

To try it out, just run the server (I enabled stats by default for now) , make a request or two, and `curl http://localhost:9090/metrics` to see the raw metrics. (This endpoint would normally be scraped by a prometheus server for analysis.)


timburks:
Looks great - does "I enabled stats by default for now" mean that you're planning to change this so that monitoring is optionally enabled (by a registry-server config setting)? I think I'd prefer that to minimize risk, but what to other Prometheus-enabled services usually do? (do they keep this on by default?)

timburks (approved):

theganyo:
Yes, that is the intent. Both of these would be explicitly enabled by config:
* prometheus stats + endpoint
* prometheus histograms

See:
```
	Monitoring: MonitoringConfig{
		Enable:             true, // TODO: actual s/b false
		EnableRequestTimes: true, // TODO: actual s/b false
```

theganyo:
Per offline convo, removed `EnableRequestTimes` option.

timburks (approved):

PR #1079, opened 2023-03-10
Automatically infer latest spec revision when importing deployments

### LINKED ISSUE #1078: Registry tool: fix `registry apply` to automatically infer spec revision references in deployment YAML
We've had an annoying difficulty when importing YAML for deployments - the deployments expect to refer to a spec revision, but when we're adding a spec in the same import, we don't know what revision id to use.

However, [here](https://github.com/apigee/registry/blob/b9f77d1620dda3f01f154d5da753519a2a5524de/cmd/registry/patch/deployment.go#L46) in the import code, we know the name of the spec we want to associate with the deployment, and if it doesn't include a revision ID, we can look up the latest revision id as long as that spec revision has already been created, so we just also have to make sure that we create spec resources (in our implementation) before we create deployments.

timburks:
Fixes #1078 

tests are coming in this PR

### DIFF
--- a/cmd/registry/patch/deployment.go
+++ b/cmd/registry/patch/deployment.go
@@ -38,8 +38,9 @@ func relativeSpecRevisionName(apiName names.Api, spec string) string {
 	return spec
 }
 
-// optionalSpecRevisionName returns a spec revision name if the subpath is not empty
-func optionalSpecRevisionName(ctx context.Context, client *gapic.RegistryClient, deploymentName names.Deployment, subpath string) (string, error) {
+// resolveSpecRevisionName returns a "full-resolved" spec revision name
+// that is relative to the root ("/") and includes a revision ID.
+func resolveSpecRevisionName(ctx context.Context, client *gapic.RegistryClient, deploymentName names.Deployment, subpath string) (string, error) {
 	if subpath == "" {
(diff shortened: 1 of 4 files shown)


cmd/registry/patch/deployment.go
```diff
@@ -39,11 +39,27 @@ func relativeSpecRevisionName(apiName names.Api, spec string) string {
 }
 
 // optionalSpecRevisionName returns a spec revision name if the subpath is not empty
-func optionalSpecRevisionName(deploymentName names.Deployment, subpath string) string {
+func optionalSpecRevisionName(ctx context.Context, client *gapic.RegistryClient, deploymentName names.Deployment, subpath string) (string, error) {
```
theganyo, line 42: As this is quite a vague function name, we should make sure the comment fully documents what to expect.

theganyo (approved):

cmd/registry/patch/deployment.go
timburks: yes, it should probably be renamed. I'll suggest something in the next commit.

timburks: `resolveSpecRevisionName` with a better comment.

PR #1086, opened 2023-03-14
ensure cmd use documents positional args

### LINKED ISSUE #167: Registry tool: subcommand usage information should include required positional arguments
Several `registry` subcommands require positional arguments (directories for the bulk upload commands, for example). These currently aren't included in usage information.

Related: we should follow a convention for this. I think `kubectl` is a good guide:

```
Usage:
  kubectl label [--overwrite] (-f FILENAME | TYPE NAME) KEY_1=VAL_1 ... KEY_N=VAL_N [--resource-version=version]
[options]
```

Note that argument values are upper-cased.

theganyo:
Fixes #167

### DIFF
--- a/cmd/registry/cmd/compute/score/score.go
+++ b/cmd/registry/cmd/compute/score/score.go
@@ -30,7 +30,7 @@ import (
 
 func Command() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "score RESOURCE",
+		Use:   "score PATTERN",
 		Short: "Compute scores for APIs and API specs",
 		Args:  cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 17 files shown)


timburks (approved):
Thanks! Beyond this, it might be good to put something in [root_test.go](https://github.com/apigee/registry/blob/main/cmd/registry/cmd/root_test.go) to  let us know when new ARGUMENT names appear that aren't in our preferred vocabulary.

cmd/registry/cmd/compute/score/score.go
```diff
@@ -30,7 +30,7 @@ import (
 
 func Command() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "score",
+		Use:   "score RESOURCE",
```
timburks, line 33: Maybe this should be PATTERN, but we can revisit that.

theganyo: No, you're right, PATTERN is correct.

theganyo:
> Thanks! Beyond this, it might be good to put something in [root_test.go](https://github.com/apigee/registry/blob/main/cmd/registry/cmd/root_test.go) to let us know when new ARGUMENT names appear that aren't in our preferred vocabulary.

This seem non-trivial, I'll add a new issue for it.

PR #1087, opened 2023-03-14
add default namespace "registry" to config props

### LINKED ISSUE #894: Registry tool (configuration): `registry config set registry.project` feels awkward
I don't think we need to change the variable names, but wonder if we could support a default `registry.` prefix that is used when it's unspecified?

`registry config set project`
`registry config set address`
`registry config set insecure`

theganyo:
Fixes #894


timburks (approved):

timburks:
\o/ "registry config set project" - thanks!

PR #1092, opened 2023-03-14
derive and encode openapi version in mimetype

### LINKED ISSUE #973: Registry tool: `registry upload csv` assumes all specs are OpenAPI  3.0.0
As noted on [registry upload csv](https://github.com/apigee/registry/wiki/registry-upload-csv), the mime type for uploaded specs is currently hard-coded. We might at least allow it to be specified with an option, but it's also probably easy enough to read it out of the document.

theganyo:
Fixes #973

### DIFF
--- a/cmd/registry/cmd/upload/csv.go
+++ b/cmd/registry/cmd/upload/csv.go
@@ -30,6 +30,8 @@ import (
 	"github.com/apigee/registry/pkg/names"
 	"github.com/apigee/registry/pkg/visitor"
 	"github.com/apigee/registry/rpc"
+	oas2 "github.com/google/gnostic/openapiv2"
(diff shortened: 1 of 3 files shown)


timburks (approved):
This works and is ok for now, but I think we should use yaml.v3 for this instead of gnostic to be more permissive of spec errors (gnostic will return an error if specs aren't valid) and future version changes (gnostic doesn't support 3.1 and I'm not sure when it would). YAML-reading is more robust and we're just checking for the presence and value of a field.

theganyo:
Interesting. I chose to use gnostic specifically _because_ it validates the specs. If we just look for a key in yaml, it's not really a known spec version, IMO.

PR #1099, opened 2023-03-16
make registry apply accept a list of files

### LINKED ISSUE #1097: Registry tool: can `registry apply` accept a list of files?
It would be nice to be able to do this:
```
registry apply -R -f starbucks.com  wordnik.com
```

theganyo:
Fixes #1097

### DIFF
--- a/cmd/registry/cmd/apply/apply.go
+++ b/cmd/registry/cmd/apply/apply.go
@@ -25,16 +25,23 @@ import (
 )
 
 func Command() *cobra.Command {
-	var fileName string
 	var project string
 	var recursive bool
 	var jobs int
 	cmd := &cobra.Command{
-		Use:   "apply (-f FILE | -f -)",
+		Use:   "apply PATH_1 ... PATH_1 | -",
 		Short: "Apply YAML to the API Registry",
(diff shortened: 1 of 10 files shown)


cmd/registry/cmd/apply/apply.go
```diff
@@ -25,16 +25,23 @@ import (
 )
 
 func Command() *cobra.Command {
-	var fileName string
 	var project string
 	var recursive bool
 	var jobs int
 	cmd := &cobra.Command{
-		Use:   "apply (-f FILE | -f -)",
+		Use:   "apply PATH_1 ... PATH_1 | -",
```
timburks, line 32: oops, I think this is deviating from kubectl (https://kubernetes.io/docs/reference/kubectl/cheatsheet/), which seems to take multiple `-f` options. I was careless about that in the original issue... I don't love what kubectl does, but think it's easier to point to it as precedent than explain why we differ.

timburks: sorry, didn't mean to shut this down -- could we accept multiple `-f` flags? I think kubectl does that

theganyo: Unsure. I can look into it if you'd like. My own impression was that this command breaks precedence with all of our other commands which is why I changed it.

PR #1101, opened 2023-03-16
make registry apply accept a list of files

theganyo:
Fixes https://github.com/apigee/registry/issues/1097

### DIFF
--- a/cmd/registry/cmd/apply/apply.go
+++ b/cmd/registry/cmd/apply/apply.go
@@ -25,19 +25,23 @@ import (
 )
 
 func Command() *cobra.Command {
+	var files []string
 	var project string
 	var recursive bool
 	var jobs int
 	cmd := &cobra.Command{
-		Use:   "apply PATH_1 ... PATH_1 | -",
+		Use:   "apply (-f FILE | -f -)",
 		Short: "Apply YAML to the API Registry",
-		Long:  "Apply YAML to the API Registry by file / folder name or stdin. Resources will be created if they don't exist yet.\n\nMore info and example usage at https://github.com/apigee/registry/wiki/registry-apply",
-		Args:  cobra.MinimumNArgs(1),
+		Long: "Apply YAML to the API Registry by files / folder names or stdin. " +
+			"Resources will be created if they don't exist yet. " +
+			"Multiple files may be specified in -f separated by commas or by repeating the -f flag." +
+			"\n\nMore info and example usage at https://github.com/apigee/registry/wiki/registry-apply",
+		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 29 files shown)


cmd/registry/cmd/apply/apply.go
```diff
@@ -25,16 +25,27 @@ import (
...
 	var recursive bool
 	var jobs int
 	cmd := &cobra.Command{
 		Use:   "apply (-f FILE | -f -)",
 		Short: "Apply YAML to the API Registry",
-		Long:  "Apply YAML to the API Registry by file name or stdin. Resources will be created if they don't exist yet.\n\nMore info and example usage at https://github.com/apigee/registry/wiki/registry-apply",
-		Args:  cobra.NoArgs,
+		Long: "Apply YAML to the API Registry by files / folder names or stdin. " +
+			"Resources will be created if they don't exist yet. " +
+			"Multiple files may be specified in -f separated by commas or by repeating the -f flag." +
```
timburks, line 37: Does `kubectl apply -f` split on commas? I didn't know about that.

cmd/registry/patch/apply.go
```diff
@@ -44,29 +44,31 @@ func Apply(ctx context.Context, client connection.RegistryClient, in io.Reader,
...
-			}
+			})
+		if err != nil {
 			return err
-		})
-	if err != nil {
-		return err
+		}
 	}
 	return patches.run(ctx, jobs)
```
timburks, line 73: Just double-checking: because we collect all the inputs before we call this, there's no sensitivity to the ordering of file names, right?

cmd/registry/cmd/apply/apply.go
theganyo: No idea if kubectl does, but this does.

cmd/registry/patch/apply.go
theganyo: What is your specific concern?

timburks: I wouldn't want a child resource to fail to apply because its parent hasn't been applied yet. That's already addressed in `patches.run()` so since we're only callling `patches.run()` once for all the inputs, we should be good.

timburks: OK, sg. We might not be checking the dependency in the tests.

cmd/registry/cmd/apply/apply.go
timburks: If kubectl doesn't do it, would you mind removing that? It's duplicative of multiple `-f` options, so it's only a small usability improvement.

theganyo: I'll remove the comment, but that's just the way that cobra works.

timburks (approved):

PR #1109, opened 2023-03-20
Accept gzip-encoded content in GetBytesForSpec()

timburks:
This follows [`visitor.FetchSpecContents()`](https://github.com/apigee/registry/blob/6781b6939b74bc960757218b68d319b4bed3d012/pkg/visitor/fetch.go#L32) in accepting spec contents as gzipped when available, but because callers expect uncompressed bytes, it also uncompresses them. It fixes problems that appear when callers try to work on specs that are over message size limits when uncompressed.

Arguably, one of `GetBytesForSpec()` and `FetchSpecContents()` should be eliminated.

#1110 for general discussion of content sizes.

### DIFF
--- a/pkg/visitor/content.go
+++ b/pkg/visitor/content.go
@@ -18,10 +18,13 @@ import (
 	"context"
 	"path"
 
+	"github.com/apigee/registry/cmd/registry/compress"
 	"github.com/apigee/registry/gapic"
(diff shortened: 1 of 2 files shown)


theganyo (approved):
LGTM, but we should test. :)

timburks:
OK, tested

PR #1111, opened 2023-03-20
return task error from WorkerPool

### LINKED ISSUE #917: Registry tool: errors encountered by concurrent workers are logged and not returned back to the calling command
As noted in the discussion of https://github.com/apigee/registry/pull/910#issuecomment-1377552891, when concurrent workers encounter errors, they [log them](https://github.com/apigee/registry/blob/98c9c9ae95a0874448f0ebc2e738a3c9ca68468d/cmd/registry/core/tasks.go#L81) instead of returning them to the calling command.

This conflicts with the goal in #827 of making error conditions testable.

theganyo:
Fixes #917

### DIFF
--- a/cmd/registry/cmd/annotate/annotate.go
+++ b/cmd/registry/cmd/annotate/annotate.go
@@ -55,7 +55,7 @@ func Command() *cobra.Command {
 				log.FromContext(ctx).WithError(err).Fatal("Failed to get client")
 			}
 
-			taskQueue, wait := tasks.WorkerPool(ctx, jobs)
+			taskQueue, wait := tasks.WorkerPoolIgnoreError(ctx, jobs)
 			defer wait()
 
 			valuesToClear := make([]string, 0)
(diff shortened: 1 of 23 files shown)


theganyo:
We use `defer wait()` pretty much everywhere. Since `wait()` will return an error now, though, I'm thinking the easiest change will be to replace all these instances to:

```
			defer func() {
				if err := wait(); err != nil {
					log.FromContext(ctx).WithError(err).Fatal("Failed to complete")
				}
			}()
```

timburks:
@theganyo OK - what you're describing could be in a `waitAndLogError()` function, which you could add and immediately deprecate, but that would reduce the LOC impact of this. 

I think at least some of these will be in functions that return errors, so we probably want to `return err` in those, which I think would be find to do in a subsequent PR.

theganyo:
> The frustrating situations that I hope we can avoid are when one task failure shuts everything down.

Oof... I was unaware of your desire for this behavior... and yes, that is explicitly what errgroup does. (Understandably, as this is often what is desirable.) So, yes, if you want to ignore errors and keep going with this implementation, you'd have to return nil from your tasks. Otherwise, I'll have to rewrite this.

timburks:
@theganyo let's go ahead with this and discuss offline whether we want to do anything else. There might be a better way to address that concern.

PR #1112, opened 2023-03-20
Fix error in "compute vocabulary" for protos, add test of proto vocabulary computation

timburks:
This will probably hurt our code coverage temporarily because it brings "compute vocabulary" into scope and only tests the proto vocabulary computation. (I will follow up with more test PRs)

### DIFF
--- a/cmd/registry/cmd/compute/vocabulary/vocabulary-proto.go
+++ b/cmd/registry/cmd/compute/vocabulary/vocabulary-proto.go
@@ -41,7 +41,7 @@ func NewVocabularyFromZippedProtos(b []byte) (*metrics.Vocabulary, error) {
 	}
 
 	for _, f := range r.File {
-		if strings.HasSuffix(f.Name, ".proto") {
+		if !strings.HasSuffix(f.Name, ".proto") {
 			continue
 		}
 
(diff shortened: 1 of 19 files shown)


cmd/registry/cmd/compute/vocabulary/vocabulary.go
```diff
@@ -100,6 +100,10 @@ func Command() *cobra.Command {
 			}
 		},
 	}
+	cmd.PersistentFlags().String("filter", "", "Filter selected resources")
+	cmd.PersistentFlags().Bool("dry-run", false, "if set, computation results will only be printed and will not stored in the registry")
+	cmd.PersistentFlags().Int("jobs", 10, "Number of actions to perform concurrently")
```
timburks, line 105: If these aren't here, the flags are unavailable when we instantiate the test command.

theganyo:
There is a lot here and I'm unsure what the original error was. Maybe we should walk through this.

cmd/registry/cmd/compute/vocabulary/vocabulary-proto.go
```diff
@@ -41,7 +41,7 @@ func NewVocabularyFromZippedProtos(b []byte) (*metrics.Vocabulary, error) {
 	}
 
 	for _, f := range r.File {
-		if strings.HasSuffix(f.Name, ".proto") {
+		if !strings.HasSuffix(f.Name, ".proto") {
```
timburks, line 44: @theganyo This is the error -- introduced in a big refactor a while ago, it was preventing the proto vocabulary computation from finding anything.

theganyo (approved):
I see, nearly all of the files are `testdata`.

Issue #1110, opened 2023-03-20
Registry server: plan for addressing message-size limits.


timburks:
gRPC has a default message size limit of 4MB. This is easily raised, so for self-hosted instances, that is probably the easiest way to support specs and artifacts that are larger than 4MB. But when other constraints limits prevent this, we're currently unable to load specs that are larger than a certain size (when compressed).

Compression helps significantly - the OpenAPI spec for Kubernetes 1.26.3 is 4620955 bytes and gzips to 285726 bytes, approximately 1/16th its uncompressed size. Because of this, we support and recommend uploading and downloading specs with gzip encoding.

Is that enough? What is a reasonable upper limit on compressed spec size?

What other upload/download options might we support?

timburks:
[The Katamari Damacy of APIs](https://github.com/APIs-guru/openapi-directory/blob/main/APIs/microsoft.com/graph-beta/1.0.1/openapi.yaml)

![image](https://user-images.githubusercontent.com/405/226406686-417dd408-bbba-408b-a998-08b7a7732beb.png)

PR #1122, opened 2023-03-23
Download specs gzipped for 'compute vocabulary' and visitor list operations.

timburks:
I noticed that `registry compute vocabulary` was fetching specs without compression (causing failures for specs that were larger than 4MB uncompressed #1110). When fixing this, I scanned the code for other calls to `GetApiSpecContents` that weren't accepting gzipped specs and found two in `pkg/visitor`. Because these functions were returning spec contents uncompressed, I also added code to uncompress gzipped specs before returning them. Looking ahead, we might want to more tightly encapsulate the calls to `GetApiSpecContents` so that we download gzipped specs as much as possible.

### DIFF
--- a/cmd/registry/cmd/compute/vocabulary/vocabulary.go
+++ b/cmd/registry/cmd/compute/vocabulary/vocabulary.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/apigee/registry/cmd/registry/compress"
 	"github.com/apigee/registry/cmd/registry/tasks"
 	"github.com/apigee/registry/pkg/connection"
 	"github.com/apigee/registry/pkg/log"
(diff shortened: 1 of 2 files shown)


pkg/visitor/fetch_test.go
```diff
@@ -91,7 +97,7 @@ func TestFetch(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Failed to fetch spec contents: %s", err)
 		}
-		if string(spec.Contents) != specContents {
+		if !bytes.Equal(spec.Contents, gzippedSpecContents) {
```
timburks, line 100: This and the following two changes show an inconsistency -- ListSpecs is now (in this PR) returning specs uncompressed. `GetSpecs` was previously downloading specs compressed and returning them without uncompressing. I think it's probably better for `GetSpecs` to also uncompress specs before returning them, but think it might overcomplicate this PR by making that change here.

theganyo (approved):

cmd/registry/cmd/compute/vocabulary/vocabulary.go
```diff
@@ -164,7 +164,7 @@ func (task *computeVocabularyTask) Run(ctx context.Context) error {
 			return nil
 		}
 	} else {
-		return fmt.Errorf("we don't know how to summarize %s", task.specName)
+		return fmt.Errorf("we don't know how to compute the vocabulary of %s", task.specName)
```
timburks, line 167: This was just a kick of a tire to get the CLA check to run again...  but it was also needed

Issue #1121, opened 2023-03-22
Registry tool: review and test `registry compute lint` and `registry compute lintstats`


timburks:
On the surface, these commands are subsumed by `registry compute conformance` and scoring, but I think it may be useful to have a direct way to invoke specific linters (for which we would use `registry compute lint`). I'm less sure that we would keep the `compute linstats` command.

PR #1124, opened 2023-03-23
Automatically uncompress all spec data returned by the visitor package.

### LINKED ISSUE #1123: `visitor.GetSpecs` should uncompress specs before returning them
As discussed here: https://github.com/apigee/registry/pull/1122#discussion_r1145601962

timburks:
This fixes #1123 by modifying `visitor.GetSpecs` (and the underlying  `visitor.FetchSpecContents`) to always return ungzipped specs. To avoid confusing downstream code, this also modifies the mime type of the returned specs to match the unzipped content (i.e. the "+gzip" is removed). This might be seen as misrepresenting the content stored on the server, but overall I think this is better because downstream code often switches on spec mime types, and this avoids accidental double-uncompression.

Moving this uncompression into the `visitor` package allows us to eliminate uncompression throughout the registry tool subcommands. Now any client code that uses `visitor` can safely expect gzipped specs to be uncompressed.

With a few other small changes, this eliminates `visitor.GetSpecBytes()`, replacing it with calls to `visitor.FetchSpecContents()`. `FetchSpecContents` is useful in situations where we are working on large numbers of specs and want to avoid loading all these specs into memory as we are putting them in task queues for processing. Instead we defer the "fetch" to when the specs are needed in the task handlers.

There's some untested code in `compute lint` that will be addressed in future PRs (#1121).

### DIFF
--- a/pkg/visitor/fetch_test.go
+++ b/pkg/visitor/fetch_test.go
@@ -30,8 +30,8 @@ func TestFetch(t *testing.T) {
 	project := names.Project{ProjectID: projectID}
 	parent := project.String() + "/locations/global"
 
-	specContents := "hello"
-	gzippedSpecContents, err := compress.GZippedBytes([]byte(specContents))
+	specContents := []byte("hello")
+	gzippedSpecContents, err := compress.GZippedBytes(specContents)
 	if err != nil {
 		t.Fatalf("Setup: Failed to compress test data: %s", err)
 	}
@@ -97,7 +97,7 @@ func TestFetch(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Failed to fetch spec contents: %s", err)
 		}
-		if !bytes.Equal(spec.Contents, gzippedSpecContents) {
+		if !bytes.Equal(spec.Contents, specContents) {
 			t.Fatalf("Fetched unexpected spec contents: wanted %q got %q", specContents, spec.Contents)
 		}
 	})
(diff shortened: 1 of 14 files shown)


pkg/visitor/fetch_test.go
```diff
@@ -97,7 +97,7 @@ func TestFetch(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Failed to fetch spec contents: %s", err)
 		}
-		if !bytes.Equal(spec.Contents, gzippedSpecContents) {
+		if !bytes.Equal(spec.Contents, specContents) {
```
timburks, line 100: This (and the two similar changes following this) fixes the inconsistency noted in #1122.

theganyo (approved):
Excellent. Love seeing all that red code. :)

timburks:
@theganyo thanks for the quick review! 

I am starting to get annoyed by uncovered code like this https://app.codecov.io/gh/apigee/registry/pull/1124/blob/pkg/visitor/fetch.go that could easily be covered by refactoring. Two levels of indentation could get that file to 100% code coverage, basically by flipping the sense of the tests and moving subsequent code inside the braces. At this point I'm just ranting, so I'll commit this as-is.

theganyo:
> I am starting to get annoyed by uncovered code like this

Yup. There is a bit of a tension there with [Go style decisions](https://google.github.io/styleguide/go/decisions#indent-error-flow), but I do think in cases like this it's easily understandable.

timburks:
@theganyo I generally agree with the "happy path" guidance, so at this point I think this is evidence that the coverage checker is wrong.

Issue #1107, opened 2023-03-19
Registry tool: can we simplify `registry` configuration in containers?


timburks:
I've been trying various ways of running the `registry` tool in a container in a kubernetes cluster, and currently this seems the best way to configure the tool - I put this at the start of whatever script that I want to run:

```
# configure the registry tool to work with the service running in the cluster
registry config configurations create cluster \
	--registry.address=registry.default.svc.cluster.local:80 \
	--registry.insecure=true \
	--registry.location=global \
	--registry.project=myproject
```

It would be nice to be able to just run a container with `kubectl run shell -ti --image IMAGE -- /bin/bash` and have it open with an already-configured `registry` connection (basically automatically doing the above step). How might we do this? A container-around-a-container? A ConfigMap that pushes in `.config/registry/active_config`?

timburks:
Running this in a dockerfile seems to work nicely:
```
# Configure the registry tool to run in a common cluster configuration.
RUN registry config configurations create cluster \
	--registry.address=registry.default.svc.cluster.local:80 \
	--registry.insecure=true \
	--registry.location=global \
	--registry.project=menagerie
```

theganyo:
Yes, that can work, but it would make the container statically configured. The easiest way to dynamically set config in K8s would be to just set env vars: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container.

theganyo:
(Yes, we had hoped to get rid of the env vars for the CLI... but also running a CLI in a container in a K8s cluster seems a little odd...?)

timburks:
> ...running a CLI in a container in a k8s cluster seems a little odd...

I have a "toollkit" container that has `registry`, `registry-experimental`, and a bunch of accessory tools installed (linters, generators, etc). I'd like to be able to run it in various ways as k8s jobs that either do a specific task (with a command-line passed in) or open a shell for exploring and trying the tools in the container. So in that sense, running the CLI makes sense to me. If I'm always using the CLI on the same project, I can just build my image so that it starts configured, which is handy, but maybe we want to keep/restore an environment-based override? For me this isn't urgent, but I expect it will come up for others.

theganyo:
You should be able to use env vars like "registry.address" etc. In fact, we're still binding the old APG_ variables: https://github.com/apigee/registry/blob/main/pkg/config/config.go#L200...  but those we'd still like to remove, I think?

theganyo:
[...]

PR #1142, opened 2023-04-04
drop the APG_ prefix from env vars

### LINKED ISSUE #1113: Registry tool: Let's drop the `APG_` prefix from environment variables.
This vestigal configuration came from the automatically-generated CLI, which we've since replaced with `registry config` subcommands. As discussed in #1107, if we remove the `APG_` prefix, then we can use `REGISTRY_ADDRESS`, etc. to configure the `registry` and related tools in containers (and anywhere).

theganyo:
Fixes #1113
See also #1107

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -44,7 +44,6 @@ var (
 	ErrReservedConfigName = fmt.Errorf("%q is reserved", ActivePointerFilename)
 
 	envBindings    = []string{"registry.address", "registry.insecure", "registry.token"}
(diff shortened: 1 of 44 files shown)


pkg/config/config.go
```diff
@@ -197,16 +196,6 @@ func ReadValid(name string) (c Configuration, err error) {
 		return
 	}
 
-	for _, env := range envBindings {
```
timburks, line 200: just curious: why is this going away?

timburks (approved):

PR #1147, opened 2023-04-14
update actions/checkout version

theganyo:
Addresses [build warning](https://github.com/apigee/registry/actions/runs/4692896484):

```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```


timburks (approved):
\o/

PR #1160, opened 2023-04-27
update project properties with registry apply

### LINKED ISSUE #907: Registry tool: `registry apply` does not support YAML project descriptions.
For open-source installations, I think I should be able to use `registry apply` and the following YAML to set project properties:
```
apiVersion: apigeeregistry/v1
kind: Project
metadata:
  name: openapi
data: 
  displayName: OpenAPI Directory
  description: APIs from the APIs.guru OpenAPI Directory
```

Following that, I should get matching output from `registry get` and `registry export` should write it into an `info.yaml` at the project level. Currently the `apply` and `export` parts of this are not implemented.

theganyo:
Partially addresses #907

### DIFF
--- a/cmd/registry/cmd/apply/apply.go
+++ b/cmd/registry/cmd/apply/apply.go
@@ -47,11 +47,11 @@ func Command() *cobra.Command {
 			}
 
 			ctx := cmd.Context()
+			c, err := connection.ActiveConfig()
+			if err != nil {
+				return err
+			}
 			if project == "" {
-				c, err := connection.ActiveConfig()
-				if err != nil {
-					return err
-				}
 				project, err = c.ProjectWithLocation()
 				if err != nil {
 					return fmt.Errorf("%s: please use --parent or set registry.project in configuration", err)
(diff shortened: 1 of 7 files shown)


cmd/registry/cmd/apply/apply.go
```diff
@@ -47,26 +47,31 @@ func Command() *cobra.Command {
...
 				project, err = c.ProjectWithLocation()
 				if err != nil {
 					return fmt.Errorf("%s: please use --parent or set registry.project in configuration", err)
 				}
 			} else if !strings.Contains(project, "/locations/") {
 				project += "/locations/global"
 			}
-			client, err := connection.NewRegistryClient(ctx)
+
+			client, err := connection.NewRegistryClientWithSettings(ctx, c)
```
timburks, line 63: nit: I was a little confused by `WithSettings` and looked up the function to see that `Settings` here is synonymous with `Config`. Could we change that to `WithConfig` in some future PR? I don't think it's important enough to delay this or the next PR that depends on this one, so no hurry.

cmd/registry/patch/apply.go
```diff
@@ -223,6 +225,8 @@ func (task *applyBytesTask) Run(ctx context.Context) error {
 
 	log.FromContext(ctx).Infof("Applying %s", task.resource())
 	switch header.Kind {
+	case "Project":
+		return applyProjectPatchBytes(ctx, task.adminClient, task.bytes)
```
timburks, line 229: It looks to me like everything in this PR is a setup for this handling this new case. That LGTM, and I guess if we are running with a hosted service, the request will just fail if a `Project` YAML is included. Also LGTM (thinking out loud).

timburks (approved):
Thanks, this looks good! I don't think my nit is worth disrupting the merge of this PR, so let's go ahead and merge this so we can get the next one merged also.

PR #1165, opened 2023-05-03
remove glob chars from spectral target filename

### LINKED ISSUE #1127: Registry tool: the Spectral linter plugin does not return results for certain spec filenames.
We noticed that a spec with a filename of `openapi (1).yaml` was not showing Spectral results in its conformance report. Changing its filename to `openapi.yaml` seemed to fix the problem. 

Currently the plugin just writes specs to whatever filename is specified for them with no validation.

theganyo:
Fixes #1127

### DIFF
--- a/cmd/registry/cmd/compute/lint/lint-openapi-spectral.go
+++ b/cmd/registry/cmd/compute/lint/lint-openapi-spectral.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 
 	"github.com/apigee/registry/pkg/application/style"
 )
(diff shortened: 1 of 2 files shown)


timburks (approved):
LGTM but I wonder if this is the best place to fix this. It seems to be doing a lot in the linter plugin to sanitize problematic filename values in the registry. Other linter plugins might have to do the same (and possibly anything else that uses spec filename fields). Let's go ahead with this and discuss if filename validation might eventually be better done at a higher level (i.e. check).

theganyo:
> problematic filename values in the registry

Yes, I considered that but as I suggested in the issue, I believe that's over-generalizing. I don't imagine that all filenames should be subject to glob validation... that would puts the burden on the API users to have to sanitize perfectly valid filenames.

theganyo:
This is basically what I was thinking: https://en.wikipedia.org/wiki/Robustness_principle :)

PR #1166, opened 2023-05-03
add debug flag that retains temporary files

### LINKED ISSUE #1130: Registry tool: linter plugins should accept a debug flag that retains temporary files.
https://github.com/apigee/registry/pull/1125#discussion_r1152255003

I think we would want to add a new field to [`LintRequest`](https://github.com/apigee/registry/blob/a38ff5815c415be5c7d608512cdd3d6c548035ed/google/cloud/apigeeregistry/v1/style/lint.proto#L125) for this. Retaining temporary files would also give us a way to extract any configuration files generated and used by the linters, such as [this Spectral configuration](https://github.com/apigee/registry/blob/a38ff5815c415be5c7d608512cdd3d6c548035ed/cmd/registry/plugins/registry-lint-spectral/main.go#L83).

theganyo:
Fixes #1130

### DIFF
--- a/cmd/registry/cmd/compute/lint/lint.go
+++ b/cmd/registry/cmd/compute/lint/lint.go
@@ -24,6 +24,7 @@ import (
 	"github.com/apigee/registry/cmd/registry/conformance"
 	"github.com/apigee/registry/cmd/registry/tasks"
 	"github.com/apigee/registry/pkg/connection"
+	"github.com/apigee/registry/pkg/log"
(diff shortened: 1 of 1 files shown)


theganyo:
FYI: #1130 suggested changing `LintRequest` to include a debug flag, but I didn't see any reason to do so for this as the behavior to remove the directory was at the level of the called of `LintRequest`.

timburks (approved):

PR #1167, opened 2023-05-04
show available plugins in registry cli

### LINKED ISSUE #1154: Registry tool: plugins
[This](https://blog.chmouel.com/2022/03/18/show-plugins-list-to-a-cli-when-using-gos-cobra-library/) suggests a way to implement plugins that might allow us to call `registry-experimental` subcommands from the `registry` tool using `registry experimental`. What's more, it could allow arbitrary additional plugins to be created and added to containers containing the `registry` tool. Those could still be called with their actual names `registry-new`, etc. so this mostly just seems nice-to-have.

theganyo:
Fixes #1154

Adds list of plugins to usage page and allows execution of these directly in registry command. Also moves `Need more help?` wiki link direclty into usage template. See below.

This is a very basic implementation in that it only looks for `registry-*` filenames on the path and doesn't have any information about the plugins. 

But give it a try and see what you think.

```
$ registry -h
A command-line tool for working with the API Registry

Usage:
registry [command]

Available Commands:
annotate    Annotate resources in the API Registry
apply       Apply YAML to the API Registry
auth        Manage authentication to the API Registry
check       Check entities in the API Registry
completion  Generate the autocompletion script for the specified shell
compute     Compute properties of resources in the API Registry
config      Maintain properties in the active configuration
delete      Delete resources from the API Registry
diff        Compare resources in the API Registry
export      Export resources from the API Registry
get         Get resources from the API Registry
help        Help about any command
label       Label resources in the API Registry
resolve     Resolve dependencies by performing actions in a specified manifest
rpc         Make direct calls to RPC methods
upload      Upload information to the API Registry

Available Plugins:
basenames
connect
decode-spec
encode-spec
experimental
graphql
lint-api-linter
lint-openapi-sample
lint-sample
lint-spectral
lint-test
server

Flags:
      --address string             the server and port of the Registry API (eg. localhost:8080)
  -h, --help                       help for registry
  -v, --version                    version for registry

Use "registry [command] --help" for more information about a command.

Need more help?
```

### DIFF
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -79,7 +79,7 @@ Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "he
 
 Available Plugins:
 {{- range Plugins}}
-  {{.}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
+{{.}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
 
 Flags:
 {{.LocalFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -17,10 +17,10 @@ package main
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


theganyo:
Yes, kubectl actually does do kind of a command hierarchy for it's plugins: https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#naming-a-plugin. Perhaps that would allow us to avoid having to hard-code exceptions to the lint-* files.

theganyo:
A few others...

The following hang when executed without arguments, so we should exclude them:
- basenames
- decode-spec
- encode-spec

This runs a server:
- graphql

Which will leave these for now:
- experimental
- connect

timburks (approved):

PR #1200, opened 2023-06-20
switch builder to golang:1.20-bullseye

### LINKED ISSUE #1199: Switch builder to golang:1.20-bullseye
This is to fix a glibc lib issue with distroless. See: https://github.com/apigee/registry/issues/1196#issuecomment-1599081638_

theganyo:
Fixes #1199


theganyo:
Note: `deployments/envoy/Dockerfile` and `containers/registry-tools/Dockerfile` still use golang:1.20 as builder as they don't use distroless and don't have the lib issue.

timburks (approved):

PR #1202, opened 2023-06-20
allow nil bodies in create resource endpoints

### LINKED ISSUE #1198: Registry server: Allow nil bodies in create resource endpoints
Create resource endpoints currently allow empty string but not nil (e.g. [here](https://github.com/apigee/registry/blob/main/server/registry/actions_apis.go#L37)). The server should be permissive and allow both nil and empty string.

theganyo:
Allow nil when creating:
* apis
* artifacts
* deployments
* projects
* specs
* versions

Fixes #1198

### DIFF
--- a/server/registry/actions_apis_test.go
+++ b/server/registry/actions_apis_test.go
@@ -157,6 +157,16 @@ func TestCreateApiResponseCodes(t *testing.T) {
 			want: codes.NotFound,
 		},
 		{
+			desc: "missing resource body",
+			seed: &rpc.Project{Name: "projects/my-project"},
+			req: &rpc.CreateApiRequest{
+				Parent: "projects/my-project/locations/global",
+				ApiId:  "valid-id",
+				Api:    nil,
+			},
+			want: codes.OK,
+		},
+		{
 			desc: "missing custom identifier",
 			seed: &rpc.Project{Name: "projects/my-project"},
 			req: &rpc.CreateApiRequest{
--- a/server/registry/actions_apis_test.go
+++ b/server/registry/actions_apis_test.go
@@ -74,6 +74,20 @@ func TestCreateApi(t *testing.T) {
 				},
 			},
 		},
+		{
+			desc: "empty resource",
+			seed: &rpc.Project{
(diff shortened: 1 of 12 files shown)


server/registry/actions_apis_test.go
```diff
@@ -142,16 +156,6 @@ func TestCreateApiResponseCodes(t *testing.T) {
...
-		{
-			desc: "missing resource body",
-			seed: &rpc.Project{Name: "projects/my-project"},
-			req: &rpc.CreateApiRequest{
-				Parent: "projects/my-project/locations/global",
-				ApiId:  "valid-id",
-				Api:    nil,
-			},
-			want: codes.InvalidArgument,
-		},
```
timburks, line 154: would it be difficult to keep these tests and change the expected value to codes.OK?

theganyo: No, I actually did that first... but: 1. it's redundant with the other test and, 2: there are no other OK checks here.

timburks: Does 1. mean there are other tests that pass nil resources to create command? I understand 2. but that's an artifact of the test organization and not really an indication of what we want to test.

theganyo: I added test for nil bodies that ensure that both there is no error and that the result is correct: https://github.com/apigee/registry/pull/1202/files#diff-25fea043835f743787d3ac3e0409c0e763af6e9bfdf07b0c7956394e3b604c48R78

theganyo: But it's fine. I'll add it back in if you'd like.

timburks: Thanks, I thought that adding those back might have kept our coverage from dropping (very slightly, but that was surprising) - anyway, this LGTM. Thanks!

timburks (approved):

server/registry/actions_apis_test.go
theganyo: I believe coverage percentage dropped because there is less code to cover. :)

PR #1217, opened 2023-08-09
add detailed help for get command

theganyo:


timburks (approved):
Thanks, I like the use of NAME instead of ID, preferring consistency with the YAML experience over the lower-level AIP style. It's unfortunate that they don't match, but I think the YAML conventions are more relevant.

